PR #13, opened 2017-05-18
fix old references

adizaimi:


bdb/locktest.c
```diff
@@ -906,7 +906,7 @@ static void genlockdesc(void)
     }
     fclose(urandom);
 }
-static void DRQS_56985015()
+static void test_56985015()
```
mponomar: test_56985015 -> lockvec_test?

adizaimi: Sure

PR #15, opened 2017-05-18
Do not ignore $(PREFIX) during install

nirbhayc:
* Use $(PREFIX) to enable non-sudo 'make install'
* Minor typo


mponomar:
On second thought - we don't need those at all in a local install.

`[ -z "$(DESTDIR)" ] && install ....`

What do you think?

PR #19, opened 2017-05-18
initial pass on the documentation Markdown for typos

patrickmmartin:
- make target that uses aspell, plus a custom dictionary
- custom dictionary 
- make target to attempt to carefully update the current dictionary


docs/Makefile
```diff
@@ -1,3 +1,10 @@
 clean:
 	rm -f src/sqlitegen/all.html src/sqlitegen/syntax_linkage.tcl
 
+
+spelings:
```
DonaldWhyte: Should be `spellings`. Although I suspect this is an ironic joke. :P

mponomar: This is one of those "immediately obvious in hindsight" things that I should have done.  Thanks Patrick.

patrickmmartin: It was deliberate - when, in 10 years' time, we run the AI spell checkers over all the comdb2 clones in the multiverse, there should only be one "orthographic issue " per repo....

docs/pages/config/config_files.md
```diff
@@ -32,28 +32,28 @@ for this file in one of these locations, in this order:
...
 
-Lines begining with '#' are treated as comments and skipped.  Other lines should consist of a directive and its value.
+Lines beginning with '#' are treated as comments and skipped.  Other lines should consist of a directive and its value.
 
 ## Database tunables.
 
 ### Cache size
 
-Cache size can be set with the the 'cache' tunable. It takes a cache size and a unit (kb, mb, gb).  For example
+Cache size can be set with the the 'cache' tunable. It takes a cache size and a unit (KB, MB, GB).  For example
```
mponomar: We only accept lower case version of those.  Should probably fix that, but for now at least going to make the docs match the code.

PR #23, opened 2017-05-19
Fix deb build.

mponomar:
Adding new toplevel files breaks the build.  It's a bit fragile.  roborivers is working on test case.


nirbhayc:
Looks good.

PR #24, opened 2017-05-19
We should not abort on error after commit in llmeta

adizaimi:
When we fail commit we should not abort -- commit itself
cleans up on failure.

### DIFF
--- a/bdb/llmeta.c
+++ b/bdb/llmeta.c
@@ -1358,17 +1358,13 @@ int bdb_llmeta_set_tables(
                         "this should not happen",
                 __func__);
         *bdberr = BDBERR_MISC;
-        free(p_buf_start);
-        p_buf = NULL;
-        return -1;
+        goto cleanup;
     }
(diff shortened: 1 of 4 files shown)


bdb/llmeta.c
```diff
@@ -2682,7 +2677,7 @@ int bdb_add_dummy_llmeta(void)
         logmsg(LOGMSG_ERROR, "%s bdb_tran_commit rc: %d bdberr: %d\n", __func__, rc,
                 bdberr);
         tran = NULL;
-        goto fail;
+        return -1;
```
mponomar: I think this code was correct.  There's no parent transaction, it's safe to retry.  And there was no "abort after commit" bug.

adizaimi: Thank you for catching. We set tran=NULL on failed commit, so goto fail is actually ok.

adizaimi:
thanks Mike.

PR #31, opened 2017-05-20
Adds ability to build the jdbc driver using a docker container

dennisgove:
This allows one to build the jdbc driver without any necessary dependencies on the host machine other than docker (ie, a particular version protoc is not required). The resulting .jar file will be placed in the standard location `<root>/cdb2jdbc/target`

To build the jdbc driver using docker run the following command in the root directory
```
$> make jdbc-docker-build
```

This also changes the version of protoc to 3.2.0

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ include libs.mk
 export SRCHOME=.
 export DESTDIR
 export PREFIX
+BASEDIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 
 # Common CFLAGS
 CPPFLAGS+=-I$(SRCHOME)/dlmalloc $(OPTBBINCLUDE)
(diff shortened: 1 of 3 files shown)


dennisgove:
I haven't yet updated any documentation - please let me know if I should update [docs/pages/programming/jdbc_api.md](docs/pages/programming/jdbc_api.md) or somewhere else.

dennisgove:
I changed the version of protobuf to 3.2.0 because I was unable to find a version of 2.6.1 that I could download. I have no preference for 3.2.0 so if you know of a place I can download 2.6.1 from I'm comfortable changing this back.

That said, tests pass with 3.2.0 so might be good to upgrade.

mponomar:
Works well for me.  We should really incorporate jdbc tests into the main test suite.  Also probably write more of them.

dennisgove:
A couple todo items before merge:

* Run container as the user so the resulting artifacts are owned by the user and not `root`.
* Add docs (to be determined where)

dennisgove:
Regarding tests - yes, absolutely. Beyond just the unit tests, we're working on a containerized build of comdb2, and the ability to spin up a comdb2 container itself. Among other things, this will allow us to add true integration tests to the cdb2jdbc code which will spin up a running comdb2 instance and execute a full integration test suite against it. Because this is all containerized, it can be run anywhere with ability to run Docker containers.

PR #33, opened 2017-05-21
Make deb-current rule less fragile

mponomar:


riverszhang89:
Rerunning with DPKG test.

PR #38, opened 2017-05-22
Use CLOCK_REALTIME to schedule cron events.

riverszhang89:
The function time() is imprecise. Using time() to schedule cron events causes
unnecessary spins on the wait condition when time() is behind the realtime clock.


riverszhang89:
This was internally reviewed by @jmleddy .

PR #39, opened 2017-05-22
Msg subsystem v4

jmleddy:


riverszhang89:
What's the secret change to fix bulk_overflow crash?

jmleddy:
No secret, just better build and tests.

jmleddy:
> Needs utility to dump/(replay?) $dbname.sqllog

less?

mponomar:
$dbname.sqllog is a binary log (protobuf events).  Your sql log is text, less is great there.  I was wondering if we need the text log.

jmleddy:
Well we have replay and a viewer for the text log, so it's more featureful at this point. If redundancy is the issue, I don't think we need the sqllog.

jmleddy:
Since the tests passed, I'm going to merge this for now. If we need to, we can revisit the conversation about getting text output of the sqlllog and whether it's worth it to keep one or the other.

mponomar:
We wanted logging for a couple of reasons.  One is the ability to replay queries (for testing, for QA, for performance tests, whatever).  That's easy to do with the protobuf log.  It's perfectly possible with a text log, but it needs more work (eg: need a clean way to log bound parameter values, escape things, etc.).  Another is to track performance metrics (per query and overall) and pass them to other systems for analysis.  Again, easier with a structured log.  Up to you.  I know binary logs aren't well liked, but if we had an easy utility to dump the data, I think we'd do better.

jmleddy:
I hate structured logs personally, but if you insist I can transition over to them. I think we ought to be consistent though, moving trc.c and statsreqs over too. In any case, escaping sql should not be an issue today, as it starts at `sql=` and ends at `\n`. Everything in between is literal. I admit I haven't tried bound parameters yet.

jmleddy:
Also I don't think it's easier to track performance metrics with a structured log.

mponomar:
trc.c is a debugging log.  statreqs would go away if we log the same information in your log.

>> Also I don't think it's easier to track performance metrics with a structured log.

Fair enough.  I was stretching my point a bit.  It's easier to pass them to another system.  Arguably.

mponomar:
(longreqs also goes away)

mponomar:
Also I don't insist.  As long as we're able to identify which metrics go with which events, and have some consistent format that's easy to parse, text is fine.  My only argument is that structured log gets you that for free (yay, no parsing).

adizaimi:
Can we have a setting, and the log can be structured or text as user wishes based on that?

PR #40, opened 2017-05-23
Constant Time Comparison

script3r:
I noticed `comdb2` makes uses of `OpenSSL` for various parts of the system. 

We could leverage `CRYPTO_memcmp` to perform constant-time comparison of secrets when `OpenSSL` is available. In case it is not available, we can fallback to a less optimized implementation. 

https://cryptocoding.net/index.php/Coding_rules#Compare_secret_strings_in_constant_time

### DIFF
--- a/berkdb/crypto/crypto.c
+++ b/berkdb/crypto/crypto.c
@@ -22,6 +22,36 @@ static const char revid[] = "$Id: crypto.c,v 1.23 2003/06/30 17:19:41 bostic Exp
 #include "dbinc/db_page.h"
 #include "dbinc/crypto.h"
 
+#ifndef HAVE_CRYPTO
+/**
+ * __crypto_compare --
+ * 	returns zero iff `len` bytes at `src` and `tgt` are equal. Time is independent
+ *	of the contents of `src` and `tgt`.
+*/
+int
+__crypto_compare(src, dst, len)
+	const void *src;
+	const void *dst;
+	size_t len;
+{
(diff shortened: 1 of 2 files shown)


berkdb/crypto/crypto.c
```diff
@@ -22,6 +22,36 @@ static const char revid[] = "$Id: crypto.c,v 1.23 2003/06/30 17:19:41 bostic Exp
...
+ *	of the contents of `src` and `tgt`.
+*/
+int
+__crypto_compare(src, dst, len)
+	const void *src;
+	const void *dst;
+	size_t len;
+{
+	const unsigned char *a = (const unsigned char *)src;
+	const unsigned char *b = (const unsigned char *)dst;
```
riverszhang89: I don't think the explicit type conversion is needed: a void pointer can be safely promoted to any other types. and both sides have `const` qualifier.

riverszhang89:
Very nice catch and very solid code. Thanks for the contribution.

script3r:
@riverszhang89 looks like we can do the same for `https://github.com/bloomberg/comdb2/blob/66536ed305f5647a44eb2a0f51726a6fc68a3989/bdb/llmeta.c#L8394`.

I'll produce a patch later in the evening. Looking to make use of `comdb2` in a personal project soon.

riverszhang89:
@script3r Awesome. Thanks! Please don't hesitate to reach out at opencomdb2@bloomberg.net if you have any questions. We look forward to your feedback :)

PR #41, opened 2017-05-23
makecluster script

adizaimi:
Add tool to create a cluster and start it.
Create DB in current directory or use --dir to specify DB dir. Example run:
```
bash> comdb2makecluster mydb node1 node2
mydb: creating
mydb: copying to cluster
mydb: starting
mydb: waiting until ready
mydb: Cluster is running with 2 nodes. You can query it with:
   cdb2sql --cdb2cfg mydb/comdb2.cfg mydb 'select 1'
```


mponomar:
I hesitated putting a variant of this in comdb2admin because I didn't know how to handle the key distribution problem.  I guess it's reasonable to expect that the machines are already set up to trust each other's keys?

riverszhang89:
I think it is totally reasonable to assume the machines are already set up properly. In fact, writing a universal SSH setup program can be very tricky:

- The program itself must be able to connect to these machines first in order to distribute SSH keys, and the SSH authentication can be password based or key based.

- If the target "machines" are just a bunch of docker contains, it is needless to distribute SSH keys.

adizaimi:
I think having a separate util that creates and starts a cluster for testing purposes is nice. While ultimately we can integrate it in comdb2admin, we were discussing to check it in for now and decide what to do later. Not sure if i should update the docs, but https://bloomberg.github.io/comdb2/cluster.html could have a top paragraph with: "Easy way of starting cluster:"<pre> makecluster --dir ~/mydb mydb node1 node2 node3</pre>

riverszhang89:
If you'd like to make it a separate script, here are the things I'd think you might want to do:

- Fix SSL issue: The script only works with $SKIP_SSL and will get stuck in an infinite loop if $SKIP_SSL is not present.
- Change the name to comdb2makecluster instead of makecomdb2cluster. I'd expect to type `comdb2` and hit `<tab>` to get all available commands. We've already messed up a bit by having `comdb2` and `cdb2` prefixes, let's at least not make it worse.
- Place it under contrib/comdb2makecluster to avoid any confusion.
- Make install also installs your script.

adizaimi:
Rivers, good suggestions, thanks.

adizaimi:
I'm not sure where would this tool get installed, $(DESTDIR)$(PREFIX)/bin/ like cdb2sql? 
(but then it should probably be in tools/ like copycomdb2. In fact, copycomdb2 should be moved to tools too, no?).

riverszhang89:
`$(PREFIX)/bin/` is fine. Pretty sure we install copycomdb2 in $PREFIX/bin too.

nirbhayc:
IMO, all scripts/bins should end up in $(DESTDIR)$(PREFIX)/bin/.

adizaimi:
Nirbhay, yes adding to the dir you are mentioning.

contrib/comdb2makecluster
```diff
@@ -0,0 +1,291 @@
...
+CDB2_CONFIG=$DBDIR/comdb2db.cfg
+CDB2_OPTIONS="--cdb2cfg ${CDB2_CONFIG}"
+comdb2task=comdb2
+
+which $comdb2task > /dev/null || failexit "$comdb2task not in path"
+which copycomdb2  > /dev/null || failexit "copycomdb2 not in path"
+
+set -e
+
+vars="DBNAME DBDIR TMPDIR CDB2_OPTIONS CDB2_CONFIG"
```
mponomar: This looks like remnants of setting up a test - not needed?

contrib/comdb2makecluster
```diff
@@ -0,0 +1,291 @@
...
+        echo "$required not set" >&2
+        exit 1
+    fi
+#    echo "$required=$q"
+done
+
+# Setup a debugger to run comdb2 server
+DEBUG_PREFIX=
+
+if [[ -n ${DEBUGGER} ]]; then
```
mponomar: Did you want to keep the debug options here?  They're very useful for `make test`, but probably not for this.

contrib/comdb2makecluster
```diff
@@ -0,0 +1,291 @@
...
+fi
+
+mkdir -p $DBDIR $TMPDIR
+
+# setup files:
+echo "$DBNAME: creating"
+LRL="$DBDIR/$DBNAME.lrl"
+> ${LRL}
+
+if [[ -z $SKIPSSL ]] ; then
```
mponomar: This is also a test setup remnant.

contrib/comdb2makecluster
```diff
@@ -0,0 +1,291 @@
...
+ssl_client_mode REQUIRE
+ssl_cert_path $DBDIR
+EOPTIONS
+fi
+
+cat >> $DBDIR/$DBNAME.lrl <<EOPTIONS
+name    $DBNAME
+dir     $DBDIR
+
+setattr MASTER_REJECT_REQUESTS 0
```
mponomar: We needed this for tests, but I don't think we want this to become a default.

contrib/comdb2makecluster
```diff
@@ -0,0 +1,291 @@
...
+    echo "portmux_bind_path $pmux_socket" >> ${LRL}
+fi
+stop_pmux="pgrep pmux > /dev/null && (exec 3<>/dev/tcp/localhost/${pmux_port} && echo exit >&3 )"
+
+
+COPIEDTOCLUSTER=${TMPDIR}/copiedtocluster.log
+export COMDB2AR_EXOPTS="-x $comdb2task"
+
+
+keygen() {
```
mponomar: Test suite?

contrib/comdb2makecluster
```diff
@@ -0,0 +1,291 @@
...
+$comdb2task --create --lrl $DBDIR/${DBNAME}.lrl --pidfile ${TMPDIR}/$DBNAME.pid $DBNAME  >$DBDIR/logs/${DBNAME}.init 2>&1
+rc=$?
+rm -f ${DBNAME}.trap
+if [[ $rc -ne 0 ]]; then
+    exit 1
+fi
+
+# start it
+if [[ -z "$CLUSTER" ]]; then
+    echo "$DBNAME: starting single node"
```
mponomar: Shouldn't be reachable.

contrib/comdb2makecluster
```diff
@@ -0,0 +1,291 @@
...
+    echo "$DBNAME: waiting until ready"
+    sleep 1
+    for node in $CLUSTER; do
+        out=$(cdb2sql ${CDB2_OPTIONS} --tabs --host $node $DBNAME 'select 1' 2>&1)
+        while  [[ "$out" != "1" ]]; do
+            sleep 2
+            out=$(cdb2sql ${CDB2_OPTIONS} --tabs --host $node $DBNAME 'select 1' 2>&1)
+        done
+    done
+    n=$(cdb2sql ${CDB2_OPTIONS} -tabs $DBNAME --host $node 'exec procedure sys.cmd.send("bdb cluster")' | grep "lsn" | wc -l)
```
mponomar: You generate a config file, mention that it can be used to configure the destination?  Up to you.

adizaimi:
I think this is now ready to be checked in. I addressed all the commends. Can we take another look before I merge? Thanks.

contrib/comdb2makecluster
```diff
@@ -0,0 +1,251 @@
...
+
+
+
+
+keygen() {
+# Setup ssl certificate
+    openssl req -x509 -newkey rsa:4096 -keyout ${DBDIR}/server.key \
+        -out ${DBDIR}/server.crt \
+        -days 365 -nodes -subj \
+        "/C=US/ST=New York/L=New York/O=Bloomberg/OU=Comdb2/CN=*.bloomberg.com" &>> $DBDIR/setup.log
```
riverszhang89: I don't think we should generate SSL keypairs for end-users.

contrib/comdb2makecluster
```diff
@@ -0,0 +1,251 @@
+#!/usr/bin/env bash
```
riverszhang89: Can you please place the script under `contrib/comdb2makecluster/`, but not directly under `contrib/`?

contrib/comdb2makecluster
adizaimi: Ok. Was not sure about ssl keys. Easy to remove.

contrib/comdb2makecluster
adizaimi: Will do.

adizaimi:
Removed ssl keygen and moved script to subdir. Tested that it works to create a db cluster:
```
> contrib/comdb2makecluster/comdb2makecluster --dir /dev/shm/trydb trydb node1 node2 
trydb: creating
trydb created successfully
trydb: copying to cluster
trydb: starting
trydb: waiting until ready
trydb: Cluster is running with 2 nodes. You can query it with:
   cdb2sql --cdb2cfg /dev/shm/trydb/comdb2db.cfg trydb default 'select 1'

>cdb2sql --cdb2cfg /dev/shm/trydb/comdb2db.cfg trydb default 'select 1'
(1=1)
```

Also it installs correctly to the designed install directory (in this case /dev/shm/akabb/):
```
> make install 
....
-- Installing: /dev/shm/akabb/bin/comdb2makecluster
> ls -al /dev/shm/akabb/bin/comdb2makecluster          
-rwxr-xr-x 1 azaimi general 6215 Jan 26 11:57 /dev/shm/akabb/bin/comdb2makecluster
```

PR #46, opened 2017-05-23
Update JDBC SSL documentation.

riverszhang89:
JDBC SSL documentation is still based on the 1st SSL implementation which required client API to initiate SSL.
The latest implementation has a much more convenient model: either side can initiate SSL.
Update JDBC SSL documentation to reflect the latest changes.

#skipbuild


riverszhang89:
Thanks Mike.

PR #47, opened 2017-05-24
Avoid possibly overflowing hosts.

script3r:
A few more details were sent to `opencomdb2@bloomberg.net`.

### DIFF
--- a/cdb2api/cdb2api.c
+++ b/cdb2api/cdb2api.c
@@ -4764,9 +4764,11 @@ int cdb2_open(cdb2_hndl_tp **handle, const char *dbname, const char *type,
         p = strchr(type, ':');
         if (p) {
             strncpy(hndl->hosts[0], type, sizeof(hndl->hosts[0])-1);
-            p = strchr(hndl->hosts[0], ':');
-            *p = 0;
-            hndl->ports[0] = atoi(p + 1);
+	    p = strchr(hndl->hosts[0], ':');
+	    if (p) {
+		*p = 0;
+		hndl->ports[0] = atoi(p + 1);
+	    }
         } else {
             strncpy(hndl->hosts[0], type, sizeof(hndl->hosts[0])-1);
             if (!allow_pmux_route) {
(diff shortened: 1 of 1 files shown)


cdb2api/cdb2api.c
```diff
@@ -4763,7 +4763,7 @@ int cdb2_open(cdb2_hndl_tp **handle, const char *dbname, const char *type,
         char *p;
         p = strchr(type, ':');
         if (p) {
-            strcpy(hndl->hosts[0], type);
+            strncpy(hndl->hosts[0], type, sizeof(hndl->hosts[0]));
```
adizaimi: sizeof - 1?

script3r: @adizaimi, Indeed, however we still need to explicitly nul-terminate: something like `strlcpy` will get us both in one call.

We can probably perform better validation elsewhere, but I'm still familiarizing myself with the code. I'll patch this with explicit nul termination.

riverszhang89: `hndl` is allocated via `calloc` so it is fine to leave `NUL`. However I'd agree that explicit nul termination definitely improves the readability.

akshatsikarwar: Also see `bb/str0.c` for related functions

script3r: @akshatsikarwar I'll take a look at bb/str0.c, I was actually looking for `misc` string functions.
@riverszhang89, @adizaimi  you are right, the -1 will suffice.

cdb2api/cdb2api.c
```diff
@@ -4763,7 +4763,7 @@ int cdb2_open(cdb2_hndl_tp **handle, const char *dbname, const char *type,
         char *p;
         p = strchr(type, ':');
         if (p) {
-            strncpy(hndl->hosts[0], type, sizeof(hndl->hosts[0]));
+            strncpy(hndl->hosts[0], type, sizeof(hndl->hosts[0])-1);
```
riverszhang89: I do think our coding style requires spaces around binary operands.

script3r:
There's a bit more, unfortunately, we can still overflow since the following two lines assume that `:` exists in the host, and proceeds to access  the memory; however, since we (possibly) trimmed, we need to account for `:` not being present.

```
p = strchr(hndl->hosts[0], ':');
*p = 0;
```

riverszhang89:
@script3r Actually if `:` is not found, `strchr` returns `NULL`, and we end up dereferencing `NULL`.

PR #50, opened 2017-05-24
add new testcase update_datetime.test 

adizaimi:
Adding new testcase update_datetime.test to show updating
a datetime field via a trigger as the result of an update
to a given table.


riverszhang89:
I don't think the current implementation is very useful. For example, I am not able to use OR with this.
It might be a better idea to make the function behave like where() - it takes a string argument which is the where clause.

adizaimi:
Rivers you are right, can not do OR. 
tbl:update() was a fun exercise, but will remove as it is not very useful (assuming I can get the sql version to work).

PR #54, opened 2017-05-25
Don't set OPTBBINCLUDE

mponomar:
No longer needed, and can break builds if /opt/bb/include files conflict with ours.


adizaimi:
This is fine

PR #55, opened 2017-05-25
Add which to CentOS install instructions

mthorpe7:
While its normally installed in the base CentOS image, its a requirement of several scripts. Else you end up with errors like:

```
berkdb/dist/geninc.sh: line 13: which: command not found
berkdb/dist/geninc.sh: line 101: -f: command not found
```


mponomar:
which is present by  default in CentOS installs, but not in CentOS docker images.  It's still a dependency.

PR #58, opened 2017-05-25
comdb2aws + supervisor

riverszhang89:
1. Use `comdb2` user. Users can override it by doing `export SSHUSER=<my_db_user>`.

2. Use supervisord to manage databases.

3. Add drop-database script.

4. Update AMI id.

5. Update documentation accordingly.


mponomar:
Other ideas:

comdb2aws ui cluster
ssh to machine with port forwarding, open browser for supervisor

(don't know if ssh-agent allows...)
query agent if it has a key, add automatically if initial ssh fails.

riverszhang89:
```
ssh to machine with port forwarding, open browser for supervisor
```

This would require a desktop environment.

riverszhang89:
In theory we could use text-based browsers if no desktop environment available (still not sure how to detect this, I guess it could be done by `ps | grep X11`?). However, I tried elinks and it did not render the webpage properly.

riverszhang89:
Thanks for the review. I will merge this after roborivers says yes.

mponomar:
> This would require a desktop environment.

It would only be useful for people with a desktop environment.   Something like this:

     [[ -x $(which xdg-open 2>/dev/null) ]] && xdg-open http://localhost:9001

Just an idea, we don't have to do this.

PR #65, opened 2017-05-26
Fixes #64: Adds folder /var/cdb2

### LINKED ISSUE #64: Run of docker container comdb2-dev fails due to missing folder
When running the container built with `make docker-dev` it fails to create a database because directory `/var/cdb2` does not exist.

dennisgove:
This fixes issue #64 by creating the missing folder


mponomar:
Fixed by #66

Issue #69, opened 2017-05-28
quick start build instructions outdated


mihalis68:
The quick start guide has wrong build instructions for Ubuntu 16. The ones in docs work but not the quick start

PR #70, opened 2017-05-28
Fix build instructions on ubuntu 16

mihalis68:
Fixes https://github.com/bloomberg/comdb2/issues/69


mihalis68:
this was my mistake ignore

PR #71, opened 2017-05-28
make the package install commands actually possible to cut and paste

mihalis68:
and use directly


mponomar:
This makes the command scroll for me.  I don't know if that's worse.  Either way, cut-n-paste of the entire command will work.

PR #72, opened 2017-05-30
WIP: need to perform validation on database name

script3r:
While attempting to create a database with 63 characters, I noticed that even though it is a valid length, it was flaking in an interesting way: the output claimed the database was created, but in reality, it wasn't.

I ran comdb2 with `strace ./comdb2 --create `perl -e "print 'h'x63"`` and noticed an odd `unlink` system call to `unlink("atas7")` and `unlink("atas6")`. I compared this against a normal execution, and noticed this was failing at the creation of `sqlite_stat_*.datas*` files.

This is a `WIP`, but I'm looking to improve on a few things:

- stronger validation for what constitutes a valid database name
- continue the usage of `PATH_MAX` found in newer code, since we have to worry not only about the length of the database name, but also the path it will reside
- we should probably get away from `unlink` and instead ask the client to remove the database if it exists? or introduce a `--force` flag? as it is, an attacker can traverse paths by providing a name that contains '..' for example.
- trying to see how we can finalize `is_valid_dbname`, and find the most suitable point to call it.

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -8852,6 +8852,38 @@ int check_current_schemas(void)
     return 0;
 }
 
+/**
+ * -- is_valid_dbname
+ * decides whether a given `dbname` is valid database name.
+ */
+int is_valid_dbname(const char *dbname) {
+	// cannot be null.
+	if(dbname == NULL) {
+		logmsg(LOGMSG_ERROR, "DATABASE NAME CANNOT BE EMPTY\n");
+		return -1;
+
+	}
+
+	// cannot be larger than `MAX_DBNAME_LENGTH`
+	size_t len;
+	len = strlen(dbname);
+	if (len >= MAX_DBNAME_LENGTH) {
+		logmsg(LOGMSG_ERROR, "INVALID DATABASE NAME, MUST BE < %d CHARACTERS\n", MAX_DBNAME_LENGTH);
+		return -1;
+	}
+	
+	// alphanumeric and _ are allowed.
+	const char *p = dbname;
+	while (*p++) {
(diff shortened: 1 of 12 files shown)


db/comdb2.c
```diff
@@ -8852,6 +8852,38 @@ int check_current_schemas(void)
...
+	len = strlen(dbname);
+	if (len >= MAX_DBNAME_LENGTH) {
+		logmsg(LOGMSG_ERROR, "INVALID DATABASE NAME, MUST BE < %d CHARACTERS\n", MAX_DBNAME_LENGTH);
+		return -1;
+	}
+	
+	// alphanumeric and _ are allowed.
+	const char *p = dbname;
+	while (*p++) {
+		if(!isalnum(*p) && *p != '_') {
```
mponomar: '_' is oddly enough not allowed.

dorinhogea: it has to do with remote tables and machine class

script3r: Got it. What about `.` or `/` or `\`, any of those allowed?

akshatsikarwar: Just `isalnum()` + `-`

mponomar: I'm not 100% sure what impact remote databases will have.  `_` has special meaning there. Everything else that's a valid SQL identifier should be ok (and things that aren't may be quotable).  Will check regarding maximum supportable length.

nirbhayc:
What's the current status of this PR? @script3r Are you working on the suggestions or waiting on review?

script3r:
@nirbhayc I was out of the country, will complete this pull request soon.

PR #76, opened 2017-05-31
Fix crash on long column names, allow long column names in SP

mohitkhullar:
Add tests


akshatsikarwar:
What are the column names when emitting synthetic row without defining emit schema?

mohitkhullar:
$col_num
https://github.com/mohitkhullar/comdb2/blob/29b17a027211d0bee557f3e82560ab5385a4ad8d/lua/sp.c#L911

akshatsikarwar:
Good

PR #78, opened 2017-05-31
don't dereference prefault_dbp if send_prefault_udp is off

adizaimi:


riverszhang89:
Skip build for now. I am testing smoke test.

berkdb/mp/mp_fget.c
```diff
@@ -1174,52 +1174,47 @@ __memp_fget(dbmfp, pgnoaddr, flags, addrp)
...
-						   to it. Add the page. */
-						__memp_add_sparse_page(dbmfp->dbenv,
-								prefault_dbp->log_filename->id,
-								prefault_dbp->fileid,
-								*pgnoaddr, sparseness);
-					}
-				}
-			}
-		}
+    ret = __memp_fget_internal(dbmfp, pgnoaddr, flags, addrp, &did_io);
```
akshatsikarwar: Use tabstop=8 noet sw=8 for berkdb code

ldennis:
LGTM

adizaimi:
online_compaction is failing so i wonder if this change is affecting it.

riverszhang89:
Does online_compaction pass if you run it manually?

adizaimi:
Does not pass running manually. Debugging via printf, seems that function __memp_fget() is called with pgcomp_enabled=1 and prefault_dbp=NULL.

riverszhang89:
Looks like we can't simply set prefault_dbp to NUL before exiting memp_fget(). The dbp may get pages multiple times and only the 1st time we may be able to trigger page prefault or compaction.

riverszhang89:
Looks great! Thanks Adi!

PR #81, opened 2017-06-01
Ignore failure to install various service files

akshatsikarwar:
Fix my usecase: PREFIX=$HOME make install
Remove duplicate pmux.service install
Remove obsolete svn info

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,11 +106,10 @@ install: all
 	[ -z "$(DESTDIR)" ] && rm -f $(DESTDIR)$(PREFIX)/bin/cdb2sockpool && ln $(DESTDIR)$(PREFIX)/bin/comdb2 $(DESTDIR)$(PREFIX)/bin/cdb2sockpool || true
 	install -D comdb2ar $(DESTDIR)$(PREFIX)/bin/comdb2ar
 	install -D pmux $(DESTDIR)$(PREFIX)/bin/pmux
-	install -D tools/pmux/pmux.service $(DESTDIR)/lib/systemd/system/pmux.service
-	install -D tools/cdb2sockpool/cdb2sockpool.service $(DESTDIR)/lib/systemd/system/cdb2sockpool.service
-	install -D contrib/comdb2admin/supervisor_cdb2.service $(DESTDIR)/lib/systemd/system/supervisor_cdb2.service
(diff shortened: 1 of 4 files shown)


akshatsikarwar:
We could do that - Need to update path to binaries in the service file. Maybe separate PR if ever there is need.

mponomar:
It's a recurring theme.  See #15 - I'd say there's a need.  This is ok as a bandaid.

akshatsikarwar:
Deb smoketest passes. RPM fails with:
`File not found: /root/rpmbuild/BUILDROOT/comdb2-7.0.0pre-1.el7.centos.x86_64/opt/bb/lib/systemd/system/pmux.service`

But file was installed to:
`/root/rpmbuild/BUILDROOT/comdb2-7.0.0pre-1.el7.centos.x86_64/lib/systemd/system/pmux.service`

akshatsikarwar:
From #15: 

> This is by design. They are files needed by other systems, and get installed outside the default $PREFIX. 

And I agree. Non-root installs won't have much use for these files.

akshatsikarwar:
Why do we require installing `pmux.service` at two different locations? My change to remove that is failing rpm smoketest looks like.

akshatsikarwar:
Have both:
https://github.com/bloomberg/comdb2/blob/master/rpmbuild/comdb2.spec#L39
and
https://github.com/bloomberg/comdb2/blob/master/rpmbuild/comdb2.spec#L46

I guess we don't need 46.

akshatsikarwar:
Rivers points out that:
https://github.com/bloomberg/comdb2/blob/master/rpmbuild/comdb2.spec#L74

cp /opt/bb/lib/systemd/system/pmux.service /etc/systemd/system

How many locations do we need to install the service file?

mponomar:
Just the one.  Why don't we just copy it to $(DESTDIR)/$(PREFIX), and install it in postinstall scripts?

akshatsikarwar:
There is no post-install. I don't have root. Thats why I'm doing a PREFIX install in the first place. I don't need these services running at boot.

mponomar:
@nirbhayc made the argument that someone installing to a directory to build an install tarball will still need the files in place.  I don't really have an opinion.  I'm ok with skipping them entirely from a build with a set PREFIX.

akshatsikarwar:
Thats fair. I didn't think of tarball installs.

PR #83, opened 2017-06-01
Logging_v2

jmleddy:
RFC: Also basically a testing branch to see if we can pass roborivers.


akshatsikarwar:
What do I run to test and see output?

jmleddy:
I'm waiting for roborivers, then will get output from him

PR #87, opened 2017-06-02
db Fix bad LRL directives.

riverszhang89:
Also update linearizable.lrl as `no_perfeck_ckp' is replace by `perfeck_ckp 0'.


riverszhang89:
Thank you Mike.

PR #90, opened 2017-06-03
Add option 'none' for compression

adizaimi:
Add none as a compression option
Reenable cdb2dump test, passing compr none to store records uncompressed
Send flush to fix clustered cdb2dump test


akshatsikarwar:
Looks good.

adizaimi:
Thanks Akshat. Now need to change the Docs.

PR #92, opened 2017-06-05
IO-less checkpoints.

riverszhang89:
With perfect checkpoints, we only sync pages with first-dirty-LSN less than
a checkpoint LSN. Mark came up with this absolutely IO-less checkpoints idea
on top of perfeck checkpoints: instead of using the begin LSN of the oldest
active transaction as the checkpoint LSN, we find the oldest transaction
(not necessarily active) which first marked a page dirty among all modified
pages in the buffer pool, and use the begin LSN of that transaction as the
checkpoint LSN. Therefore we don't need to flush anything because all dirty
pages' LSN-s will be greater than the checkpoint LSN.

The only downside is that recovery takes more time under IO-less checkpoints.

### DIFF
--- a/bdb/file.c
+++ b/bdb/file.c
@@ -1500,7 +1500,7 @@ static int bdb_flush_int(bdb_state_type *bdb_state, int *bdberr, int force)
     start = time_epochms();
     rc = ll_checkpoint(bdb_state, force);
     if (rc != 0) {
-logmsg(LOGMSG_ERROR, "txn_checkpoint err %d\n", rc);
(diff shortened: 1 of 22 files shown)


adizaimi:
We should have a knob to enable/disable ioless checkpoint, if possible. Also it would be nice to have a more descriptive name for 'perfect checkpoint'.
Thanks

riverszhang89:
Renaming it to ioless checkpoint?

markhannum:
Let's merge this & start testing- I have more to discuss in person.  Okay to merge.

riverszhang89:
Adi & Mark: Thanks for reviewing it.

PR #95, opened 2017-06-05
fix smoketest -- rpm and deb

adizaimi:


adizaimi:
Thanks. Waiting for roborivers.

PR #96, opened 2017-06-05
cdb2api Use our local timezone with localtime.

jmleddy:


cdb2api/cdb2api.c
```diff
@@ -1965,9 +1967,15 @@ static int cdb2_send_query(cdb2_hndl_tp *hndl, SBUF2 *sb, char *dbname,
 
     if (env_tz == NULL) {
         env_tz = DB_TZNAME_DEFAULT;
-    }
+    } else if (strcmp(env_tz, "localtime") == 0) {
```
mponomar: Make those calls here, once?

https://github.com/bloomberg/comdb2/blob/master/cdb2api/cdb2api.c#L95

I'm not a huge fan of calls that set state outside of comdb2's domain, but I guess this is the portable way to do it...

jmleddy: Yeah that makes more sense.

cdb2api/cdb2api.c
```diff
@@ -1965,9 +1967,15 @@ static int cdb2_send_query(cdb2_hndl_tp *hndl, SBUF2 *sb, char *dbname,
...
+        tzset();
+        int tzlen = strlen(tzname[0]) + strlen(tzname[1]) + 4;
+        int offset = timezone / (60 * 60);
 
-    sqlquery.tzname = env_tz;
+        local_tz = malloc(tzlen);
```
riverszhang89: `alloca` seems better than `malloc` here.

jmleddy:
The comdb2 server does not understand POSIX timezones. Before this can work, we would have to support POSIX timezones in the server. Alternately, we can cut the /usr/share/zoneinfo part of the localtime link.

PR #99, opened 2017-06-07
Master should run the query if client gives master node with direct cpu flag

mohitkhullar:
Giving master node in --host for cdb2sql should work.


adizaimi:
Looks good.

mohitkhullar:
thanks

PR #100, opened 2017-06-07
minor fixes related to running in cluster mode

adizaimi:
* Fix initial sending to master from master error
* cdb2api: if in direct cpu mode dbinfo shoould not get processed
   more trace in debug mode


riverszhang89:
@adizaimi Please take a look at https://github.com/bloomberg/comdb2/pull/101 whenever you have a chance.

Comparing to server sending back a real row which says "SSL_REQUIRE" and client unpacking the packet and then doing a string compare, my approach is much simpler: server sends back SQL_RESPONSE_SSL header with 0 payload, client initiates SSL if it sees SQL_RESPONSE_SSL .

cdb2api/cdb2api.c
```diff
@@ -1257,6 +1257,11 @@ static int try_ssl(cdb2_hndl_tp *hndl, SBUF2 *sb, int indx)
     cdb2_ssl_sess_list *store;
     SSL_SESSION *sess;
 
+    if (hndl->debug_trace) {
```
riverszhang89: I am not sure if we should add debug code here: It offers very limited information, and makes the code more difficult to read.

cdb2api/cdb2api.c
```diff
@@ -3267,34 +3304,44 @@ static int cdb2_run_statement_typed_int(cdb2_hndl_tp *hndl, const char *sql,
...
     if (type == RESPONSE_HEADER__DBINFO_RESPONSE) {
-        /* The master sent info about nodes that might be coherent. */
-        newsql_disconnect(hndl, hndl->sb, __LINE__);
-        CDB2DBINFORESPONSE *dbinfo_response = NULL;
-        dbinfo_response =
-            cdb2__dbinforesponse__unpack(NULL, len, hndl->first_buf);
-        parse_dbresponse(dbinfo_response, hndl->hosts, hndl->ports,
+        if (hndl->flags & CDB2_DIRECT_CPU) { 
+            /* direct cpu should not do anything with dbinfo */
+            PRINT_RETURN(-1);
```
riverszhang89: This is wrong: CDB2_DIRECT_CPU may have multiple hosts.

cdb2api/cdb2api.c
```diff
@@ -3267,34 +3304,44 @@ static int cdb2_run_statement_typed_int(cdb2_hndl_tp *hndl, const char *sql,
     }
 
     if (type == RESPONSE_HEADER__SQL_RESPONSE_SSL) {
+#if WITH_SSL
         hndl->s_sslmode = PEER_SSL_REQUIRE;
-        try_ssl(hndl, hndl->sb, hndl->connected_host);
+        /* server wants us to use ssl so turn ssl on in same connection */
+        int resp = try_ssl(hndl, hndl->sb, hndl->connected_host);
```
riverszhang89: We should always retry the query in case the node crashes during the connection attempt.

cdb2api/cdb2api.c
```diff
@@ -3072,6 +3104,11 @@ static int cdb2_run_statement_typed_int(cdb2_hndl_tp *hndl, const char *sql,
     int run_last = 1;
 
 retry_queries:
+    if (hndl->debug_trace) {
+        fprintf(stderr, "td %u %s line %d retry_queries: hndl->host %d (%s)\n", (uint32_t)
+                pthread_self(), __func__, __LINE__, hndl->connected_host, (hndl->connected_host>=0?hndl->hosts[hndl->connected_host]:""));
```
riverszhang89: Format.

cdb2api/cdb2api.c
adizaimi: Yeah, can remove.

cdb2api/cdb2api.c
adizaimi: Even if DIRECT_CPU has multiple hosts, receiving dbinfo from server should not be allowed because a) you already know the host list to try, and
b) you should not contact any other host which will be the case if you get and process dbinfo.

cdb2api/cdb2api.c
adizaimi: ok.

cdb2api/cdb2api.c
riverszhang89: Under this logic: If I have 2 hosts, and the 1st node goes down or incoherent, I lose the ability to retry on the 2nd node.

adizaimi: I just tested this, and it works: started a 3 node cluster, passed two of the nodes on the cmd line to an cdb2sql session interactive, did select comdb2_host() several times (got node1), killed node1 and reran select and it responded with node2. Unless I'm doing this wrong, it is working as I ment it. The two hosts in the cmd line both go in the local copy of dbinfo and when one is not avail, we try with the other.

roborivers: If you make master the 1st host in the list, does it still work? Keep in mind that we reach here only if we get `RESPONSE_HEADER__DBINFO_RESPONSE`. The retry logic upon database crashes is not handled by code here.

roborivers: Try this: enable MASTER_REJECT_REQUESTS, and make master the 1st node.

riverszhang89: Uh oh... Wrong account :disappointed:

adizaimi: If one of the two nodes is master say node1, your select comdb2_node gets node2. If you then kill node2, your next select returns 'cannot connect to db'. 
I assume you want in this case to process dbinfo?
Maybe only in --host mode should we reject dbinfo response.

adizaimi: My previous post was incorrect (i was using wrong port thats why could not connect). It can connect just fine to master--Mohit enabled master to serve when directcpu flag is on. So there are no issues with your scenario.

riverszhang89: Okay. If the connected node becomes incoherent and sends back dbinfo response, under your logic, do we retry on next node or return an error to application (assuming user has defined multiple hosts)?

adizaimi: If we get back dbinfo, we return error -- you are correct, that is not appropriate behavior, we should retry instead with next available node. I will change that. Thanks.

riverszhang89:
Changes look good. Please run `git diff master...<your_branch> -U0 | clang-format-diff -p1 -i` before merging.

PR #101, opened 2017-06-07
SSL: Reuse existing connections when server requires SSL.

riverszhang89:
Currently client APIs drop the plaintext connection and start a new SSL connection to server, if server requires SSL.
The SSL negotiation, however, can be more efficient by reusing the same connection.

### DIFF
--- a/cdb2api/cdb2api.c
+++ b/cdb2api/cdb2api.c
@@ -1231,7 +1231,7 @@ static int send_reset(SBUF2 *sb)
 }
 
 #if WITH_SSL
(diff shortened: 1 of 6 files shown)


adizaimi:
Change looks good, but I think Mohit can also take a look wrt the the new type we return for this.

riverszhang89:
@adizaimi Thanks for finding the bug & reviewing the PR. :+1:

riverszhang89:
Thanks Mohit & Adi!

PR #103, opened 2017-06-07
cdb2jdbc: Max query time.

riverszhang89:
`set timeout` is deprecated for protobuf-based APIs as of R6. Change Statement.setQueryTimeout() to use `maxquerytime` instead.


mohitkhullar:
looks good

riverszhang89:
Thanks Mohit!

PR #105, opened 2017-06-08
fix readonly select error when temptables are created

dorinhogea:
Queries that need to write to a temptable should be allowed.
Example:
"select * from T order by C limit N"


dorinhogea:
thanks

PR #107, opened 2017-06-08
reqlog Fix garbage on stdout.

jmleddy:
The problem was that instead of dump I had used logmsg before.

### DIFF
--- a/db/reqlog.c
+++ b/db/reqlog.c
@@ -1227,16 +1227,24 @@ int reqlog_logf(struct reqlogger *logger, unsigned event_flag, const char *fmt,
     return 0;
 }
 
-/* Log a string literal. Wrapper for logf */
+/* Log a string literal. */
 int reqlog_logl(struct reqlogger *logger, unsigned event_flag, const char *s)
 {
     if (logger && (logger->mask & event_flag)) {
-        if ((logger->dump_mask & event_flag) &&
-            ((logger->event_mask & event_flag) == 0)) {
-            /* Just dump, don't bother allocating */
-	  logmsg(LOGMSG_ERROR, NULL, s, strlen(s));
-        } else if (logger->event_mask & event_flag) {
-            reqlog_logf(logger, event_flag, s);
+        if (logger->event_mask & event_flag) {
+            struct print_event *event;
+            event = malloc(sizeof(struct print_event));
+            if (!event) {
+                logmsg(LOGMSG_ERROR, "%s:malloc failed\n", __func__);
+                return -1;
+            }
+            event->event_flag = event_flag;
+            event->length = -1; /* to indicate length is unknown */
+            strcpy(event->text, s);
+            reqlog_append_event(logger, EVENT_PRINT, event);
+        }
+        if (logger->dump_mask & event_flag) {
+            dump(logger, NULL, s, strlen(s));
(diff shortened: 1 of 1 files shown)


db/reqlog.c
```diff
@@ -1227,16 +1227,24 @@ int reqlog_logf(struct reqlogger *logger, unsigned event_flag, const char *fmt,
...
+        if (logger->event_mask & event_flag) {
+            struct print_event *event;
+            event = malloc(sizeof(struct print_event));
+            if (!event) {
+                logmsg(LOGMSG_ERROR, "%s:malloc failed\n", __func__);
+                return -1;
+            }
+            event->event_flag = event_flag;
+            event->length = -1; /* to indicate length is unknown */
+            strcpy(event->text, s);
```
adizaimi: Text is const char *text so you can't strcpy to it. Maybe strdup.

jmleddy: Yes good point.

adizaimi:
looks good. 
I think the way to test this one is:
'exec procedure sys.cmd.send("logmsg level debug")'
'exec procedure sys.cmd.send("reql results")'
'exec procedure sys.cmd.send("reql go")'
'select * from t1 limit 1'
and watch output in the db log should look something like:
2017/06/08 19:47:04 sql request 2 msec from ip-1-70-17-7 rc 0
2017/06/08 19:47:04   2 __memp_pg calls took 0 ms70-17-7 rc 0
2017/06/08 19:47:04   select * from t1 limit 1, ncols=2, rowcount=1, cost=20.000000
2017/06/08 19:47:04 row 1 = (1, x'1234')imit 1, ncols=2, rowcount=1, cost=20.000000
2017/06/08 19:47:04 ----------, x'1234')imit 1, ncols=2, rowcount=1, cost=20.000000

riverszhang89:
```
(gdb) bt
#0  __strcpy_sse2_unaligned () at ../sysdeps/x86_64/multiarch/strcpy-sse2-unaligned.S:636
#1  0x00000000004ff85a in reqlog_logl (logger=0x6363c50, event_flag=1, s=0x8fe02a "blockop") at db/reqlog.c:1243
#2  0x0000000000428aff in handle_ireq (iq=0x7f49657b6048) at db/sltdbt.c:324
#3  0x00000000004259df in thd_req (vthd=0x61c5470) at db/handle_buf.c:513
#4  0x000000000089f183 in thr_func (arg=0x608ab58) at bb/comdb2_pthread_create.c:231
#5  0x00007f4972e00064 in start_thread (arg=0x7f49657b2700) at pthread_create.c:309
#6  0x00007f4971cec62d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
(gdb) f 1
(gdb) p event->text
$3 = 0x0
```

db/reqlog.c
riverszhang89: @adizaimi https://github.com/bloomberg/comdb2/blob/master/db/reqlog_int.h#L43
text is not const.

adizaimi: Hmm, it was in one of the versions of that file which I saw. Sorry for the confusion, I don't know when it moved to the .h file.

adizaimi:
Check for s at func start?

PR #109, opened 2017-06-08
Fix build

mponomar:
Fixes package builds.
Fixes #15
Passes smoke test, and actual package installs.  Package installs were broken, but in a way that didn't trigger on smoke tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -128,6 +128,7 @@ pmux
 cdb2sql
 cdb2sockpool
 comdb2ar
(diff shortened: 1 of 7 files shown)


adizaimi:
looks good. 
Might I suggest a testcase of storing replay data and running cdb2_replay to a new db?

mponomar:
> Might I suggest a testcase of storing replay data and running cdb2_replay to a new db?
It's not remotely ready.  As soon as it is, of course.

PR #110, opened 2017-06-08
Add :Z option to docker-build to run in SELinux

Deon:
Adding the `:Z` option to `make docker-build`, so that the container can build in an environment with SELinux.

CC @mponomar


riverszhang89:
I don't think we should make `Z` the default option. Per https://github.com/moby/moby/issues/30934, docker developers are planning to deprecate the option entirely.

mponomar:
Run with `--privileged` instead?

PR #111, opened 2017-06-08
have debg reqlog be printed with LOG_USER level

adizaimi:
It currently is set to LOG_INFO and does not show up
with default logging level.

### DIFF
--- a/bb/logmsg.c
+++ b/bb/logmsg.c
@@ -48,6 +48,8 @@ static int level_to_syslog(loglvl lvl) {
             return LOG_DEBUG;
         case LOGMSG_INFO:
             return LOG_INFO;
+        case LOGMSG_USER:
+            return LOG_USER;
         case LOGMSG_WARN:
             return LOG_WARNING;
         case LOGMSG_ERROR:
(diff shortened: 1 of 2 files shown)


bb/logmsg.c
```diff
@@ -48,6 +48,8 @@ static int level_to_syslog(loglvl lvl) {
             return LOG_DEBUG;
         case LOGMSG_INFO:
             return LOG_INFO;
+        case LOGMSG_USER:
+            return LOG_USER;
```
riverszhang89: `LOG_USER` is a facility code. The function returns a priority code.

adizaimi: Sorry, I don't understand what that means. In logmsg.c we do the following: 
107         int syslog_level = level_to_syslog(lvl);
108         syslog(syslog_level, "%s", msg);
If lvl is LOGMSG_USER it will currently return critical, and that does not look right to me.

adizaimi: Ok, from man syslog: The priority argument is formed by ORing the facility and the level values. 
So SYS_USER is a facility...I understand now. Still, returning CRITICAL is probably not good idea, so should return something else, maybe LOG_INFO.

PR #116, opened 2017-06-12
now() cannot be passed blindly to remote session

dorinhogea:
Now() needs to be stable during a query run (multiple calls should return the same value).

### DIFF
--- a/sqlite/expr.c
+++ b/sqlite/expr.c
@@ -5665,15 +5665,19 @@ static char* sqlite3ExprDescribe_inner(
       if(strncasecmp(pExpr->u.zToken, "now", 3) == 0)
       { 
         dttz_t dt;
+        int prec;
+        
+        if(pExpr->x.pList && pExpr->x.pList->nExpr == 1) {
+           DTTZ_TEXT_TO_PREC(pExpr->x.pList->a[0].pExpr->u.zToken,
+                             prec, 0, break);
+        } else {
+          prec = gbl_datetime_precision;
+        }
+        
+        timespec_to_dttz(&v->tspec, &dt, prec);
 
-        if(pExpr->x.pList && pExpr->x.pList->nExpr == 1 &&
-           strncasecmp(pExpr->x.pList->a[0].pExpr->u.zToken, "us", 2) == 0)
-          timespec_to_dttz(&v->tspec, &dt, DTTZ_PREC_USEC);
-        else
-          timespec_to_dttz(&v->tspec, &dt, DTTZ_PREC_MSEC);
-
-        return sqlite3_mprintf("cast(%lld.%u as datetime)",  
-                               dt.dttz_sec, dt.dttz_frac);
+        return sqlite3_mprintf("cast(%lld.%*.*u as datetime)",  
+                               dt.dttz_sec, prec, prec, dt.dttz_frac);
       }
       /* default, pass the function remotely */
       if( pExpr->x.pList->nExpr <= 0 )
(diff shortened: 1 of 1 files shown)


sqlite/expr.c
```diff
@@ -5661,6 +5661,21 @@ static char* sqlite3ExprDescribe_inner(
     case TK_FUNCTION: {
       char *ret, *ret2;
       int i;
+      /* dh: not all the functions can be blindely passed remotely! */
+      if(strncmp(pExpr->u.zToken, "now", 3) == 0)
```
nirbhayc: Make this strncasecmp as well.
Also, the comparison should perhaps be on the length of zToken as strncmp("nowzzz", "now", 3) would also return 0 (unless parser catches it beforehand). Ditto for the cmp below.

dorinhogea:
"now" is a token already recognized by sqlite

nirbhayc:
@dorinhogea I see

nirbhayc:
Hit the wrong button. Reopening.

sqlite/expr.c
```diff
@@ -5661,6 +5661,21 @@ static char* sqlite3ExprDescribe_inner(
     case TK_FUNCTION: {
       char *ret, *ret2;
       int i;
+      /* dh: not all the functions can be blindely passed remotely! */
+      if(strncasecmp(pExpr->u.zToken, "now", 3) == 0)
+      { 
+        dttz_t dt;
+
+        if(pExpr->x.pList && pExpr->x.pList->nExpr == 1 &&
```
riverszhang89: The precision of `now()` depends on the default datetime precision. It is millisecond in code however can be changed via LRL.

Please use `DTTZ_TEXT_TO_PREC()` instead. Here is an example:
https://github.com/bloomberg/comdb2/blob/master/sqlite/dttz.c#L352

sqlite/expr.c
```diff
@@ -5661,6 +5661,21 @@ static char* sqlite3ExprDescribe_inner(
...
+           strncasecmp(pExpr->x.pList->a[0].pExpr->u.zToken, "us", 2) == 0)
+          timespec_to_dttz(&v->tspec, &dt, DTTZ_PREC_USEC);
+        else
+          timespec_to_dttz(&v->tspec, &dt, DTTZ_PREC_MSEC);
+
+        return sqlite3_mprintf("cast(%lld.%u as datetime)",
```
riverszhang89: :+1: Clever.

dorinhogea: What happens when a db is configured with MSEC and gets a USEC precision string?

riverszhang89: The database will happily use USEC. The default datetime precision rule only applies to `now()` (no arguments).

riverszhang89:
@roborivers is tired and needs a break.

sqlite/expr.c
```diff
@@ -5661,6 +5661,31 @@ static char* sqlite3ExprDescribe_inner(
...
+                     ||
+                     pExpr->x.pList->a[0].pExpr->u.iValue == DTTZ_PREC_USEC)) {
+            prec = pExpr->x.pList->a[0].pExpr->u.iValue;
+          }
+        }
+default_prec:
+        timespec_to_dttz(&v->tspec, &dt, prec);
+
+        return sqlite3_mprintf("cast(%lld.%*.*u as datetime)",  
+                               dt.dttz_sec, prec, prec, dt.dttz_frac);
```
riverszhang89: I don't think 0-padding the fractional seconds is correct. We're casting a double to datetime here. The fraction here means this much time in seconds.

riverszhang89: Oh wait. Dorin you are right. Please ignore my comment.

dorinhogea:
Thank you!

PR #119, opened 2017-06-12
check first, lock afterwards

dorinhogea:
Slightly improved versus R6 fix.  It handles the situation when we have a large number of sql sessions all accessing the same remote table for the first time.


dorinhogea:
thank you!

PR #124, opened 2017-06-13
call cdb2_close correctly

mohitkhullar:


adizaimi:
looks good.

mohitkhullar:
thanks

nirbhayc:
@mohitkhullar Also an assert((db != NULL) in cdb2_close()?

mohitkhullar:
@nirbhayc NULL is valid input to cdb2_close, its handled as no-op internally.

PR #125, opened 2017-06-13
get the quotes right for doc

adizaimi:
#skipbuild


riverszhang89:
Please #skipbuild it if documentation changes only.

PR #128, opened 2017-06-14
Fix path for bubble diagram generator.

mponomar:


mponomar:
Looks like this was done in #129

dorinhogea:
also addressed README src/sqlitegen path

PR #123, opened 2017-06-13
new syntax 'PUT TIME PARTITION nm RETENTION Y'

dorinhogea:
NOTE: the DDL success/abort callbacks are moved outside the bplog transaction.
This feature allows increasing the number of shards for a time partition.

PR #129, opened 2017-06-14
Revert "new syntax 'PUT TIME PARTITION nm RETENTION Y'"

dorinhogea:
Reverts bloomberg/comdb2#123


dorinhogea:
fixed more docs here

PR #131, opened 2017-06-14
Allow increase timepart retention 2

dorinhogea:
putting back the code I have accidentally removed.  now we should have both code and docs.


dorinhogea:
changes approved in https://github.com/bloomberg/comdb2/pull/123

PR #133, opened 2017-06-14
Fix slow datetime

mohitkhullar:
increases performance of queries with datetime fields by ~50%


dorinhogea:
This looks good

adizaimi:
The indenting looks off in this function tho.

PR #135, opened 2017-06-15
fix build issue

henrywoo:
to fix build error:

make: *** No rule to make target 'tools/cdb2_sqlreplay/cdb2_sqlreplay.c', needed by 'tools/cdb2_sqlreplay/cdb2_sqlreplay.o'.  Stop.

### DIFF
--- a/tools/module.mk
+++ b/tools/module.mk
@@ -30,12 +30,12 @@ cdb2sql: tools_LDLIBS+=$(LIBREADLINE)
 cdb2sql: $(cdb2sql_OBJS)
 	$(CC) $(tools_LDFLAGS) $^ $(tools_LDLIBS) -o $@
 
-cdb2replay_OBJS:=tools/cdb2_sqlreplay/cdb2_sqlreplay.o
+cdb2replay_OBJS:=cdb2_sqlreplay.o
 cdb2replay_CFLAGS=-Icson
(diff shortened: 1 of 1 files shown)


mponomar:
Thanks @henrywoo.  I can't reproduce that build error - does it work in the latest master?

nirbhayc:
@mponomar I did see this error for a specific commit (don't remember), but its fixed now.

henrywoo:
@mponomar  Can you reproduce it in Linux?(I am using Linux.)

henrywoo:
@mponomar it doesn't work in the latest master. but it did work last week. it's caused by the recent several commits related to cdb2_sqlreplay. My environment is Ubuntu 17.04.

jmleddy:
If I'm reading this right this would put the `.o` in the top level dir

jmleddy:
Also I think the build error is a race, or else has to do with .d, which is why it's harder to reproduce. Try doing a `git clean -Xi` and see if that doesn't resolve it.

henrywoo:
it works after git clean. Close it.

mponomar:
For anyone interested:  cdb2_sqlreplay used to be a .c, and got re-implemented as a .cpp.  The autogenerated dependencies didn't pick that up.  `make clean` doesn't clean dependency information, but `git -dfx` does.

PR #136, opened 2017-06-15
Don't grab the bdb lock in trigger_unregister_node

markhannum:
Getting the bdb-lock here can cause an abort if the pthread_key holding the lock isn't set yet.  My testing showed that this occurs in connection-refresh code.  Holding the bdb-lock in trigger_register_node prevents the master from swinging, but holding it here is unnecessary.

### DIFF
--- a/db/trigger.h
+++ b/db/trigger.h
@@ -33,7 +33,6 @@ void trigger_start(const char *);
 int trigger_register_req(trigger_reg_t *);
 int trigger_unregister_req(trigger_reg_t *);
 int trigger_unregister_node(const char *node);    // will get bdblock
-int trigger_unregister_node_lk(const char *node); // get bdblock, then call
 int trigger_registered(const char *);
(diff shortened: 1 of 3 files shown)


adizaimi:
I saw a similar deadlock on master in the following scenario:
1) master did schemachange and waits for replicant to apply 
2) replicant crashes while performing change
3) master attempts to disconnect replicant node.
(I will attempt to replicate the issue--lost the stack trace).

db/trigger.h
```diff
@@ -33,7 +33,6 @@ void trigger_start(const char *);
 int trigger_register_req(trigger_reg_t *);
 int trigger_unregister_req(trigger_reg_t *);
 int trigger_unregister_node(const char *node);    // will get bdblock
```
akshatsikarwar: Fix comment

akshatsikarwar:
Looks good otherwise

PR #137, opened 2017-06-15
Fix replay: snapinfo_outrc is definitive

markhannum:
I'm leaving the log-message in place as it's useful for identifying bugs for clients which do not generate a snapinfo blkseq.


markhannum:
Withdrawing for now: Adi pointed out that this code is a mess & doesn't do what I thought it did.

PR #138, opened 2017-06-15
cdb2jdbc: Use % specifiers instead of string concatenation for log messages.

riverszhang89:
String concatenation is expensive and right now we do it without knowing if it is actually needed or not.
It can be avoided using string formatter: The string processing is done only when the given log level
needs to be logged.

### DIFF
--- a/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
+++ b/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
@@ -89,7 +89,7 @@ public class Comdb2Handle extends AbstractConnection {
     private boolean isRead;
     private String lastSql;
     private long rowsRead;
-    private byte[] uuid;
+    private String uuid;
     private String stringCnonce;
     private byte[] cnonce;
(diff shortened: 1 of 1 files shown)


riverszhang89:
> should we consider only making the logging call if we know we're at the correct log level?`

It is done. See https://github.com/bloomberg/comdb2/pull/138/files#diff-9df1523253533f004529205d90e9f2aeR277

Using replaceable parameters brings us huge performance gains. I wrote a small JDBC SELECT * program to fetch data from a table which has roughly 20 million rows. Here are the numbers I've got:

- With master: 0m58.856s
- With the patch: 0m31.857s

Off-topic: I also wrote a C version. It took 0m30.670s - almost the same amount of time as the patched cdb2jdbc.

[JDBC code](https://gist.github.com/riverszhang89/db2086e15db4937bbb6c62b763fba67c)
[C code](https://gist.github.com/riverszhang89/28e30560666e97ee3ef7e1534d9d1c6d)

mponomar:
Hah!  Sorry, I misread the patch.

riverszhang89:
No worries! Thanks for reviewing it.

PR #97, opened 2017-06-06
disable skip scan for remote sql

dorinhogea:
The skip scan is almost always more expensive for remote sql, as shown in a few production runs.  This change will disable it.

PR #139, opened 2017-06-15
disable skipscan by means of put commands

adizaimi:
* can be used in this way in cdb2sql:
  put skipscan enable t1
  put skipscan disable t1
  Note: enable does not work because of code added recently in https://github.com/bloomberg/comdb2/pull/97 (once disabled, it stays disabled).
* Evoke callback for replicants and master to reload analyze stats, to reeval the disable skipscan options.


akshatsikarwar:
I prefer if this is done using bpfunc and use GET/PUT. This way its SQL and can be run on anynode.

db/appsock_handler.c
```diff
@@ -788,6 +786,7 @@ static void *thd_appsock_int(SBUF2 *sb, int *keepsocket,
                 sbuf2printf(sb, ">disableskipscan cleared\n");
             sbuf2printf(sb, "SUCCESS\n");
             sbuf2flush(sb);
+            bdb_llog_analyze(thedb->bdb_env, 1, &bdberr);
```
ldennis: is this always on master?

adizaimi: Good point. It may not--adding check. Thanks.

akshatsikarwar:
Why is there send ? Only get put.

adizaimi:
Akshat, there is no send cmd (I removed it). I added long ago to R6 ability to do: 
   comdb2sc db disableskipscan t1
I could remove it since nothing is using it yet (PR to comdb2sc is pending), and backport put cmd to R6.
I did not add a put cmd -- thought it was useless, just use send("tableparams").

tests/analyze.test/t10_01.req.out
```diff
@@ -1 +0,0 @@
-SUCCESS
```
ldennis: Why did you disable my analyze concurrent test??

adizaimi: Sorry, bad merge. Will put back asap.

PR #145, opened 2017-06-16
Baby Transactional DDL

ldennis:
First cut work: (I dont want to keep this too big)
- schemachange code refactoring (and re formatting)
- baby transactional ddl support -- mix of ddl + dml on non-overlapping tables
    (replicants sides still need more work)
eg
```
begin
create t1
alter t2
truncate t3
drop t4
update t0
commit
```
- I didnt do any foreign-key related coding yet (i.e. adding multiple tables transactionally with circular dependencies would fail)


akshatsikarwar:
Can you interactive rebase this to have 3 commits - initial and all of your code changes and tests. It'll make review easier for me and look decent in commit history.

bdb/attr.h
```diff
@@ -352,7 +352,7 @@ DEF_ATTR(PLANNER_WARN_ON_DISCREPANCY, planner_warn_on_discrepancy, BOOLEAN, 0)
 DEF_ATTR(PLANNER_EFFORT, planner_effort, QUANTITY, 1)
 DEF_ATTR(SHOW_COST_IN_LONGREQ, show_cost_in_longreq, BOOLEAN, 1)
 /* Delay restarting SC by this many seconds from db up*/
-DEF_ATTR(SC_RESTART_SEC, sc_restart_sec, QUANTITY, 60)
+DEF_ATTR(SC_RESTART_SEC, sc_restart_sec, QUANTITY, 0)
```
adizaimi: I'm worried that making this 0, may result in some unexpected side effects with the db not being up. One example is fastinit calls autoanalyze_table at the end to zero out analyze counters, which may fail if the db is not up yet. We should test plenty.

ldennis: This was only used by resume_schema_change() for resuming a table alter (in function wait_to_resume()), no?

schemachange/module.mk
```diff
@@ -17,7 +17,8 @@ schemachange/sc_records.o	\
 schemachange/sc_schema.o	\
 schemachange/sc_stripes.o	\
 schemachange/sc_struct.o	\
-schemachange/sc_util.o
+schemachange/sc_util.o      \
+db/comdb2uuid.o
```
akshatsikarwar: Why is this here?

ldennis: it uses that to check uuid in resume schema change.. and libschemachange.a needs it

akshatsikarwar: Yes, by why do you need the obj file in this library? Just the include should be sufficient and finally the linker will find and resolve references when building `comdb2` binary.

akshatsikarwar:
Can you resubmit for testing with last changes?

akshatsikarwar:
Looks good to me. If passing tests, plz go ahead and merge. As discussed, keep working on cleaning up drop table.

schemachange/module.mk
ldennis: oh.. this was for old tools.. cleaning up

bdb/attr.h
adizaimi: Yes, used to resume schemachange after bringing up db in single node.

akshatsikarwar:
Is this compatible with R6 schemachanges?

akshatsikarwar:
Just thinking ahead when we deploy and have mix of 6&7

ldennis:
Short answer is no. Because I write extra fields (rqid+uuid) for schemachange into llmeta. maybe later we can introduce an option to disable writing those or an option to disable the resume_schema_change code?

ldennis:
Only resume schemachange logic is not backwards compatible. others should be fine

schemachange/sc_drop_table.c
```diff
@@ -45,7 +45,7 @@ static int delete_table(struct db *db, void *trans)
     remove_constraint_pointers(db);
 
     int rc, bdberr;
-    if ((rc = bdb_close_only_tran(db->handle, trans, &bdberr))) {
+    if ((rc = bdb_close_only_tran(db->handle, NULL, &bdberr))) {
```
akshatsikarwar: Not using `trans`? also it should be `tran_type` not `void`.

ldennis: It just closes the handle. I thought you said close doesn't need tran. And if the whole thing fails later, we will reopen.

ldennis: And this is same as your initial version

akshatsikarwar: It didn't.. You added it, no? Should be consistent, in any case. Cleanup to remove needing it, or use it everywhere.

ldennis: Ok I will change them back to not taking tran

akshatsikarwar:
Can rebase to make pretty. If it passes test, this is ok to merge.
In separate PR keep working on making compatible with R6 schemachange. Its ok to fail to resume a schemachange during transition. But should have option to do a schemachange whether R6 is master or replicant. Option will be disabled by default. When deploying, we'll enable so R6->7 can happen.

ldennis:
rebased again to make history pretty..

PR #148, opened 2017-06-16
Clarify a point about growing a cluster.

### LINKED ISSUE #147: adding/removing nodes from cluster
the documentation leaves out a lot of details about this.
when a node is added/removed from a cluster:

1. do i need to update the config of every single other node in the cluster with the updated node's hostname?
2. do all the other nodes need to be rolling restarted to read this new config?
3. does all the client code need to be updated too to add this node's hostname to their driver initialization call?

mponomar:
Fixes #147


riverszhang89:
Please #skipbuild for documentation changes only.

PR #152, opened 2017-06-17
Instrument and fix blkseq replay and durable-lsn return codes

markhannum:
* Adds the test_blkseq_replay switch to randomly return NOT_DURABLE from the master
* Fix blkseq-replay case
* Return SQLITE_CLIENT_CHANGENODE from the replicant for the NOT_DURABLE case


adizaimi:
Looks ok.

PR #153, opened 2017-06-18
Register test from the jepsen testsuite

markhannum:
Ported Kyle Kingsbury's register test to comdb2.  The test has 10 threads running READ, WRITE, and COMPARE-AND-SET operations against a single row in the database, while another thread introduces random network partitions against the cluster.  My changes included some local changes to Kyle's jepsen and knossos code, so I'm now including these (with my changes) in our repository.

I've tested this fairly exhaustively this weekend against a 5-node AWS cluster.  Comdb2 passes every time.  Future work (possibly soon) will be to write a C version of this test.


db/comdb2.c
```diff
@@ -2304,6 +2305,12 @@ static int read_lrl_option(struct dbenv *dbenv, char *line, void *p, int len)
         logmsg(LOGMSG_INFO,
                "Tables will be initialized with instant schema-change support\n");
         gbl_init_with_instant_sc = 1;
+    } else if (tokcmp(line, ltok, "dont_init_with_compr") == 0) {
```
akshatsikarwar: Why do we need this? Just add `init_with_compr none` to lrl or db which may require it.

akshatsikarwar: Also, why does register test need compression disabled?

markhannum: Thanks Akshat- yeah, it's redundant.  I need it disabled to be able to read logfiles easily.

markhannum:
Withdrawing- there's a bug in my clojure code: it's not reporting the database state correctly to knossos.  I'll submit again after retesting.

PR #155, opened 2017-06-19
Increase linux performance

mohitkhullar:
by 5-10% for linux server, extra 5-10%  if client is also on linux


jmleddy:
How do we know any of this improves performance?

mohitkhullar:
I benchmarked it against queries with datetime/integer index & fields.

jmleddy:
Oh sorry I didn't see the first comment.

dorinhogea:
please collapse those commits into one

PR #156, opened 2017-06-19
Logging: Convert all times in log to microseconds.

riverszhang89:
Continue reporting LONG EVENTS, etc. as we do now, but report microseconds for event logging.

### DIFF
--- a/bdb/bdb_api.h
+++ b/bdb/bdb_api.h
@@ -138,12 +138,12 @@ struct bdb_queue_stats {
 
 /* Multiplication usually takes fewer CPU cycles than division. Therefore
    when comparing a usec and a msec, it is preferable to use:
-   usec <comparison operator> USEC(msec)  */
-#ifndef MSEC
-#define MSEC(usec) ((usec) / 1000)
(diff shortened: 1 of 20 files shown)


bdb/bdb_api.h
```diff
@@ -135,29 +135,39 @@ struct bdb_queue_stats {
 };
 
 /* This is identical to bb_berkdb_thread_stats in db.h */
+
+/* Multiplication usually takes fewer CPU cycles than division. Therefore
+   when comparing a usec and a msec, it is preferable to use:
+   usec <comparison operator> USEC(msec)  */
+#ifndef MSEC
+#define MSEC(usec) ((usec) / 1000)
```
adizaimi: Suggestion: May it help if we rename this USEC_to_MSEC so its immediately clear what it does? (and USEC_TO_MSEC() below)

riverszhang89: Good point.

riverszhang89:
Thank you Mike.

riverszhang89:
Thank you Adi.

PR #160, opened 2017-06-20
improve help for cdb2sql

adizaimi:


riverszhang89:
Wait.. roborivers did say the format was incorrect...

adizaimi:
yes, apparently the \ at the end are not necessary.

PR #161, opened 2017-06-20
Provide way to run unit tests without running a db

adizaimi:
By setting global variable COMDB2_UNITTEST in the makefile, we can run
a designated unit test [or tests] without running a backing db.
Convert crle test to use such feature.


tests/testcase.mk
```diff
@@ -12,6 +12,7 @@ export CDB2_OPTIONS=--cdb2cfg $(CDB2_CONFIG)
 export COMDB2_ROOT=$(TESTDIR)
 export comdb2ar=$(SRCHOME)/comdb2ar
 export comdb2task=$(SRCHOME)/comdb2
+export COMDB2_UNITTEST?=0
```
timmui: `stop` target is still calling `unsetup` without checking for the flag (Github is not letting me add a line comment on that line)

adizaimi: Thanks. Added check in setup/unsetup like you had it in your branch.

PR #162, opened 2017-06-20
Column names should be case insensitive.

sblind1:
Column names are currently case sensitive in ResultSet implementation. This fix will aloow to look up column via case insensitive names.


riverszhang89:
Nifty - Thanks!

PR #163, opened 2017-06-21
comdb2ar and cdb2_sqlreplay have C++ mains and are not aliased to comdb2

markhannum:
Redoing an accidentally undone change


markhannum:
Mike says good

PR #112, opened 2017-06-09
Initial signed int overflow detection

timmui:
Initial implementation of a set of utility functions for checking signed integer overflow when certain operations are performed.

PR #166, opened 2017-06-21
Signed int overflow lib changes and tests

timmui:
* Fixes from #112 and unit tests for the signed integer overflow library
* Clarified library and changed return values (breaking change)
* Added unit tests

### DIFF
--- a/bbinc/int_overflow.h
+++ b/bbinc/int_overflow.h
@@ -20,10 +20,10 @@
 extern "C" {
 #endif
 
-int check_overflow_ll_add (long long a, long long b);
-int check_overflow_ll_sub (long long a, long long b);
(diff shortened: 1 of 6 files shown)


timmui:
@markhannum

bbinc/int_overflow.h
```diff
@@ -20,10 +20,10 @@
 extern "C" {
```
nirbhayc: I am wondering if this is really needed.

nirbhayc:
Looks good.

bbinc/int_overflow.h
timmui: Not sure. I've seen other files in `/bbinc` with it so I did the same

markhannum:
agree - looks great

PR #169, opened 2017-06-22
On startup, wait for coherent before "I AM READY"

adizaimi:
Currently when a node is coming up, it prints 'I AM READY' before it can serve clients because it may not be coherent yet. This patch will have a node wait until it becomes coherent before it proceeds to print it is ready.

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -8574,7 +8574,7 @@ void create_marker_file()
     if (tmpfd != -1) close(tmpfd);
 }
 
-void set_timepart_and_handle_resume_sc() 
+static void set_timepart_and_handle_resume_sc() 
 {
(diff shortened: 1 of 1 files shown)


nirbhayc:
LGTM

db/comdb2.c
```diff
@@ -8622,6 +8622,20 @@ struct tool tool_callbacks[] = {
    NULL
 };
 
+
+void wait_for_coherent()
+{
+    const unsigned int cslp = 10000;         // 10000us == 10ms
+    const unsigned int wrn_cnt = 5 * 1000000 / cslp; // 5s
```
mohitkhullar: warning won't come after 5s here.

PR #171, opened 2017-06-22
db: Providing an LRL option to control VDBE sorter virtual memory usage

riverszhang89:
By default, sqlite VDBE sorter maps up to 2GB of the process's virtual memory to temporary files.
Comdb2 is multi-threaded. Each SQL thread has its own VDBE. Hence it could cause total database size to exceed allowable `ulimit -v'.
We should at least have a tunable to control VDBE virtual memory usage.

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -2648,6 +2648,15 @@ static int read_lrl_option(struct dbenv *dbenv, char *line, void *p, int len)
         gbl_sqlite_sorter_mem = ii;
     }
 
+    else if (tokcmp(tok, ltok, "sqlsortermaxmmapsize") == 0) {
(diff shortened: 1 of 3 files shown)


ldennis:
Looks like this is not just for sorters but a global config for SQLite. Would it be better if the tunable's name is `sqlitermaxmmapsize`? Otherwise, looks good to me.

riverszhang89:
Thanks for reviewing it. The config is indeed a size cap for all database files in stock sqlite. However we only utilize vdbe sorter temporary files. Therefore the name `sqlsortermaxmmapsize` makes more sense for us.

ldennis:
sounds reasonable

riverszhang89:
Thanks Lingzhi.

PR #172, opened 2017-06-23
Jepsen register test

markhannum:
Port of Kyle Kingsbury's CAS-Register test with a database fix.  This pr fixes an issue where serializable transactions skip backfill entirely, and are therefore unable to perform either point-in-time, HASql reconnects, or durable-lsn unwinding correctly.  The bug does not affect vanilla serializable isolation where the snapshot begins at the current end-of-log.

The CAS-Register test is designed to verify that a distributed system is linearizable in the face of network partitions.  I have been running the CAS-Register test against this fix on a 5-node AWS cluster under this version of the code for the past 24 hours- we are passing successfully.  I will continue to test this over the next few days in parallel with the code review.

Future work will be to port this test to C and Java.


markhannum:
To Mike: he knows Clojure.

riverszhang89:
Please format before merging.
http://comdb2.s3-website-us-east-1.amazonaws.com/formatting-a-pr.html

PR #173, opened 2017-06-23
analyze empty tables

ldennis:
- sqlite patch: https://www.sqlite.org/src/fdiff?sbs=1&v1=0d0ccf7520a201d8&v2=62a35801482e72fa
- I set the tunable to OFF so that it wont break existing tests which assume no entry for empty tables.
- added a simple test case to analyze.test


adizaimi:
Looks good to me.

PR #174, opened 2017-06-23
do not try to send information if recovery mode

dorinhogea:
Database should not create communication threads if it will come down immediately.  Otherwise it is spewing like this:
[ERROR] net_udp_send:sendto: Socket operation on non-socket
dest=fakehost, addr=0.0.0.0


adizaimi:
Should we leave the watchdog thread to run? Or it will do nothing just like the others.
Looks fine to me otherwise.

PR #175, opened 2017-06-23
Change comdb2ar to copy btree files aggressively

markhannum:
Comdb2ar previously relied on file_vers_map.  This code no longer works as comdb2sc has been deprecated.  Actually, there are several situations where this code never worked.  Consider the case where the database is not running (and cannot respond to the comdb2sc command), or the case where a dropped table is still referenced in the database's logs.

The solution is to copy everything that looks like a btree.  The downside is that it is possible that we'll serialize more than we need .. the upside is that the code works correctly.


markhannum:
Reviewed with Mike

PR #179, opened 2017-06-25
Delete deprecated functions from sqlanalyze.c

adizaimi:
Removing deprecated functions run_sql() and analyze_table_int_old() which have not been used in a long time but were kept for reference.


ldennis:
Looks good

PR #180, opened 2017-06-25
CSC2less DDL

nirbhayc:


sqlite/comdb2build.c
```diff
@@ -1675,3 +1673,2269 @@ void comdb2WriteTransaction(Parse *pParse)
...
+    COMDB2_TYPE("decimal32", "decimal32", 0),
+    COMDB2_TYPE("decimal64", "decimal64", 0),
+    COMDB2_TYPE("decimal128", "decimal128", 0),
+    COMDB2_TYPE("float", "float", 0), COMDB2_TYPE("double", "double", 0),
+
+    /* Additional types mapped to a Comdb2 type. */
+    COMDB2_TYPE("varchar", "cstring", COMDB2_TYPE_FLAG_ALLOW_ARRAY),
+    COMDB2_TYPE("char", "cstring", COMDB2_TYPE_FLAG_ALLOW_ARRAY),
+    COMDB2_TYPE("text", "vutf8", COMDB2_TYPE_FLAG_ALLOW_ARRAY),
+    COMDB2_TYPE("date", "datetime", 0)};
```
dorinhogea: lets not map date to datetime

sqlite/comdb2build.c
```diff
@@ -1675,3 +1673,2269 @@ void comdb2WriteTransaction(Parse *pParse)
...
+void comdb2AlterTableStart(
+    Parse *pParse, /* Parser context */
+    Token *pName1, /* First part of the name of the table. */
+    Token *pName2, /* Second part of the name of the table. */
+    int dryrun     /* Whether its a dryrun? */
+    )
+{
+    struct comdb2_ddl_context *ctx;
+
+    if (isRemote(pName1, pName2)) {
```
dorinhogea: "main"."tablename" is still local, i don't think isRemote is correct.

nirbhayc: I see. I simply reused isRemote(). In your comment, does "main" have a special meaning or it's just the database name?

sqlite/comdb2build.c
nirbhayc: Ok, will do.

sqlite/comdb2build.c
dorinhogea: "main" is special, it is always the local db.  Also someone could use the dbname.tblname, where dbname is the local db name.

nirbhayc:
@dorinhogea Could you please have a look at the last 2 patches?

sqlite/comdb2build.c
```diff
@@ -55,15 +55,45 @@ int readIntFromToken(Token* t, int *rst)
...
+    /* Check if the second token is empty. */
+    if ((*t2)->n == 0) {
+        /* No schema name. */
+        return 0;
+    } else {
+        /* We must have both the tokens set. */
+        assert((*t1)->n > 0 && (*t2)->n > 0);
+
+        /* t1 must be a local schema name. */
+        if ((strncasecmp((*t1)->z, thedb->envname, (*t1)->n) == 0) ||
```
dorinhogea: This looks wrong. Second name  (t2) is the database name; t1 is the table name.

nirbhayc: @dorinhogea : The format used in SQL is either
```
(1) <schema>.<table|index> or
(2) <table|index>
```
In case (1), token t1 holds the schema name while in case (2), the same token t1 holds the table or index name. There's a check at the top of this function that detects whether schema name is supplied. If it is, then token t1 will always have the schema name.
```
static inline int isRemote(Parse *pParse, Token **t1, Token **t2)
{
    /* Check if the second token is empty. */
    if ((*t2)->n == 0) {
        /* No schema name. */
        return 0;
    } else {
```

akshatsikarwar:
This is passing all tests now? Can we enable stock SQLite tests from `yast` and friends?

nirbhayc:
@akshatsikarwar just retriggered the tests.

nirbhayc:
@akshatsikarwar stock sqlite tests == yast_r7.test ?

akshatsikarwar:
Yes. How do you plan on testing this? `yast`'s `tester.tcl` converts SQLite DDL to csc2. We could try to reuse those tests. Although they are not directly usable because they rely on SQLite-ism of not requiring column types. Any other test suite you can think of?

nirbhayc:
Yes, I look into those and stopped after noticing type-less columns. I have, however, added tests/ddl_no_csc2.test, which covers most of the new constructs.

akshatsikarwar:
Nice. Expression index works?

nirbhayc:
Not yet.

PR #183, opened 2017-06-27
Hasql fix

markhannum:
Use client-side supplied snapshot information at transaction start.  Use the LSN of the commit record rather than the prev-record to decide whether it should be part of backfill.

### DIFF
--- a/berkdb/rep/rep_record.c
+++ b/berkdb/rep/rep_record.c
@@ -5120,23 +5120,6 @@ get_committed_lsns(dbenv, inlsns, n_lsns, epoch, file, offset)
                     break;
                 }
 
-                if (txn_rl_args->prev_lsn.file < file ||
-                    (txn_rl_args->prev_lsn.file == file &&
-                     txn_rl_args->prev_lsn.offset <= offset)) {
-                    if (gbl_extended_sql_debug_trace) {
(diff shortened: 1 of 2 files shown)


berkdb/rep/rep_record.c
```diff
@@ -5089,108 +5089,89 @@ get_committed_lsns(dbenv, inlsns, n_lsns, epoch, file, offset)
...
+                    txn_rl_args->lflags & DB_TXN_LOGICAL_COMMIT) {
+                    if (*n_lsns + 1 >= curlim) {
+                        curlim = (!curlim) ? 1000 : 2 * curlim;
+                        if (!(newlsns = (DB_LSN *)realloc(
+                                  lsns, curlim * sizeof(DB_LSN)))) {
+                            logmsg(LOGMSG_ERROR, "%s:%d Too complex "
+                                                 "snapshot (realloc "
+                                                 "failure at trns %d)\n",
+                                   __FILE__, __LINE__, *n_lsns);
+                            ret = 22
```
markhannum: Make this ENOMEM

adizaimi:
Mark, were you able to have a testcase that would reproduce this bug? It would be nice to catch it with a test.

markhannum:
I'm writing a C version of the register test in a different branch.  I'm hoping that the C test will reproduce the issue faster than the clojure one.

PR #192, opened 2017-06-29
dont exit cdb2sql on ctrl_c

adizaimi:
Ctrl_C will clean out the current input line.
Ctrl_d will exit if input line is blank.
Does not cancel currently running sql yet.

### DIFF
--- a/tools/cdb2sql/cdb2sql.c
+++ b/tools/cdb2sql/cdb2sql.c
@@ -83,6 +83,10 @@ static int istty = 0;
 static char *gensql_tbl = NULL;
 static char *prompt = main_prompt;
 
+
+int sigaction(int signum, const struct sigaction *act,
+                     struct sigaction *oldact);
+
 static void hexdump(FILE *f, void *datap, int len)
 {
     u_char *data = (u_char *)datap;
@@ -887,6 +891,26 @@ static void replace_args(int argc, char *argv[])
     }
 }
 
+
+static int int_handler_cnt = 0;
+
+/* If ctrl_c was pressed to clear existing line and go to new line
+ * If we see two ctrl_c in a row we exit.
+ * However, after a ctrl_c if user typed something 
+ * (rl_line_buffer is not empty) and then issue a ctrl_c then dont exit.
+ */
+static void int_handler(int signum)
+{
+    if(strlen(rl_line_buffer) > 0) 
+	int_handler_cnt = 0;
+    if (++int_handler_cnt > 1) 
+        exit(1);
+    printf("\n");
+    rl_on_new_line();
+    rl_replace_line("", 0);
+    rl_redisplay();
+}
+
 int main(int argc, char *argv[])
 {
     static char *filename = NULL;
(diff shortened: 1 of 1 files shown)


tools/cdb2sql/cdb2sql.c
```diff
@@ -83,6 +83,10 @@ static int istty = 0;
 static char *gensql_tbl = NULL;
 static char *prompt = main_prompt;
 
+
+int sigaction(int signum, const struct sigaction *act,
+                     struct sigaction *oldact);
```
nirbhayc: Formatting?

tools/cdb2sql/cdb2sql.c
```diff
@@ -887,6 +891,26 @@ static void replace_args(int argc, char *argv[])
     }
 }
 
+
+static int int_handler_cnt = 0;
+
+/* If ctrl_c was pressed to clear existing line and go to new line
```
nirbhayc: CTRL+C should make the control jump to the next (empty) prompt, after printing C^.  2 CTRL+C's should behave in a similar way, just jump an additional prompt.
CTRL-D, OTOH should kill cdb2sql iff the input buffer is empty, a NOOP otherwise.

nirbhayc: This is how bash behaves, and thus, most likely what users would expect from cdb2sql.

akshatsikarwar: > CTRL-D, OTOH should kill cdb2sql iff the input buffer is empty, a NOOP otherwise.

`dbname> blah blah CTRL-D` shouldn't be NOP. Its should be as if `dbname> blah blah ENTER`.

nirbhayc: This is how bash behaves.

adizaimi: ctrl-d is same as ctrl-m (enter)?

akshatsikarwar: only if there is input on line

akshatsikarwar: if cmdline is empty then C-d is exit

akshatsikarwar: doesn't readline already do this for us?

akshatsikarwar:
> This is how bash behaves.

Nah - I just tried it.

tools/cdb2sql/cdb2sql.c
adizaimi: Is that something that you see in any other readline type prompt? Bash certainly does not do anything on ctrl-d (when line has characters). Maybe PG or mysql do it?

akshatsikarwar: Doing it for me..

akshatsikarwar:
Confirmed - readline is already doing the right thing for EOF. Just handle C-c.

akshatsikarwar:
Discrepancy was due to vi keybindings (readline should process same .inputrc and behave accordingly).

PR #193, opened 2017-06-29
Eventlog: Print cnonce for txn, print deadlock loop participants

adizaimi:
* always pass cnonce to master node, save cnonce in thread specific variable
* print cnonce for sql and txn types in events log
* print deadlock cycle participants with cnonce information
* add testcase to get two transactions to deadlock and get cycle reported in events log
* reql stat to show events log status and file info

Eventlog for the deadlock loop will appear only if setting is turned on:
 `on print_deadlock_cycles`
Eventlog entry will look like this:
 `{"time": 1499873882991794,"host": "node1","deadlock_cycle": [{"cnonce": "8323329-27057-959039-1013043383","lid": "0x80000012","lcount": 5},{"cnonce": "8323329-27056-958938-1013043383","lid": "0x80000010","lcount": 5,"victim": true}]}`


dorinhogea:
why are we skipping regression tests for code changes?

adizaimi:
Dorin, it is not ready yet, thus did not want to have testsuite run on this. Will remove the #skip when it's ready to be tested, hopefully very soon. Thanks.

adizaimi:
simple_timepart and sc_timepart passed in my laptop. Will resubmit, then merge master into my branch.

dorinhogea:
why added the commits here already in the master?

adizaimi:
Wow, dont know why it looks like that, but I rebased from origin and i get this crap. I can try again from a better copy.

db/toblock.c
```diff
@@ -5418,12 +5418,20 @@ static int toblock_main_int(struct javasp_trans_state *javasp_trans_handle,
...
-            int t = time_epoch();
-            memcpy(p_buf_fstblk, &t, sizeof(int));
-            rc = bdb_blkseq_insert(thedb->bdb_env, parent_trans, bskey,
-                                   bskeylen, buf_fstblk,
-                                   p_buf_fstblk - buf_fstblk + sizeof(int),
-                                   &replay_data, &replay_len);
+            extern int gbl_always_send_cnonce;
+            // if RC_INTERNAL_RETRY && replicant_can_retry don't add to blkseq
+            if (outrc == ERR_BLOCK_FAILED && err.errcode == ERR_VERIFY && 
+                (gbl_always_send_cnonce && iq->snap_info.replicant_can_retry)) {
```
markhannum: Use iq->have_snap_info instead of gbl_always_send_cnonce for this check .. it's unlikely, but possible to toggle gbl_always_send_cnonce to false while the block processor is executing.  This scenario would allow a verify-error to be committed to blkseq.

markhannum:
Wow .. this is fantastic.  We can trace deadlock cycles all the way back to client requests.  There was one change I mentioned in toblock.c - okay to merge after that.

We'll have to discuss rollout strategies, and backwards compatibility with R6, but I think we should do that under a different pull request.

db/toblock.c
markhannum: Make this a "review"

adizaimi:
Thank you Mark. This was longtime in waiting and its finally in.

PR #201, opened 2017-06-30
Tunables refactoring

nirbhayc:
* Unify all tunables into a global list
* Attach properties to all the tunables (like type, readonly, etc.)
* Provide a common interface to update tunables
* comdb2_tunables virtual table
* Introduce : PUT TUNABLE <name> <value> to update tunables at run time
* Preserve the old methods of updating tunables at runtime


dorinhogea:
How do you check the setting of a tunable when looking at a core file?

nirbhayc:
@dorinhogea All tunables are linked via a list thedb->tunables.

dorinhogea:
It might be difficult to find it, how many do we have?

dorinhogea:
Maybe have their names as in index in a global array instead of linked list?

nirbhayc:
A little over 800.

adizaimi:
Array idea seems quite nice, the offset can be the enum VALUE_OF_OPTION_X

nirbhayc:
I too like the array approach, but the problem is we do not have the total tunables' count (including ones from all the sub-modules), this will require us to do reallocs, which in turn would ask for regeneration of the tunables hash.
Another future problem that I see is, this will make it difficult to support dynamic addition/removal of tunables (say if in future we decide to add support to install/uninstall plugins that have their own tunables) .

Also generating enum for all tunables from different modules seem ugly to me. I will leave it on to the hash to detect duplicates.

dorinhogea:
Realloc is not a problem, hash them by their string "name".  No need to rehash when adding more tunables.  Associate a define name matching the string, with a constant value that matches the array index.  Ideally this should be auto-generated with some macros.  Hashing is a nightmare to use while debugging.

dorinhogea:
Check the bdb attributes.  It is pretty powerful; access straighforward the variable, or use function api.  Set them at runtime using file configuration or message traps.

dorinhogea:
Have we considered, since we provide DDL for setting these options, to switch to DDL in lrl? (instead of custom format parsing).  This will provide a unified way for run-time and static configuration option processing.

dorinhogea:
allowing DDL (and DML) in lrl open other opportunities as well

dorinhogea:
(i.e db configuration would be just another SQL file fed during startup, or manually during an interactive session)

bb/walkback.c
```diff
@@ -26,10 +26,10 @@
 /* #include <hexdump.h> */
 #endif
 
-static int warnthresh = 50;
-static unsigned long long count = 0, ivcount = 0;
+int walkback_enabled = 1;
```
dorinhogea: promoting these to global... should they be renamed as gbl_...?

bdb/bdb_net.c
```diff
@@ -130,7 +130,8 @@ static unsigned int fail_udp = 0;
 static unsigned int recd_udp = 0;
 static unsigned int recl_udp = 0; /* problem with recv'd len */
 static unsigned int rect_udp = 0; /* problem with recv'd to */
-static int ack_trace = 0;
+
+int ack_trace = 0;
```
dorinhogea: gbl_?

jmleddy:
Even though I haven't reviewed it I'm really excited about this merge :+1:

sqlite/ext/comdb2/tunables.c
```diff
@@ -0,0 +1,207 @@
...
+{
+    systbl_tunables_cursor *pCur = (systbl_tunables_cursor *)pVtabCursor;
+    pCur->rowid = 0;
+    return SQLITE_OK;
+}
+
+static int systblTunablesNext(sqlite3_vtab_cursor *cur)
+{
+    systbl_tunables_cursor *pCur = (systbl_tunables_cursor *)cur;
+    pCur->rowid++;
```
nirbhayc: That won't be needed as vtable's eof  function (xEof) is called after every next to check if we have reached the end.

db/db_tunables.c
```diff
@@ -0,0 +1,1181 @@
...
+        *num = n;
+        return 0;
+    }
+    return 1;
+}
+
+/* Parse the next token and store it into a buffer. */
+#define PARSE_TOKEN                                                            \
+    tok = segtok((char *)value, value_len, &st, &ltok);                        \
+    memcpy(buf, tok, ltok);                                                    \
```
nirbhayc: Done.

db/db_tunables.c
```diff
@@ -0,0 +1,1181 @@
...
+    }
+    return 0;
+}
+
+/*
+  Update the tunable at runtime.
+
+  @return
+    0           Success
+    1           Failure
```
nirbhayc: Good point. Though this change is not needed now, but it will be once I fix process_message to use handle_runtime_tunable(). Fixed.

sqlite/comdb2build.c
```diff
@@ -1675,3 +1675,17 @@ void comdb2WriteTransaction(Parse *pParse)
 {
     pParse->write = 1;
 }
+
+void comdb2putTunable(Parse *pParse, Token *name, Token *value)
+{
+    char *t_name;
+    char *t_value;
+
+    create_string_from_token(NULL, pParse, &t_name, name);
```
nirbhayc: Fixed.

bb/walkback.c
nirbhayc: Done.

bdb/bdb_net.c
nirbhayc: Yes, fixed.

sqlite/ext/comdb2/tunables.c
dorinhogea: strong assumption from higher layer, more robust to simply not point to random memory

ldennis:
Can you update the comdb2sys.test expected output as well?

PR #202, opened 2017-07-03
change Mem back to original type after sqliteVdbeMemDecimalBasicArith…

ldennis:
…metics

fixed:
```
[drop table if exists t1] rc 0                                        
[create table t1 {                                                    
    tag ondisk {                                                      
        int id                                                        
        decimal32 dec                                                 
        double num                                                    
    }                                                                 
    keys {                                                            
        "ID" = id                                                     
    }                                                                 
}] rc 0                                                               
(rows inserted=1)                                                     
[insert into t1 values(1, 5, 10)] rc 0                                
(id=1, dec='5', num=10.000000)                                        
[select * from t1] rc 0                                               
[update t1 set dec = dec * '2', num = num * '2' where id = 1] failed with rc 113 incompatible values from SQL interval (day-second, year-month or decimal) to bdouble field 'num' for table 't1'
(id=1, dec='5', num=10.000000)                                        
[select * from t1] rc 0                                               
[update t1 set dec = dec * 2, num = num * 2 where id = 1] failed with rc 113 incompatible values from SQL interval (day-second, year-month or decimal) to bdouble field 'num' for table 't1'
(id=1, dec='5', num=10.000000)                                        
[select * from t1] rc 0                
```

Both  `update t1 set dec = dec * '2', num = num * '2' where id = 1` and `update t1 set dec = dec * 2, num = num * 2 where id = 1` should not fail.

The problem is when we do `dec = dec * 2`, we convert integer 2 to interval (with decimal type) and we dont change it back. Therefore, `num * 2` fails.

### DIFF
--- a/sqlite/vdbemem.c
+++ b/sqlite/vdbemem.c
@@ -2149,22 +2149,7 @@ int sqlite3VdbeMemDecimalfy(Mem *pMem)
       }
    }
 
-   /* don't touch these 
-   if( pMem->flags & MEM_Dyn ){
-     sqlite3DbFree(pMem->db, pMem->z);
-   }
-
-
-   pMem->n = 0;
-   pMem->z = 0;
-   pMem->flags &= ~(MEM_Int|MEM_Str|MEM_Static|MEM_Dyn|MEM_Ephem);
-   */
-
-   if(pMem->flags & MEM_Dyn)
-   {
-      sqlite3DbFree(pMem->db, pMem->z);
-       pMem->z = 0;
-   }
+   sqlite3VdbeMemRelease(pMem);
    pMem->flags = MEM_Interval;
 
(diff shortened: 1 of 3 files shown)


sqlite/vdbemem.c
```diff
@@ -3021,6 +3026,8 @@ int sqliteVdbeMemDecimalBasicArithmetics(
   }
 
 done:
+   if( decimalfied )
+     memcpy(b, &sav, sizeof(Mem));
```
dorinhogea: b->z could be freed by sqlite3VdbeMemDecimalfy

ldennis: Why do we have `don't touch these` in comment but free b->z later? Why do we need to free b->z?

PR #203, opened 2017-07-03
Hash tables and queues for efficient lookup.

nirbhayc:


db/comdb2.c
```diff
@@ -875,39 +874,33 @@ struct db *getdbbynum(int num)
...
-    int idx;
+    struct db *db;
+    int idx = -1;
 
-    /*should be changed to a hash table*/
-    for (idx = 0; idx < thedb->num_dbs; ++idx) {
-        if (strcasecmp(thedb->dbs[idx]->dbname, p_name) == 0)
-            return idx;
+    db = hash_find_readonly(thedb->db_hash, &p_name);
+    for (int i = 0; i < thedb->num_dbs && db; ++i) {
```
dorinhogea: Please use db->dbs_idx with pointer validation instead of array scan

db/comdb2.c
```diff
@@ -8805,38 +8820,16 @@ void delete_db(char *db_name)
         exit(1);
     }
 
+    /* Remove the table from hash. */
+    hash_del(thedb->db_hash, thedb->dbs[idx]);
+
```
dorinhogea: What protects this hash against concurrent getdbbyname() lookups?

db/comdb2.c
nirbhayc: Even better. Thanks!

db/comdb2.c
nirbhayc: It's protected by delete_db()'s acquisition of wrlock on thedb_lock.

PR #204, opened 2017-07-04
Update shadows fix

markhannum:
Contains the C version of Kyle Kingsbury's register test, along with fixes for crashes that occurred as a result of not handling recover_deadlock/get_curtran failures correctly.


markhannum:
I'm separating this into 2 different pull requests

PR #207, opened 2017-07-05
formatting

dorinhogea:
basically playing with isspace()

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -5715,8 +5715,7 @@ static struct dbenv *newdbenv(char *dbname, char *lrlname)
         return NULL;
     }
 
-    if (lrlname)
-        pre_read_lrl_file(dbenv, lrlname, dbname);
+    if (lrlname) pre_read_lrl_file(dbenv, lrlname, dbname);
 
     /* if we havn't been told not to load the /bb/bin/ config files */
     if (!gbl_nogbllrl) {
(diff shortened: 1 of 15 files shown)


db/comdb2.c
```diff
@@ -5715,8 +5715,7 @@ static struct dbenv *newdbenv(char *dbname, char *lrlname)
         return NULL;
     }
 
-    if (lrlname)
-        pre_read_lrl_file(dbenv, lrlname, dbname);
+    if (lrlname) pre_read_lrl_file(dbenv, lrlname, dbname);
```
adizaimi: I wonder if we should leave the existing one line if stmts as is, ie. not put them in one line like here, but leave them like it was before.

riverszhang89:
Thanks Dorin.

dorinhogea:
thank you!

db/comdb2.c
dorinhogea: Yes, I would like that, it is easier to set a breakpoint on condition match if two lines. Change roborivs?

riverszhang89: Roborivers uses the coding style in [this doc](https://bloomberg.github.io/comdb2/contrib.html#indentation-etc
).

adizaimi: Well, you can try with AllowShortIfStatementsOnASingleLine: false, or try omitting it totaly.
I did not think we decided to force if stmts on same line -- if we did that's silly and we should not enforce.

PR #208, opened 2017-07-06
Sqlite serial

mohitkhullar:
For sqlite serialization and deserialization keep datetimes and intervals in native endian format


dorinhogea:
how is remote sql gonna work between different endian systems?

mohitkhullar:
will need to make changes for remote db to send its endianess so that we can send data in correct sqlite format.

PR #209, opened 2017-07-06
Selectv tran cleanup

akshatsikarwar:


db/glue.c
```diff
@@ -392,20 +392,14 @@ int trans_start_set_retries(struct ireq *iq, tran_type *parent_trans,
...
-tran_type *trans_start_socksql(struct ireq *iq, int ignore_newer_updates,
-                               int trak)
+tran_type *trans_start_socksql(struct ireq *iq, int trak)
 {
     void *bdb_handle = bdb_handle_from_ireq(iq);
     tran_type *out_trans = NULL;
     int bdberr = 0;
 
     iq->gluewhere = "bdb_tran_begin_socksql";
-    if (gbl_extended_sql_debug_trace) {
```
dorinhogea: This trace probably should be kept without the ignore_newer_updates info.  It tracks transaction begins, aborts, commits and so on.

dorinhogea:
I think it is easier for reviewers if we separate cleanup projects from fixes and new code.  A cleanup would be relatively straightforward, but can have a larger change footprint and a higher chance of merge conflicts.   Keeping cleanup separated would shorten the amount of review time and reduce conflicts for everyone.  In addition, reverting a surgery fix that misfired would not require losing the cleanup too.

akshatsikarwar:
> I think it is easier for reviewers if we separate cleanup projects from fixes and new code.

I agree. In this case changes were very inter-dependent. I've authored 2 commits so everything you noted applies (reviewing, merging, reverting, etc)

db/glue.c
akshatsikarwar: OK

db/sqlglue.c
```diff
@@ -4904,47 +4880,24 @@ static int sqlite3BtreeBeginTrans_int(Vdbe *vdbe, Btree *pBt, int wrflag,
...
     clnt->dml_tables = hash_init_str(0);
 
-    switch (clnt->dbtran.mode) {
-    case TRANLEVEL_SOSQL:
-    case TRANLEVEL_RECOM:
-        /* if the query is a select, punt here
-         * SELECTV inside after BEGIN creates a transaction (is_recording)
-         */
-        assert(clnt->sql);
+    if (vdbe == NULL || pBt->is_temporary) {
```
dorinhogea: When is vdbe NULL?

db/sqlglue.c
```diff
@@ -5006,10 +4957,12 @@ int sqlite3BtreeCommit(Btree *pBt)
     if (clnt->selectv_arr)
         currangearr_coalesce(clnt->selectv_arr);
 
+    if (clnt->ctrl_sqlengine == SQLENG_NORMAL_PROCESS) {
```
dorinhogea: This is too early, we need first to commit, make sure we don't repeat if verify, than reset has_recording.
Example:
"insert into ... selectv ..." gets verify error

dorinhogea: In addition, for standalone queries, it is possible that sqlite engine decides to abort and calls Rollback instead of commit; we should reset the flag as well.

db/sqlinterfaces.c
```diff
@@ -2380,6 +2380,7 @@ int handle_sql_commitrollback(struct sqlthdstate *thd,
 
     if (rc == SQLITE_OK) {
         /* send return code */
+        clnt->has_recording = 0;
```
dorinhogea: This is too early; we need to capture rc !=SQLITE_OK as well.  
For an more robust cleanup, in case of multi-statement transaction, I recommend cleaning up the flag at a higher level (in handle_non_sqlite_request, which captures begin, wrong-state, and commit-rollback case).

db/sqlglue.c
```diff
@@ -4885,47 +4880,24 @@ static int sqlite3BtreeBeginTrans_int(Vdbe *vdbe, Btree *pBt, int wrflag,
     clnt->ddl_tables = hash_init_str(0);
     clnt->dml_tables = hash_init_str(0);
 
-    switch (clnt->dbtran.mode) {
-    case TRANLEVEL_SOSQL:
```
dorinhogea: Removing this is great!

dorinhogea:
From what I understand, to summarize, the selectv fix replaces run-time "has_recording" activation with parser phase setup, which addresses the bug properly.  A consequence of this is removing the "has_recording" flag reset at the start of transaction, which makes cleaning it up a bit more convoluted.  I have tried to cover all return paths, please see review comments for details.  We need to cleanup at the conclusion of each transaction, after the commit error (if any) is checked, and on every return path, successful or not.

db/sqlglue.c
akshatsikarwar: During bootstrap (sqlite_master + stats)

akshatsikarwar:
Thanks. Do you think reset of flag during `reset_clnt` is sufficient? In that case I can remove all others.

dorinhogea:
reset_clnt() is only sent when a client is donating the socket to sockpool, which doesn't cover cases when client gets a "logical" error (like a constraint error) and keep using the same session/socket.

dorinhogea:
Additional issue: cached statements skip phase prepare phase and need additional mechanism to set flag

akshatsikarwar:
> Additional issue: cached statements skip phase prepare phase and need additional mechanism to set flag

Good catch!

akshatsikarwar:
Thoughts?

dorinhogea:
LGTM

PR #211, opened 2017-07-07
fixes for occ sql, time partitions, remote sql following code static …

dorinhogea:
…analysis

### DIFF
--- a/db/osqlrepository.c
+++ b/db/osqlrepository.c
@@ -305,12 +305,11 @@ int osql_repository_rem(osql_sess_t *sess, int lock, const char *func, const cha
                     p, found_uuid);
         }
 
-        abort();
         if (lock) {
             theosql_pthread_rwlock_unlock(&theosql->hshlck);
         }
 
-        return -2;
+        abort();
     }
     else {
         memcpy(uuid_list[current_uuid], sess->uuid, sizeof(uuid_list[0]));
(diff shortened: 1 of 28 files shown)


db/osqlrepository.c
```diff
@@ -305,12 +305,11 @@ int osql_repository_rem(osql_sess_t *sess, int lock, const char *func, const cha
                     p, found_uuid);
         }
 
-        abort();
         if (lock) {
```
adizaimi: If we are aborting, why unlock?

db/translistener.c
```diff
@@ -1166,22 +1166,31 @@ int javasp_load_procedure_int(const char *name, const char *param,
...
+    }
+    if (paramvalue) {
         p->qname = strdup(name);
+        if(!p->qname) goto oom;
+    }
     if (param)
         p->param = strdup(param);
     else
         p->param = strdup("<sc>");
+    if(!p->param) goto oom;
```
adizaimi: Does not really matter, but we are not freeing p->qname if we goto oom from here. Nor p->name either.

adizaimi:
looks good to me.

db/osqlrepository.c
dorinhogea: Looks like the abort is meant more like a temporary patch that will allow future investigation in what produce a certain type of error (and probably we didn't figure out how to unwind properly the error).

db/translistener.c
dorinhogea: gonna fix that anyway

riverszhang89:
1. Please squash less meaningful commits into one.
2. Please merge your branch after passing regression.

PR #214, opened 2017-07-10
reload_analyze: dont get_curtran if we already have one

ldennis:
- reload_analyze: dont get_curtran if we already have one
- added testcase t10_01.sh to analyze.test to run a update thread, a select thread and an analyze thread in parallel for 3 minutes. This test makes sure writes, reads and analyzes can run concurrently.

This PR fixes the issue where we already hold a curtran from `sqlengine_work_appsock` and then we call `check_thd_gen->reload_analyze` in `get_prepared_stmt` which tries to get a curtran again.


riverszhang89:
Does the patch fix intermittent analyze test failure?

ldennis:
No, this patch was not meant for that.

ldennis:
Adi reviewed this in bbgithub

PR #215, opened 2017-07-10
fix remsql

ldennis:
I guess this is a day0 bug as originally we had `if (ixschema->flags & SCHEMA_DATACOPY == 0)` And it happened to work because `SCHEMA_DATACOPY == 0` is always 0. I guess #211 was trying to fix this but it actually broke it.


dorinhogea:
looks good, thank you

PR #216, opened 2017-07-10
TRUNCATE [TABLE] tab-name

nirbhayc:
* Allow TRUNCATE command to optionally accept TABLE keyword.
* In many RDBMSes, it's either mandatory or optional.


akshatsikarwar:
Thank you Thank you Thank you!

PR #217, opened 2017-07-10
index on expression fixes

ldennis:
- fix index on expressions on string segfault in schemachange.
- added experimental "enable_new_indexes" lrl option to let replicants ship key values to master for all tables. (w/o the option, replicants only send key values to master for tables that contain indexes on expressions.)
- added a test (idxexpr_alltypes.test) to test indexes on expressions on all supported data types.

### DIFF
--- a/bdb/bdb_verify.c
+++ b/bdb/bdb_verify.c
@@ -915,15 +915,16 @@ static int bdb_verify_ll(
                 }
 
             } else if (bdb_state->ixcollattr[ix]) {
-                if (dbt_data.size !=
-                    (sizeof(unsigned long long) + bdb_state->ixcollattr[ix])) {
+                if (dbt_data.size != (sizeof(unsigned long long) +
+                                      4 * bdb_state->ixcollattr[ix])) {
                     ret = 1;
-                    locprint(sb, lua_callback, lua_params, "!%016llx ix %d decimal payload wrong size "
-                                    "expected %d got %d\n",
-                                genid_flipped, ix,
-                                sizeof(unsigned long long) +
-                                    bdb_state->ixcollattr[ix],
-                                dbt_data.size);
+                    locprint(sb, lua_callback, lua_params,
+                             "!%016llx ix %d decimal payload wrong size "
+                             "expected %d got %d\n",
+                             genid_flipped, ix,
+                             sizeof(unsigned long long) +
+                                 4 * bdb_state->ixcollattr[ix],
+                             dbt_data.size);
                     goto next_key;
                 }
                 memcpy(&genid_right, (uint8_t *)dbt_data.data, sizeof(genid));
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -12508,7 +12508,7 @@ int indexes_expressions_data(struct schema *sc, const char *inbuf, char *outbuf,
                              const char *tzname)
 {
     Mem *m = NULL;
-    Mem *mout = NULL;
+    Mem mout = {0};
     int nblobs = 0;
     struct field_conv_opts_tz convopts = {.flags = 0};
     struct mem_info info;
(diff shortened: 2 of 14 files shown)


db/sqlglue.c
```diff
@@ -12248,12 +12251,37 @@ void bind_verify_indexes_query(sqlite3_stmt *stmt, void *sm)
...
+                if (pTo->zMalloc == NULL) return SQLITE_NOMEM;
+                memcpy(pTo->zMalloc, pFrom->zMalloc, pTo->szMalloc);
+                pTo->z = pTo->zMalloc;
+            } else if (pFrom->z && pFrom->n) {
+                pTo->n = pFrom->n;
+                pTo->szMalloc = pFrom->n + 1;
+                pTo->zMalloc = malloc(pTo->szMalloc);
+                if (pTo->zMalloc == NULL) return SQLITE_NOMEM;
+                memcpy(pTo->zMalloc, pFrom->z, pFrom->n);
+                pTo->zMalloc[pFrom->n] = 0;
```
riverszhang89: You might want to add Mem_Term flag here.

db/sqlglue.c
```diff
@@ -12488,7 +12516,8 @@ int indexes_expressions_data(struct schema *sc, const char *inbuf, char *outbuf,
         tzname = "America/New_York";
 
     sql = strbuf_new();
-    memset(&mout, 0, sizeof(Mem));
+    mout = (Mem *)malloc(sizeof(Mem));
```
riverszhang89: Why can't we use the memory on stack?

db/sqlglue.c
ldennis: This is a fake Mem structure and it is not used in VDBE. zMalloc will be freed in `indexes_expressions_data`. See below

db/sqlglue.c
```diff
@@ -12530,15 +12559,17 @@ int indexes_expressions_data(struct schema *sc, const char *inbuf, char *outbuf,
...
-                        "%x, index type %d\n",
-                __func__, rc, f->name, mout.flags, f->type);
+        logmsg(LOGMSG_ERROR,
+               "%s: rc %d failed to form index \"%s\", result flag "
+               "%x, index type %d\n",
+               __func__, rc, f->name, mout->flags, f->type);
         goto done;
     }
-    sqlite3VdbeMemRelease(&mout);
+    if (mout->zMalloc) free(mout->zMalloc);
```
ldennis: This is where zMalloc will be freed. And this Mem does not go into VDBE. And I think it is ok to leave it as is?

db/sqlglue.c
ldennis: Sure. I was just trying to be safe as this is passed to another thread (`dispatch_sql_query`). But this function wont exit until the sql returns. So I think it is ok to put it back to stack.

bdb/bdb_verify.c
```diff
@@ -915,15 +915,16 @@ static int bdb_verify_ll(
                 }
 
             } else if (bdb_state->ixcollattr[ix]) {
-                if (dbt_data.size !=
-                    (sizeof(unsigned long long) + bdb_state->ixcollattr[ix])) {
+                if (dbt_data.size != (sizeof(unsigned long long) +
+                                      4 * bdb_state->ixcollattr[ix])) {
```
riverszhang89: Could you explain why you multiply it by 4?

db/sqlglue.c
```diff
@@ -2294,6 +2296,7 @@ int new_indexes_syntax_check(struct ireq *iq)
         return -1;
 
     sql_mem_init(NULL);
+    thread_memcreate(indexes_thread_memory);
```
riverszhang89: Why do you need 2 thread-local allocators (sqlite and thread_mem) ?

db/sqlglue.c
riverszhang89: It is better to use sqlite malloc module as it is always thread-local. See https://github.com/bloomberg/comdb2/pull/42.

bdb/bdb_verify.c
ldennis: See [here](https://github.com/bloomberg/comdb2/blame/master/db/tag.c#L7459). This was a verify day1 bug for decimal indexes? But I think it would be better to use a marco or sizeof(short).. But this is out of this PR's scope. Fixed this to make verify pass.

riverszhang89: What about https://github.com/bloomberg/comdb2/blob/master/bdb/bdb_osqllog.c#L3732.

ldennis: This might be a bug too. But this one will only affect snapshot. I will address this later in my other snapshot related project.

riverszhang89: Suggestion: If we need to multiply bdb_state->ixcollattr[ix] by 4 every time, why don't we just store 4x

ldennis: I guess we wanted to store the number of decimal fields in a compound key. I agree that this needs to be cleaned up in the future.

ldennis:
Thanks a lot for reviewing. :+1:

PR #219, opened 2017-07-10
Fixes for generation-changes between curtrans

markhannum:
This fixes database crashes when a generation change leaves a transaction without a curtran.  Rather than continuing (and aborting when the lockerid is no longer found), the error should be propagated to the client-api to retry against another node.


db/sqlinterfaces.c
```diff
@@ -7605,14 +7606,25 @@ static int fsql_writer(SBUF2 *sb, const char *buf, int nbytes)
         }
         if (rc == 0) { // descriptor not ready, write will block
             if (gbl_sql_release_locks_on_slow_reader && !released_locks) {
-                release_locks("slow reader");
+                rc = release_locks("slow reader");
+                if (rc) {
+                    logmsg(LOGMSG_ERROR,
+                           "%s release_locks generation changed\n", __func__);
+                    return -(SQLITE_DEADLOCK);
```
adizaimi: Maybe return -1 if the value itself does not matter?

PR #221, opened 2017-07-10
Addressing comdb2_malloc performance issue.

riverszhang89:
We use a small history list to store the most recent N (N is a compile-time option) memory blocks obtained from system, and trim only if the top-most block is in the history list. The change significantly reduces the number of allocation calls, and improves data locality.


riverszhang89:
Seeing malloc corruption from regression test. Closing.

PR #222, opened 2017-07-11
bdb_direct_count: retry on deadlocks

ldennis:
Select count(*) returns -5 unknown error on deadlocks with the direct count feature.
Direct count should retry on c_get() if we get a deadlock.
This patch also fixes most of the intermittent analyze.test failures (while some other failures might be related to low disk headroom of robo).

I ran analyze.test in a tight loop for over 3 hours without any failures.

### DIFF
--- a/bdb/cursor.c
+++ b/bdb/cursor.c
@@ -8087,16 +8087,7 @@ static void *db_count(void *varg)
         return NULL;
     }
     int64_t count = 0;
-    extern int gbl_move_deadlk_max_attempt;
-    int max_retries =
-        gbl_move_deadlk_max_attempt >= 0 ? gbl_move_deadlk_max_attempt : 500;
-    int nretries = 0;
-    while (1) {
-        rc = dbc->c_get(dbc, &k, &v, DB_NEXT | DB_MULTIPLE_KEY);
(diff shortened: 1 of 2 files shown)


adizaimi:
Akshat, i wanted to remind you to not do direct_count if trans is snapshot iso or larger (as you may get updates to the btree that are not in your snapshot). We should have a testcase for that actually.

akshatsikarwar:
yup - need to change intrans to check for isolation level instead

akshatsikarwar:
@ldennis Need to change check for `rc`. Current code `or`s the `rc`s from different stripes. Need to check each one individually.

ldennis:
Just took a look at how `intrans` is set. `intrans` will be 0 for readonly trans in `TRANLEVEL_SOSQL` and `TRANLEVEL_RECOM`. See [here](https://github.com/bloomberg/comdb2/blob/master/db/sqlglue.c#L4910). So it might be ok to just use `!pCur->clnt->intrans`. I am not sure if there are any other corner cases. So checking the tran level might be better and it would be at least less confusing.

akshatsikarwar:
Thanks Lingzhi.

PR #223, opened 2017-07-11
fix partial index where clause csc2 parser

### LINKED ISSUE #120: CSC2: Partial index needs to be defined on a separate line
The following command fails:
```
create table t1 { schema { int i int j } keys { "idx"  = i { where j > 10 } } } $$
```
Error:
```
test> create table t1 { schema { int i int j } keys { "idx"  = i { where j > 10 } } } $$
[create table t1 { schema { int i int j } keys { "idx"  = i { where j > 10 } } }] failed with rc -5 OP #3 BLOCK2_SEQV2(824): ERROR at line   0: syntax error: {
```

The partial index definition is expected to be on a separate line. Lifting this limitation by fixing the lexer could be tricky (and potentially error prone). Better move the CSC2 parsing to lemon (as suggested by @akshatsikarwar).

ldennis:
- Use regexp to match where clause in csc2:
```
dqstring \"(\\.|[^\n\\\"])*\"
sqstring \'(\\.|[^\n\\\'])*\'
where [{][\t ]*[Ww][Hh][Ee][Rr][Ee][\t ]({sqstring}|{dqstring}|[^}\"\'])*[\t ]*[}]

```
- Added testcase to make sure some corner cases work
- This should fix #120 for now


ldennis:
This only affects partial indexes parsing. I am merging this in and closing the related issue.

PR #224, opened 2017-07-11
Un-reserve keywords that were reserved by #180

nirbhayc:
- Add them under Lemon's fallback ID, instead.


mohitkhullar:
Can we add a test where column/table name is an unreserved keyword?

nirbhayc:
I think parserdiagnose.test does that already.

PR #226, opened 2017-07-12
Eventlog for tags

jmleddy:


db/eventlog.c
```diff
@@ -637,10 +770,9 @@ static void eventlog_process_message_locked(char *line, int lline, int *toff)
             return;
         }
     } else if (tokcmp(tok, ltok, "flush") == 0) {
-        gzflush(eventlog, 1);
+        eventlog_cb->flush();
     } else {
-        logmsg(LOGMSG_ERROR, "Unknown eventlog command\n");
-        return;
+        eventlog_help();
```
adizaimi: can we maybe print this as part of reql help?

dorinhogea:
any chance we not call them tags? they are not tags :)

PR #228, opened 2017-07-13
fix and cleanup tests

adizaimi:
* dont accumulate tons of longreqs for sirace test
* add type intervaldsus to alltypes test in comdb2sys
* add type intervaldsus to alltypes test in comdb2sys
* remove bad lrl setting sc_logical_txn from few tests
* move keywords test into more appropriate comdb2sys test
* remove bad attr MAX_SHADOW_ROWS from tests
* delete bad lrl option enable_true_serializable from tests
* downgrade to warn logmsg set gbl_recoverable_lsn
* new_snapshot is not a valid option


nirbhayc:
@adizaimi Looks ok. Thanks.

PR #229, opened 2017-07-13
mainly no read random mem

dorinhogea:


ldennis:
You are trying to fix access of `p_db->dbnum = thedb->dbs[idx]->dbnum;` when `thedb->dbs[idx]` is uninitialized, right?
Looks good to me but please format code before merging in. And it would be better if you can do squash and merge. Thanks.

dorinhogea:
yes, thedb->dbs[idx] points to random memory, possible not allocated; also setting dbnum to random values would affect potential internal legacy comdbg clients

PR #233, opened 2017-07-14
Fix a typo in datetime ms/us coercion.

riverszhang89:
The bug causes the database to generate incorrect results when a client
selects back a datetime value with a fraction of 6 milliseconds using
client type CDB2_DATETIMEUS, or selects back a datetimeus value
with a fraction of 3 microseconds using client type CDB2_DATETIME.

The bug has been in the codebase since datetimeus was first introduced.


riverszhang89:
@adizaimi Thanks for reviewing it.

PR #235, opened 2017-07-15
Move member variables around to get smaller size

adizaimi:
Size of structure went from 2608 to 2504 bytes.
Also cleaned up and remove old and unused variables.


db/comdb2.h
```diff
@@ -1428,13 +1359,64 @@ struct ireq {
...
+
+    int ixused;    /* what index was used? */
+    int ixstepcnt; /* how many steps on that index? */
+
+    /* which queues were added to? (so we can wake up their consumer
+     * threads).  num_queues_hit==MAX_QUEUE_HITS_PER_TRANS+1 means that
+     * we'll have to wake up all queues on commit - oh well. */
+    unsigned num_queues_hit;
+
+    unsigned long long last_genid;
```
riverszhang89: You may get even smaller size by moving this line up after `uint64_t txnsize`.

db/comdb2.h
```diff
@@ -1252,18 +1253,14 @@ struct ireq {
     /* REGION 1 */
     /************/
     char *frommach; /* machine that the buffer came from */
```
riverszhang89: If the purpose is to reduce `sizeof(struct ireq)`, may I propose to re-order region 1 as well.

db/comdb2.h
adizaimi: moved it but it does not give any savings on i86. But probably does in solaris.

db/comdb2.h
adizaimi: Thanks. Moved a couple of things around, but alas did not result in any savings.

PR #237, opened 2017-07-17
SQLite: Separate Comdb2 syntax from SQLite core syntax.

riverszhang89:
Some of SQLite's optimizations assume certain VDBE opcodes are consecutive, for example, OP_IdxXX and OP_SeekXX. However we may well break the assumption: Comdb2 rules are defined in the middle of SQLite core syntax, thus the 2 groups of tokens are mixed up.

It can be easily fixed by moving our syntax to the end of the lemon grammar file.


ldennis:
Nice, this fixes the broken test. Looks good to me.

riverszhang89:
Thanks Lingzhi.

PR #238, opened 2017-07-17
DDL: Fix the mapping of blob type with no size

nirbhayc:
- Also add check/test for invalid sizes


nirbhayc:
Thanks for reviewing @riverszhang89 !

PR #243, opened 2017-07-18
tranddl: fix blockprocessor deadlock retry case

ldennis:
- move osql_postcommit_handle and osql_postabort_handle to toblock_main
- correctly test deadlock retry logic in rowlocks mode by forcing a retry rcode
- dont add blkseq when forcing a retry
- this should fix intermittent ptranfail_socksql.test failures

### DIFF
--- a/db/toblock.c
+++ b/db/toblock.c
@@ -5458,7 +5458,8 @@ add_blkseq:
                 uuid_t u;
                 memcpy(&u, iq->seq, iq->seqlen);
                 comdb2uuidstr(u, us);
-            }             /* force a parent-deadlock for cdb2tcm */
+            }
+            /* force a parent-deadlock for cdb2tcm */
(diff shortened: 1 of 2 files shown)


db/toblock.c
```diff
@@ -5551,6 +5552,25 @@ static int toblock_main_int(struct javasp_trans_state *javasp_trans_handle,
             }
         } else /* rowlocks */
         {
+            /* force a parent-deadlock for cdb2tcm */
+            if ((tcm_testpoint(TCM_PARENT_DEADLOCK)) && (0 == (rand() % 20))) {
+                logmsg(LOGMSG_USER, "tcm forcing parent retry in rowlocks\n");
```
nirbhayc: Should this be LOGMSG_DEBUG instead?

ldennis: Sure, but this was copy&paste code..

ldennis:
No tranddl logic change and the deadlock retry code is only for rowlocks testing. Merging in to continue working on rowlocks code.

PR #244, opened 2017-07-18
Have select sleep(N), perform N times sleep(1)

adizaimi:
This is to be able to kill client running sql query and not clog
server (which prior to this patch would keep sleeping for N seconds).

### DIFF
--- a/sqlite/func.c
+++ b/sqlite/func.c
@@ -395,6 +395,7 @@ static void substrFunc(
   }
 }
 
+extern int comdb2_sql_tick();
 /* COMDB2 MODIFICATION */
 /*
 ** Implementation of the sleep() function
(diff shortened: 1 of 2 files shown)


sqlite/func.c
```diff
@@ -411,7 +412,11 @@ static void sleepFunc(sqlite3_context *context, int argc, sqlite3_value *argv[])
     sqlite3_result_int(context, -1);
     return;
   }
-  rc = sleep(n);
+  for(int i = 0; i < n; i++) {
+    rc = sleep(1);
```
nirbhayc: Please remove rc, as its never accessed.

adizaimi:
Thank you.

PR #249, opened 2017-07-19
cdb2sql: send cancel cnonce to connected db when hitting ctrl-c

adizaimi:


sqlite/func.c
```diff
@@ -416,7 +416,7 @@ static void sleepFunc(sqlite3_context *context, int argc, sqlite3_value *argv[])
   for(i = 0; i < n; i++) {
     sleep(1);
     if( comdb2_sql_tick() )
-      break;
+      break;  /* COMDB2: we could also return error sqlite3_result_error(context, "Interrupted", -1); */
```
nirbhayc: You will get a formatting error here. Please break the comment.

tools/cdb2sql/cdb2sql.c
```diff
@@ -694,7 +695,9 @@ static void process_line(char *sql, int ntypes, int *types)
     sqlstr[len] = '\0';
 
     int start_time_ms, run_time_ms;
+    gbl_in_stmt = 1;
```
nirbhayc: Use of gbl_in_stmt seems redundant to me. If we have a nonce, let's send it to the server and let the server take care of terminating the query if a match found.

tools/cdb2sql/cdb2sql.c
```diff
@@ -887,6 +890,38 @@ static void replace_args(int argc, char *argv[])
...
+    cdb2_hndl_tp *cdb2h_2 = NULL;
+    int rc;
+    if (dbhostname) {
+        rc = cdb2_open(&cdb2h_2, dbname, dbhostname, CDB2_DIRECT_CPU);
+    } else {
+        rc = cdb2_open(&cdb2h_2, dbname, dbtype, 0);
+    }
+    if(rc) { 
+        if (debug_trace)
+            printf("Open db to cancel sql failed\n");
```
nirbhayc: It should be printed to stderr and please also print the return code.

sqlite/func.c
adizaimi: sqlite has its own formatting larger than 80 chars. This should be fine.

tools/cdb2sql/cdb2sql.c
adizaimi: It is useless to talk to the server if there is no query actually running. Imagine doing ctrl_c several times in a row--you would talk to the server trying to cancel something without good reason.

nirbhayc:
```
Imagine doing ctrl_c several times in a row--you would talk to the server trying to cancel something without good reason.
```
That's not right. The first Ctrl-C should kill the query and the next one should kill the client.

adizaimi: What you are explaining is what I had implemented, until the consensus was to behave like bash and allow ctrl_c in a row to do nothing but just go to newline. Thus i removed that behavior.

nirbhayc: oh ok. got a little confused.

PR #251, opened 2017-07-19
Early retry on updates

ldennis:
- PORTED: early retry on updates
- enable recom_replay.test and add t3 to test the ported feature


db/comdb2.c
```diff
@@ -6480,6 +6482,9 @@ static void register_all_int_switches()
     register_int_switch("abort_invalid_query_info_key",
                         "Abort in thread-teardown for invalid query_info_key",
                         &gbl_abort_invalid_query_info_key);
+    register_int_switch("early_verify",
+                        "Give early verify errors for updates in SQLite layer",
+                        &gbl_early_verify);
```
nirbhayc: You might also want to update the expected tunables.test output.

ldennis: ah.. thanks

ldennis:
Reviewed internally.

PR #252, opened 2017-07-19
fix sql comparisons

ldennis:
The bug is "select x=x from t1" returning (x=x=x'302e30')
The cause was SQLite P5 SQLITE_KEEPNULL bit overlapped with SQLITE_AFF_MASK, which led to failure of writting result to pOut, leaving the result undefined.

See comments in code for more details.


dorinhogea:
Is this something we can push to sqlite?

sqlite/vdbe.c
```diff
@@ -2495,7 +2495,7 @@ case OP_Ge: {             /* same as TK_GE, jump, in1, in3 */
     pOut = &aMem[pOp->p2];
     iCompare = res;
     res2 = res2!=0;  /* For this path res2 must be exactly 0 or 1 */
-    if( (pOp->p5 & SQLITE_KEEPNULL)!=0 ){
+    if( (pOp->p5 & SQLITE_KEEPNULL)==SQLITE_KEEPNULL ){
```
adizaimi: This comparison is redundant--original works fine--i would not change it since there is no need.

ldennis: No, original wont work if we are combining STOREP2 and JUMPIFNULL. If we reach here, STOREP2 is set already. If we use `!=0`, it would be always true.

adizaimi: Sorry, can you give an example?
In general:
0110 & 0100   -> 0100     no need to compare with 0100, can just do if(x!=0).
and
0010 & 0100   -> 0000    
I think i am misunderstanding something.

ldennis: STOREP2 is 0x20 -- `0010 0000`
KEEPNULL is 0x30 -- `0011 0000`
If we reach the code at this line, it is guaranteed that STOREP2 is set, i.e. p5 is `**1* ****`.
But we want to check if KEEPNULL is also set here, and `**1* **** & 0011 0000` is always nonzero. So instead we want to check `**1* **** & 0011 0000 == 0011 0000` to make sure both bits are set.

adizaimi: Ok, sorry. My misunderstanding, KEEPNULL is not a simple mask.
thanks

adizaimi:
looks good

sqlite/sqliteInt.h
```diff
@@ -1772,7 +1772,26 @@ struct CollSeq {
...
+** KEEPNULL is checked only when comparison results are stored to P2
+** (i.e. STOREP2 is set)
+** Jump is only taken when STOREP2 is not set and JUMPIFNULL is only checked
+** when we actually jump insteaed of storing result to P2. In other words,
+** STOREP2 and JUMPIFNULL are mutually exclusive. Therefore, it should be ok to
+** combine these two for SQLITE_KEEPNULL and use 0x30.
+**
+** -- Lingzhi Deng
+*/
+#define SQLITE_KEEPNULL     0x30  /* Used by vector == or <> */
```
adizaimi: Do you think it is easier to understand if you 
#define SQLITE_KEEPNULL              SQLITE_JUMPIFNULL  & SQLITE_STOREP2
I totally missed the value of SQLITE_KEEPNULL

ldennis: Logically, `SQLITE_KEEPNULL` has nothing to do with `SQLITE_JUMPIFNULL`. In my opinion, defining in such way would be confusing too. This is kinda hacky as we are running out of bits.

ldennis:
@dorinhogea SQLite works fine and does not necessarily need this . But we sure can ask Hipp to reserve more P5 bits for us.

PR #255, opened 2017-07-20
Remote read sql fails inside SP after SP that writes

mohitkhullar:
This is when we do transactions inside SP, reset the iswrite flag after commit/rollback.


adizaimi:
This looks fine. Is there a way to add test case for this to sp.test?
Thanks

mohitkhullar:
added testcase

ldennis:
Looks good

PR #256, opened 2017-07-20
Document cdb2_bind_index

mohitkhullar:


docs/pages/programming/c_api.md
```diff
@@ -325,6 +326,37 @@ Parameters:
 |*valueaddr*| input | The value pointer of replaceable param | The value associated with this pointer should not change between bind and [cdb2_run_statement](#cdb2runstatement), and for numeric types must be signed. |
 |*length*| input | The length of replaceable param | This should be the sizeof(valueaddr's original type), so 1 if it's a char, 4 for float... |
 
+### cdb2_bind_index
```
nirbhayc: How is this different from cdb2_bind_param (besides being faster)?

mohitkhullar: We can use unnamed parameters in SQL.

nirbhayc: Ok. Can we also add this to the description?

adizaimi:
Can we add a testcase with this call?

adizaimi:
We can add to one of the existing testcases.

docs/pages/programming/c_api.md
```diff
@@ -325,6 +326,37 @@ Parameters:
...
+```
+int cdb2_bind_index(cdb2_hndl_tp *hndl, int index, int type, const void *varaddr, int length);
+```
+
+Description:
+
+This routine is used to bind value pointers to replaceable params in sql statement. The index starts from 1, and increases for every new parameter in the statement. This version of cdb2_bind_* is faster than cdb2_bind_param.
```
nirbhayc: This routine is used to bind value pointers to replaceable params in sql statement. ->
This routine is used to bind value pointers to replaceable params based on their position/index in the SQL statement.

tests/cdb2bind.test/runit
```diff
@@ -0,0 +1,24 @@
...
+    echo "Testcase requires <dbname> argument."
+    exit 1
+
+fi
+
+# Driver 
+driver=client
+
+# Run tasks
+$driver $dbnm
```
adizaimi: this will never fail since we echo Success regardless? Maybe check $?

mohitkhullar: there is set -e @ top, this bash script will exit in case of error.
This will never reach Success if we get non-zero return from binary.

adizaimi: great, i missed that. thanks

adizaimi:
LGTM.

PR #257, opened 2017-07-20
Send dbnames in error in case of incorrect database

mohitkhullar:


db/sqlinterfaces.c
```diff
@@ -7862,7 +7862,8 @@ int handle_newsql_requests(struct thr_handle *thr_self, SBUF2 *sb,
             strcasecmp(sql_query->dbname, thedb->envname)) {
             logmsg(LOGMSG_ERROR, "DB name mismatch query:'%s' actual:'%s' \n",
```
adizaimi: In that case, now you can do logmsg(LOGMSG_ERROR, errstr);
(may need to add \n at the end of errstr)

mohitkhullar: done

nirbhayc:
Let's have PID stored in the client handle at the init time and return error if/when it's found to be different from handle's PID.

PR #259, opened 2017-07-20
sqlite_ext Fix 'columnnumber' field in comdb2sys_keycomponents.

jmleddy:


sqlite/ext/comdb2/keycomponents.c
```diff
@@ -167,7 +167,7 @@ static int systblFieldsColumn(
       break;
     }
     case STFIELD_COLNO: {
-      sqlite3_result_int64(ctx, (sqlite3_int64)pField->idx);
+      sqlite3_result_int64(ctx, (sqlite3_int64)pCur->iFieldid);
```
nirbhayc: If this doesn't break the existing comdb2sys tests, then I recommend adding one.

adizaimi: I fixed something minor in comdb2sys test output--probably related to this fix.

PR #262, opened 2017-07-21
Tunables refactoring: 2nd iteration

nirbhayc:
- Move config-related code to db/config.c
- Cleanup pre- & dynamic tunables
- More tests


ldennis:
Overall it looks good. Do you mind moving the int switches to a better place as well? And another thing is that I am not sure how many internal users are depending on LRLs like "disable_\*"/"enable_\*". We may want to check with DBAs in the future to further refactor those into switches (or whatever cleaner).

nirbhayc:
@ldennis Right. Switches are, more or less, already implemented in a separate unit, and in order to be backward compatible - allow the use of "on/off <switch-name>" in lrl and in sys.cmd.send() at runtime - I think, for the time being, it's ok to leave them as is. We should replace/deprecate enable/disable/do/dont tunables in favor of single boolean tunable, so that they can be removed in some future version.

riverszhang89:
You forgot to adjust analyze test output. Please make sure all tests are passing before merging.

nirbhayc:
@riverszhang89 Thanks for the fix.

riverszhang89:
No problem!

PR #263, opened 2017-07-22
Only LOGMSG_USER messages should be sent to the user

nirbhayc:
- All other messages must be directed to the server log.
- Change all messages in process_command to LOGMSG_USER
as they are intended for the user.
- Clang-formatting


riverszhang89:
We will still need to send LOGMSG_ERROR messages to comdb2sc during the transition from legacy systems to modern DDL/DML. It is too soon to do the change.

nirbhayc:
In that case, LOGMSG_xxx should have OR-able values.

PR #264, opened 2017-07-22
Rename struct db to dbtable

adizaimi:


nirbhayc:
What not rename to 'table'?

adizaimi:
Thought to keep at least some old resemblance. 
Is this ok?

nirbhayc:
Renaming it to table should not need any old resemblance. :) Anyway, that was just an opinion. I am ok with renaming it to dbtable as well. Thank you for taking care of this.

adizaimi:
Thanks Nirbhay. The reason dbtable might be more useful is to allow us to still understand (until fixed) all the references that now look like this:
struct *dbtable db;
somevar->usedb = db;

nirbhayc:
I agree.

PR #265, opened 2017-07-24
Incremental backup

dennis-wei:
Implemented incremental backups as an option in comdb2ar


tools/comdb2ar/comdb2ar.cpp
```diff
@@ -28,19 +30,33 @@ extern "C" {
...
 "  optionally specify destination directories for the lrl file and data",
 "  directory; if ommitted the paths will be taken from the lrl in the",
 "  serialised input stream.",
 "  -C strip     strip cluster nodes lines from lrl file",
 "  -C preserve  preserve cluster nodes lines in lrl file",
+"  -I create    create the incremental meta files while serialising",
+"  -I inc       create an increment for the incremental backup",
+"  -I restore   restore from a sequence of base_backup | increments",
+"  -b <path>    location to store/load the incremental backup",
+"  -k/K         do-not/do keep all logs in an incremental restoration",
```
nirbhayc: What if none of these options (-k/-K) were used? Would the logs be kept or not?

nirbhayc: So, I see that "keep_all_logs = false;" by default, that is, we implicitly have '-k' set. So, I would rather have just one of the above 2 switches, only to enable keep_all_logs, the absence of which implies the default behavior.

tools/comdb2ar/serialise.cpp
```diff
@@ -263,11 +267,30 @@ static void serialise_file(const FileInfo& file, volatile iomap *iomap=NULL,
...
+                verify_checksum(pagebuf + n, pagesize, file.get_crypto(), file.get_swapped(), &verify_bool, &verify_cksum);
+
+                if(verify_bool){
                     // checksum verified
                     n += pagesize;
                     retry = 5;
+
+                    // If we are in incremental mode, on initial backup creation we want to create the diff files
+                    if(incr_create){
+                        incrFile.write((char *) &(LSN(pagep).file), 4);
```
markhannum: Could you endianize these files? Test by doing a restore on ibm using linux generated increment files (and vice-versa).

markhannum:
I think you're missing a gile:
make
make: *** No rule to make target 'tools/comdb2ar/incr_manifest.o', needed by 'comdb2ar'.  Stop.

tools/comdb2ar/serialise.cpp
```diff
@@ -893,36 +950,122 @@ void serialise_database(
...
+
+        // Serialise the manifest file
+        serialise_string("MANIFEST", manifest.str());
+
+    // Incremental Mode
+    } else {
+        manifest << "# Manifest for serialisation of increment produced on "
+            << getDTString() << std::endl;
+
+
```
markhannum: The "logdelete version" / "recovery_options" logic above should apply to incremental backups as well.

tools/comdb2ar/serialise.cpp
```diff
@@ -893,36 +950,122 @@ void serialise_database(
...
+                if(pages_list.empty()){
+                    new_files.push_back(*it);
+                    write_incr_manifest_entry(manifest, *it, pages_list);
+                // Keep track of the file name and the list of pages to be backed-up
+                } else {
+                    incr_data_files.push_back(*it);
+                    page_number_vec.push_back(pages_list);
+                    write_incr_manifest_entry(manifest, *it, pages_list);
+                    total_data_size += data_size;
+                }
```
markhannum: This works.  Because you are recording the pages in a separate step, other pages could be modified- this is okay, though: logfiles are being held so these changes will be restored in recovery.

tools/comdb2ar/serialise.cpp
```diff
@@ -988,10 +1173,62 @@ void serialise_database(
...
+            throw Error(ss.str());
+        }
+
+        // Iterate through the files, index-aligning with the list of changed pages
+        std::vector<FileInfo>::const_iterator data_it = incr_data_files.begin();
+        std::vector<std::vector<uint32_t>>::const_iterator
+            page_it = page_number_vec.begin();
+
+        ssize_t data_written = 0;
+        // First, serialise any complete log files that are in the .txn
```
markhannum: The order is wrong here: you have to serialize changed pages before logfiles.  So, record all changed pages, then record all logfiles.  Logfiles have to be last or recovery will not work.

tools/comdb2ar/increment.cpp
```diff
@@ -0,0 +1,523 @@
...
+        std::memcpy(pagebuf, char_pagebuf, pagesize);
+
+        uint32_t cksum;
+        bool cksum_verified;
+        verify_checksum(pagebuf, pagesize, false, false,
+                            &cksum_verified, &cksum);
+        if(!cksum_verified){
+            std::ostringstream ss;
+            ss << "serialise_file:page failed checksum verification";
+            throw SerialiseError(filename, ss.str());
```
markhannum: You should retry this.  Checksum failure can happen legitimately if we happen to read a page which was in the process of being paged-out.  Copy the logic in serialise_file.

tools/comdb2ar/serialise.cpp
markhannum: Sorry- it's there, just later

tools/comdb2ar/increment.cpp
```diff
@@ -0,0 +1,523 @@
...
+            if(!process_incr_manifest(manifest_text, datadestdir,
+                    updated_files, new_files, deleted_files,
+                    file_order, options, manifest_sha)){
+                // Failed to read a coherant manifest
+                std::ostringstream ss;
+                ss << "Error reading manifest";
+                throw Error(ss);
+            }
+
+            if(manifest_sha != sha_fingerprint){
```
markhannum: I can't see process_incr_manifest yet- it this calculating a SHA from the date from the manifest line that says: "# Manifest for serialisation of increment produced on <XXXX>"?  If so, this won't work as your serialise code makes separate invocations of getDTString to write this to the manifest and to produce the ".sha" file.  A simple workaround would be to make getDTString an indempotent function.

dennis-wei: process_incr_manifest should be there now, for some reason the file was caught in .gitignore

the sha is written as a string in the manifest, all SHA's are pulled from the same timestring

markhannum: Gotcha- it's stored directly as 'PREV', not calculated .. cool

tools/comdb2ar/increment.cpp
```diff
@@ -0,0 +1,523 @@
...
+    const std::string& lrldestdir,
+    const std::string& datadestdir,
+    const std::string& dbname,
+    std::set<std::string>& table_set,
+    std::string& sha_fingerprint,
+    unsigned percent_full,
+    bool force_mode,
+    bool& is_disk_full,
+    const std::string& incr_path,
+    bool keep_all_logs
```
markhannum: I think we discussed this one - we should never keep all logs from all the increments: full-recovery will never work if there are holes in the logstream.  Keep only the logfiles from the last increment.

markhannum:
Missing another file:
make: *** No rule to make target 'tools/comdb2ar/incr_deserialise.o', needed by 'comdb2ar'.  Stop.

markhannum:
Dennis would you add your testcases to this?

tests/incremental_backup.test/runit
```diff
@@ -0,0 +1,140 @@
...
+  cdb2sql ${CDB2_OPTIONS} ${dbname}_restore default "select 1" > sqloutput 2>&1
+  sqloutput=$(echo sqloutput)
+  while [ $sqloutput != "\(1=1\)" -a $count -le 5 ]; do
+      sleep 1
+      let count=count+1
+      cdb2sql ${CDB2_OPTIONS} ${dbname}_restore default "select 1" > sqloutput 2>&1
+      sqloutput=$(echo sqloutput)
+  done
+
+  if [ $count -ge 30 ] ; then
```
markhannum: This should this be 5 (or the above loop should be 30)

tests/incremental_backup.test/runit
```diff
@@ -0,0 +1,140 @@
...
+
+# Keep track of which test we're on to do special stuff for DROP TABLE test
+counter=3
+
+# BASIC FUNCTIONALITY TEST
+resetdb
+for statement in `ls t1-*.stmt`; do
+  cdb2sql ${CDB2_OPTIONS} -f $statement $dbname default
+
+  # Need to force checkpoint for DROP TABLE test
```
markhannum: Force a checkpoint for every test, not just drop-table

tests/incremental_backup.test/runit
```diff
@@ -0,0 +1,140 @@
...
+function resetdb {
+  backuplist=()
+  rm -rf ${TMPDIR}/backups ${TMPDIR}/restore ${TMPDIR}/increment
+  mkdir -p ${TMPDIR}/backups ${TMPDIR}/restore ${TMPDIR}/increment
+
+  cdb2sql ${CDB2_OPTIONS} $dbname default "DROP TABLE t1"
+  sleep 1
+  make_base
+  cdb2sql ${CDB2_OPTIONS} $dbname default "CREATE TABLE t1  { `cat t1.csc2 ` }" || failexit "create failed"
+  sleep 1
```
markhannum: I wouldn't rely on these sleeps- force checkpoints for all of these.. maybe introduce a 'force_checkpoint' function.

markhannum:
Finally got a "real" bug: log-sequence-error ..

markhannum:
Checksum error seems fixed.  Investigating a log-sequence-error-on-recovery issue where the page-LSN is 0:0

markhannum:
I'm wrong- still seeing checksum errors ..

markhannum:
Thanks again Dennis!

PR #268, opened 2017-07-25
Get rid of block processor opcodes in error messages

ldennis:


markhannum:
Awesome - looks great - good to merge

PR #270, opened 2017-07-25
Local replication using SP

mohitkhullar:
poc


nirbhayc:
@mohitkhullar I have a few questions:
1) When does one need this? As a local backup?
2) Why cook this SP into the server code instead of providing it as a separate routine loaded by users when needed?

mohitkhullar:
This is needed for local copy of remote db. The SPs contain logic of comdb2rcv.tsk, the user now has to just add one line in lrl and they are all set.

db/comdb2.c
```diff
@@ -6740,6 +6754,7 @@ int main(int argc, char **argv)
     if (comdb2ma_stats_cron() != 0)
         abort();
 
+    local_rep_sched();
```
riverszhang89: Is it possible to call it only if there's a local replicant (to reduce the number of threads)?

PR #271, opened 2017-07-25
Virtual table to list thread pool information.

nirbhayc:


sqlite/ext/comdb2/threadpools.c
```diff
@@ -0,0 +1,288 @@
...
+
+    case COLUMN_MAX_QUEUE_OVERRIDE:
+        sqlite3_result_int(ctx, thdpool_get_maxqueueoverride(pool));
+        break;
+    case COLUMN_MAX_QUEUE_AGE_MS:
+        sqlite3_result_int(ctx, thdpool_get_maxqueueagems(pool));
+        break;
+    case COLUMN_EXIT_ON_CREATE_FAIL:
+        sqlite3_result_text(
+            ctx, YESNO(thdpool_get_exit_on_create_fail(pool) == 1), -1, NULL);
```
riverszhang89: Zero for false and non-zero for true. So you don't need to compare the value with 1.

nirbhayc: You're right. I will update that.

sqlite/ext/comdb2/threadpools.c
```diff
@@ -0,0 +1,288 @@
...
+    systbl_threadpools_cursor *pCur = (systbl_threadpools_cursor *)cur;
+    pCur->pool = thdpool_next_pool(pCur->pool);
+    pCur->rowid++;
+    return SQLITE_OK;
+}
+
+static int systblThreadPoolsEof(sqlite3_vtab_cursor *cur)
+{
+    systbl_threadpools_cursor *pCur = (systbl_threadpools_cursor *)cur;
+    return (pCur->rowid >= listc_size(&threadpools)) ? 1 : 0;
```
riverszhang89: You only need to `return (pCur->pool == NULL)` here as `thdpool_next_pool()` returns `NULL` if no next node.

riverszhang89:
We adhere to SQLite coding standards (2-space indentation, Hungarian notation and etc) when modifying the module. It is probably not required for our system tables, but is always a good practice.

PR #274, opened 2017-07-27
dont leak on error cases

adizaimi:
Plug two instances of leaks in case when there is an error condition.


mohitkhullar:
looks good

adizaimi:
Thank you Mohit.

PR #276, opened 2017-07-27
Specify versions for Docker and Compose

timmui:
Add docker and compose versions that are confirmed to be working to the docs. An older version was found to be incompatible. #skipbuild

### DIFF
--- a/contrib/docker/README.md
+++ b/contrib/docker/README.md
@@ -8,10 +8,10 @@
 This utility was built and tested using the following versions of Docker Engine and Compose:
 
 Docker Engine
-* version 17.06.0-ce, build 02c1d87
+* version 17.06.0-ce
 
 Docker Compose
-* version 1.14.0, build c7bdf9e
+* version 1.14.0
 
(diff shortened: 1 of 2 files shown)


timmui:
@mponomar If you have a different version of Docker/Compose, can you also add it to the list of tested versions? @markhannum discovered that an earlier version of Docker didn't work.

riverszhang89:
https://github.com/bloomberg/comdb2/blob/master/contrib/docker/docker-compose.yml#L2

The compose file is version 2 compliant. Simply change the version down to 2 and it will work on most versions of docker engine.

riverszhang89:
For reference: https://docs.docker.com/compose/compose-file/compose-versioning/#compatibility-matrix

mponomar:
```
[mike@xps /home/mike]$ docker --version
Docker version 17.06.0-ce, build 3dfb8343
[mike@xps /home/mike]$ docker-compose --version
docker-compose version 1.14.0, build unknown
[mike@xps /home/mike]$ 
```

PR #277, opened 2017-07-28
Cdb2sql: better behavior when entering ctrl-c

adizaimi:
Improve behavior when entering ctrl-c: if request is taking
too long, entering Ctrl-C will send cancel to server and wait for
server to return control. This way prompt ">" is really displayed
when cdb2sql really has control again. If query is taking longer
to return control (because server is corrupted for instance),
the only recourse user now has is to exit the program and that
can now be achieved by entering Ctrl-C again:
```
ubuntu@ip12345:~/comdb2$ ./cdb2sql biginplace26064 @localhost:port=19047 -
biginplace26064> ^C
biginplace26064> ^C
biginplace26064> ^C
biginplace26064> ^C
biginplace26064> select sleep(10)
^C
Requesting to cancel query (press Ctrl-C to exit program). Please wait...
^C
ubuntu@ip12345:~/comdb2$
```

### DIFF
--- a/tools/cdb2sql/cdb2sql.c
+++ b/tools/cdb2sql/cdb2sql.c
@@ -50,7 +50,8 @@ static char *dbname = NULL;
 static char *dbtype = NULL;
 static char *dbhostname = NULL;
 static char main_prompt[MAX_DBNAME_LENGTH + 2];
-static char gbl_in_stmt = 0;
+static unsigned char gbl_in_stmt = 0;
+static unsigned char gbl_sent_cancel_cnonce = 0;
 
 /* display mode */
 enum {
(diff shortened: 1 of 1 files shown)


adizaimi:
biginplace26064> select sleep(10)
^\Quit (core dumped)
Ctrl+\ is not very nice. 
Ctrl-U does not do anything in cdb2sql.

riverszhang89:
Okay then. Could you please change the prompt: It is either `Ctrl + C` or `^C`.

adizaimi:
It's correct in the patch. Thanks.

tools/cdb2sql/cdb2sql.c
```diff
@@ -929,9 +933,14 @@ void send_cancel_cnonce(const char *cnonce)
 static void int_handler(int signum)
 {
     printf("\n");
-    rl_on_new_line();
-    rl_replace_line("", 0);
-    rl_redisplay();
+    if (gbl_in_stmt && !gbl_sent_cancel_cnonce)
+        printf("Requesting to cancel query (press Ctrl-C to exit program). Please wait...\n");
```
riverszhang89: This line is definitely more than 80 characters.

adizaimi:
Thanks Rivers. Fixed in the #278 PR.

PR #279, opened 2017-07-28
Enable auto-registration of JDBC Driver.

riverszhang89:
This PR adds an explicit step in the POM file to copy the META-INF/services directory to the target JAR file. The PR includes a new services file with one line that contains the full class name of the Driver to auto-register. This avoids the need to explicitly load the class before using it (and therefore makes the JDBC driver work with the Kafka JDBC Connector).


riverszhang89:
The change is reviewed and tested internally.

PR #285, opened 2017-07-31
Blob size limit

mponomar:
Check blob sizes properly, don't allow them to exceed max size.


db/record.c
```diff
@@ -917,6 +929,18 @@ int upd_record(struct ireq *iq, void *trans, void *primkey, int rrn,
...
+        for (int i = 0; i < maxblobs; i++) {
+            if (blobs[i].exists && blobs[i].length > MAXBLOBLENGTH) {
+                reqerrstr(iq, COMDB2_ADD_RC_INVL_BLOB,
+                        "blob size (%d) exceeds maximum (%d)", blobs[i].length, MAXBLOBLENGTH);
+                retrc = ERR_BLOB_TOO_LARGE;
+                ERR;
+            }
+        }
+    }
+
```
nirbhayc: Could the above 2 maxblob checks be turned into a function?

db/sqlglue.c
```diff
@@ -866,6 +866,14 @@ static int mem_to_ondisk(void *outbuf, struct field *f, struct mem_info *info,
         }
     }
 
+    if ((f->type == SERVER_BLOB || f->type == SERVER_BLOB2 || f->type == SERVER_VUTF8) && m->n > MAXBLOBLENGTH) {
```
nirbhayc: roborivers' gonna complain about this line :)

mponomar: Good point...

nirbhayc: Also, can we flip this condition around && ? (m->n > MAXBLOBLENGTH && ( ... ))

db/tag.h
```diff
@@ -115,7 +115,8 @@ typedef enum convert_errcode {
     CONVERT_FAILED_INVALID_PARTIAL_TYPE,
     CONVERT_FAILED_BAD_BLOB_PROGRAMMER,
     CONVERT_FAILED_INPUT_TAG_HAS_INVALID_FIELD,
-    CONVERT_FAILED_INDEX_EXPRESSION
+    CONVERT_FAILED_INDEX_EXPRESSION,
+    CONVERT_FAILED_BLOB_SIZE
```
nirbhayc: (Optional) I prefer ending the last entry with a comma, so that next addition will result in a single-line diff.

db/record.c
mponomar: Not sure that'll make it clearer.  If anything, that code should be folded into check_blob_consistency

db/record.c
```diff
@@ -2895,6 +2887,21 @@ static int check_blob_buffers(struct ireq *iq, blob_buffer_t *blobs,
...
 
+static int check_blob_sizes(struct ireq *iq, blob_buffer_t *blobs, int maxblobs) {
+    int retrc = 0;
+
+    for (int i = 0; i < maxblobs; i++) {
+        if (blobs[i].exists && blobs[i].length > MAXBLOBLENGTH) {
+            reqerrstr(iq, COMDB2_ADD_RC_INVL_BLOB,
+                    "blob size (%d) exceeds maximum (%d)",
+                    blobs[i].length, MAXBLOBLENGTH);
+            retrc = ERR_BLOB_TOO_LARGE;
```
nirbhayc: On error, should be ok to return directly from here (no need for retrc). I'd also move this definition at the place of declaration.

mponomar:
Thanks Nirbhay.  Let's see if Roborivers approves.

bbinc/cdb2_constants.h
```diff
@@ -19,7 +19,7 @@
 
 #define COMDB2_MAX_RECORD_SIZE 16384
 #define LONG_REQMS 2000
-#define MAXBLOBLENGTH 256 * 1024 * 1024 /* TODO: set a good maximum here */
+#define MAXBLOBLENGTH 255 * 1024 * 1024 /* TODO: set a good maximum here */
```
riverszhang89: Comdb2sys [expected output](https://github.com/bloomberg/comdb2/blob/master/tests/comdb2sys.test/comdb2sys.expected#L277) needs to be adjusted.

db/record.c
```diff
@@ -2869,6 +2888,19 @@ static int check_blob_buffers(struct ireq *iq, blob_buffer_t *blobs,
     return 0;
 }
 
+static int check_blob_sizes(struct ireq *iq, blob_buffer_t *blobs, int maxblobs)
+{
+    for (int i = 0; i < maxblobs; i++) {
+        if (blobs[i].exists && blobs[i].length > MAXBLOBLENGTH) {
```
riverszhang89: `blobs[i].length` is `size_t`. We set blob length to `-2` if the blob is not used by the query. Therefore you need to cast `blobs[i].length` to signed.

mponomar: Thanks Rivers.  That's what the blob[i].exists is for - that should be set to 0 if the blob ins't used.

riverszhang89: Yeah, that'd be ideal. Unfortunately we don't always set it to 0.

PR #287, opened 2017-08-01
fix functional/partial index asserts

ldennis:
1. I got an optimization which passes in NULL blob if a partial index/expression index doesn't involve blobs so that we could avoid fetching unnecessary blobs. So the assertion should really be `assert(!blobs || f->blob_index < maxblobs);`

2. If the index is a byte array, return SQLite is blob. The assertion was wrong and now removed.


adizaimi:
looks great. Thanks

PR #290, opened 2017-08-01
Genid48 unit test

markhannum:
I wrote this in an effort to understand some behavior that our Jepsen testing unearthed.  Genid48 code turned out not to be at fault.


nirbhayc:
Runtime-only no? I did not find a use-case for that. The tunables not marked as READONLY can be set at start via lrl or, later, at runtime. Said that I can add the support for runtime-only if that's something we really need.

mponomar:
It's more of a command to run, not a tunable (no persistent value to set).

nirbhayc:
How about supporting arbitrary user variables using SET with connection life span? They can be used in such scenarios.
```
> SET user-var val;
```

mponomar:
SET is for per-session settings.  PUT has been culturally appropriated for permanent settings.  This is neither.  It's fine for this to exist as a command since we're not likely to use this outside of testing (or if we ever have to reset a genid48 sequence because we run out)

PR #291, opened 2017-08-02
Fix qcopy test

mohitkhullar:
cdb2sql will give error until db is up, however the script exits on any error

### DIFF
--- a/tests/qcopy.test/runit
+++ b/tests/qcopy.test/runit
@@ -1,7 +1,7 @@
 #!/bin/bash
 bash -n "$0" | exit 1
 
-set -e
+set -x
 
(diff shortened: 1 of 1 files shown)


tests/qcopy.test/runit
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 bash -n "$0" | exit 1
 
-set -e
+set -x
```
mponomar: I meant -e, but didn't account for cdb2sql failing because the db isn't up yet.   If you don't mind, I'll change this to expect cdb2sql to fail, but not the rest of the script.

mohitkhullar: ok

PR #292, opened 2017-08-02
Poll in pgfree recover

markhannum:
Introduce artificial latency in the page-free recovery handler.  If the page-free transaction's commit-record doesn't list the page-lock for the free'd page, then a parallel-rep thread can reuse the page before it's added to the free-list.  Stock-berkley doesn't have parallel rep, and therefore doesn't need to acquire the pagelock for the free'd page in the replicant.


linearizable/scripts/recreatedb
```diff
@@ -5,6 +5,18 @@ if [[ -n "$1" ]]; then
...
 fi
 
+lrlpath=""
+if [[ -n "$1" ]]; then
+    lrlpath=$1
+    shift
+fi
+
+add_tables=1
+if [[ "$1" == "noadd" ]]; then
```
nirbhayc: So $1 can be "noadd" or lrlpath? Where's recreatedb script sourced/used?

tests/pg_free_recovery.test/runit
```diff
@@ -0,0 +1,149 @@
...
+{
+    [[ "$debug" == 1 ]] && set -x
+    [[ "$trace" == 1 ]] && echo "Running deleteall"
+    xxx=$(cdb2sql ${CDB2_OPTIONS} $dbname $stage "delete from t1 limit 10000") ; xx=${xxx##*=} ; x=${xx%\)}
+    while [[ -n "$x" && "$x" != 0 ]]; do
+        xxx=$(cdb2sql ${CDB2_OPTIONS} $dbname $stage "delete from t1 limit 10000") ; xx=${xxx##*=} ; x=${xx%\)}
+    done
+}
+
+function beforedg
```
nirbhayc: I'd rename this..:) took me a few seconds to figure.

linearizable/scripts/recreatedb
markhannum: It was a convenience script I used while testing for linearizability.  I'm now moving these tests to our normal testsuite so this will be removed very soon.

markhannum:
THANK YOU NIRBHAY!

PR #294, opened 2017-08-02
Put get_genid into a message-trap.

markhannum:
Nirbhay pointed out that get_genid isn't actually a tunable, and probably shouldn't be mis-stated as such.

### DIFF
--- a/db/process_message.c
+++ b/db/process_message.c
@@ -5102,7 +5102,8 @@ int process_command(struct dbenv *dbenv, char *line, int lline, int st)
         int *genptr = (int *)&genid;
         flipptr[0] = htonl(genptr[1]);
         flipptr[1] = htonl(genptr[0]);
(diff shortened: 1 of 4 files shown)


db/process_message.c
```diff
@@ -5089,6 +5089,20 @@ int process_command(struct dbenv *dbenv, char *line, int lline, int st)
...
+            return -1;
+        }
+        int dtafile = toknum(tok, ltok);
+        unsigned long long flipgenid = 0;
+        unsigned long long genid = get_genid(thedb->bdb_env, dtafile);
+        int *flipptr = (int *)&flipgenid;
+        int *genptr = (int *)&genid;
+        flipptr[0] = htonl(genptr[1]);
+        flipptr[1] = htonl(genptr[0]);
+        logmsg(LOGMSG_USER, "0x%016llx 0x%016llx %llu\n", genid, flipgenid, genid);
```
nirbhayc: You might want to break this line.

markhannum:
Thanks Nirbhay!

PR #297, opened 2017-08-03
cdb2sql readline completion

adizaimi:
Uses a virtual table to get the possible completion tokens. 
It does not do a parse tree completion (that would be much harder project), but for now only queries 
the db for possible tokens (including tables and column names) regardless of the current line.
May need a switch to turn this feature on, also may want to put functionality in a separate file.


riverszhang89:
Can this be done using existing system tables? Since we already have comdb2_keywords, it seems needless to introduce another "completion" virtual table.

adizaimi:
Good point. The same could probably be achieved by sending query that UNIONs comdb2_keywords with comdb2_tables and comdb2_columns. I was hoping in future to get some state into this, so the completion list would be the correct subset of the kewords based on the current input and the parse tree -- and that should be done on the server end.

adizaimi:
5cfdbf1 does what you say Rivers, but i want to get this to work in a more fancy way in future, so will need that new vtable.

sqlite/ext/misc/completion.c
```diff
@@ -0,0 +1,583 @@
...
+            if( zSql==0 ) return SQLITE_NOMEM;
+            zSep = " UNION ";
+          }
+          sqlite3_finalize(pS2);
+          sqlite3_prepare_v2(pCur->db, zSql, -1, &pCur->pStmt, 0);
+          sqlite3_free(zSql);
+          */
+          sqlite3_prepare_v2(pCur->db, 
+                  "SELECT tablename FROM comdb2_tables "
+                  "WHERE tablename NOT LIKE 'sqlite_stat\%' ORDER BY 1",
```
mponomar: Don't think you need to escape '%' - not in printf

sqlite/ext/misc/completion.c
```diff
@@ -0,0 +1,583 @@
...
+            if( zSql==0 ) return SQLITE_NOMEM;
+            zSep = " UNION ";
+          }
+          sqlite3_finalize(pS2);
+          sqlite3_prepare_v2(pCur->db, zSql, -1, &pCur->pStmt, 0);
+          sqlite3_free(zSql);
+          */
+          sqlite3_prepare_v2(pCur->db, 
+                  "SELECT columnname FROM comdb2_columns " 
+                  "WHERE tablename NOT LIKE 'sqlite_stat\%' ORDER BY 1",
```
mponomar: No need to escape.

sqlite/ext/misc/completion.c
```diff
@@ -0,0 +1,583 @@
...
+                  "WHERE tablename NOT LIKE 'sqlite_stat\%' ORDER BY 1", 
+                  -1, &pCur->pStmt, 0);
+        }
+        iCol = 0;
+        pCur->j = 0;
+        eNextPhase = COMPLETION_FUNCTIONS;
+        break;
+      }
+      case COMPLETION_FUNCTIONS: {
+        static char *cfuncs[] = {"comdb2_version()", "table_version()", "partition_info()", "comdb2_host()", "comdb2_port()", "comdb2_dbname()", "comdb2_prevquerycost()", "sys.cmd.send()"};
```
mponomar: May want to fetch from SQLite

tools/cdb2sql/cdb2sql.c
```diff
@@ -144,6 +142,182 @@ void cdb2sql_usage(int exit_val)
...
+        }
+        if (rc) {
+            if (debug_trace)
+                fprintf(stderr, "cdb2_open rc %d %s\n", rc, cdb2_errstr(cdb2h));
+            cdb2_close(cdb2h_2);
+            return ((char *) NULL);
+        }
+        rc = cdb2_run_statement(cdb2h_2, sql);
+        if (rc) {
+            fprintf(stderr, "failed to run sql '%s'\n", sql);
```
riverszhang89: It is okay if the statement can't be auto-completed (for instance, the backend is R6 which does not have completion table). Hence we should silently ignore instead of printing an error here.

riverszhang89: In cdbsql CLI, type 
select * fro <kbd>TAB</kbd>

And database will crash.

The pull request needs more work.

adizaimi:
Thanks Rivers, crash comes from assert (now that master is doing asserts).
https://github.com/bloomberg/comdb2/pull/316 will fix that.

PR #299, opened 2017-08-03
Reverse commit genid and cleanup

markhannum:
Incremental work on durable-LSNs:
* Remove durable-block code: if a node isn't up-to-date with the most recent durable LSN, ask the client to retry against a different node
* Remove UDP broadcast of durable lsns: we've changed the design to make the master is the sole authority on this.
* abort-on-ridiculous set_durable_lsn
* Zero'ing the commit-context in rep_verify_match is longer needed: a generation change across curtrans is sufficient to fail and restart an sql-session.
* Add tunable to abort if clnt->high_availability is incorrectly set: I don't understand why this is happening yet.
* Other various instrumentation & tunables.


markhannum:
Akshat reviewed these changes under a different pr

PR #300, opened 2017-08-03
Fix qcopy testcase.

mponomar:


mohitkhullar:
you might want to remove || after cdb2sql command which is handling the error case right now.

PR #301, opened 2017-08-04
tests should not load global lrl files

adizaimi:
This is needed so tests are isolated from machine setup, but also enables us to test the --no-global-lrl parameter.

### DIFF
--- a/tests/tunables.test/runit
+++ b/tests/tunables.test/runit
@@ -9,6 +9,7 @@ fi
 
 for testreq in `ls t*.sql` ; do
     testname=`echo $testreq | cut -d "." -f 1`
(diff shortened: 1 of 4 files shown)


tests/tunables.test/runit
```diff
@@ -9,6 +9,7 @@ fi
 
 for testreq in `ls t*.sql` ; do
     testname=`echo $testreq | cut -d "." -f 1`
+    #cmd="cdb2sql --tabs -s -f $testreq ${CDB2_OPTIONS} $dbname default "
```
nirbhayc: Leftover?

adizaimi: no, needed for later.

Issue #286, opened 2017-07-31
failed create index 


akshatsikarwar:
Trying to use some of the new features and see an unusual error message. Putting here to so I don't forget. Possibly, bug in my branch for which I was writing a testcase.

```
indexes have been added or removed
No index I in old schema
No index I in old schema
[ERROR] verify_new_temp_sc_db: bdb_get_file_version_data failed for newdb data 0
[ERROR] Schema change error: some of the newdb's file versions are the same or less than some of db's.
[ERROR] Schema change error: we will delete the newdb's file version entries from llmeta and quit.
[ERROR] Schema change error: the new master should be able to resume this sc starting over with a fresh newdb
waiting for threads 2/2/3 num queue 0
busy  tid -532228352  time   343 ms  blockop (100) SORESE#          (null) SOCKSQL RQST 1 where toblock bdb_tran_begin done
busy  tid -694118656  time    11 ms  blockop (100) SORESE#          (null) SOCKSQL RQST 1 where done executing bdb_wait_for_seqnum_from_all done
idle  tid -689903872 
idle  tid -677259520 
idle  tid -685689088 

Waiting for threads to stop
3 registered threads running:-
   0) tid 8555:req, SORESE#          (null) SOCKSQL RQST 1: blockop [BLOCK2_SEQV2]
   1) tid 8577:appsock-pool-sql, fd 38 (127.0.0.1): waiting for query
   2) tid 8588:sql-engine-pool, fd 38 (127.0.0.1), rqid=0000000000000001 4ffce365-fcff-419: sql: create index i on t(i)
------
```

akshatsikarwar:
And today:
```
indexes have been added or removed          
No index I in old schema                    
No index I in old schema                    
tag .ONDISK has new field 1 (named j)       
[ERROR] Couldn't save schema change seed    
[ERROR] Couldn't save schema change seed    
[ERROR] Schema change error: Table doesn't exists                           
```

akshatsikarwar:
I'll try if occurs with CSC2 schemachange as well.

akshatsikarwar:
Fixed

PR #302, opened 2017-08-04
DDL schema locking

akshatsikarwar:
Should address issue #286

### DIFF
--- a/db/osqlblockproc.c
+++ b/db/osqlblockproc.c
@@ -1231,6 +1231,7 @@ int osql_bplog_reqlog_queries(struct ireq *iq)
     return 0;
 }
 
+static pthread_mutex_t ddl_lk = PTHREAD_MUTEX_INITIALIZER;
 static int apply_changes(struct ireq *iq, blocksql_tran_t *tran, void *iq_tran,
                          int *nops, struct block_err *err, SBUF2 *logsb)
 {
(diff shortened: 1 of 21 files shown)


nirbhayc:
Looks good to me. I'd suggest not to squash the commits while merging.

akshatsikarwar:
holy moly

db/osqlblockproc.c
```diff
@@ -1279,17 +1280,14 @@ static int apply_changes(struct ireq *iq, blocksql_tran_t *tran, void *iq_tran,
     listc_init(&iq->bpfunc_lst, offsetof(bpfunc_lstnode_t, linkct));
 
     /* go through the complete list and apply all the changes */
+    if (iq->tranddl) pthread_mutex_lock(&ddl_lk);
```
ldennis: Can we have an option to enable nonblocking on this with trylock? IIRC, we used to return "schema change in progress" error.

akshatsikarwar: Sure - that's a good idea.

akshatsikarwar:
Failure in `reqlog` accessing free'd memory looks legit. `pglogs_asof` doesn't look related.

ldennis:
Looks good

ldennis:
With this patch, I wonder if the broken sc seed is still worth fixing.

akshatsikarwar:
why does `sc_resume` keep failing?

riverszhang89:
Not sure. master is failing on that test as well.

PR #303, opened 2017-08-04
Remove no_transaction flag

akshatsikarwar:
One less kludge flag to remember

### DIFF
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -4831,7 +4831,7 @@ int sqlite3BtreeBeginTrans(Vdbe *vdbe, Btree *pBt, int wrflag)
 #endif
 
     /* already have a transaction, keep using it until it commits/aborts */
-    if (clnt->intrans || clnt->no_transaction ||
+    if (clnt->intrans || vdbe == NULL ||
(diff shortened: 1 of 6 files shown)


db/sqlglue.c
```diff
@@ -4831,7 +4831,7 @@ int sqlite3BtreeBeginTrans(Vdbe *vdbe, Btree *pBt, int wrflag)
 #endif
 
     /* already have a transaction, keep using it until it commits/aborts */
-    if (clnt->intrans || clnt->no_transaction ||
+    if (clnt->intrans || vdbe == NULL ||
```
ldennis: Does this handle cases of `reload_analyze` under snapshot isolation? Would clnt->intrans be set to 1 later in this function? (I think [Line 4901](https://github.com/bloomberg/comdb2/pull/303/files#diff-c3ebbfffbf6cf6842b5727af7e3a5e99R4901) only checks for mode <= RECOM,)

akshatsikarwar:
You're right, maybe we need this flag after all.

ldennis:
Well, maybe we could do the R6 trick to temporarily change dbtran mode to lower mode for `reload_analyze`?

PR #305, opened 2017-08-04
Add comdb2_gitversion() function

adizaimi:
Add git version to the executable, this is useful to determine build.
The original intention was to make sure that we are running the same 
server executable in all the nodes of a cluster. 

Turns out that this is not useful to check from setup: if you commit a change
the git version will change, however the executable was built earlier with
a previous git hash--setup can't rely on hash because of mismatch
between git version in the executable, and the version in current dir.
Thus, i am moving this to a check in the basic.test so we can
detect the mismatch but only in that one test.

To try and catch wrong executable, we would need to take md5sum of
executable at the beginning of the tests, and check with every
running setup that the md5sum matches.

### DIFF
--- a/sqlite/func.c
+++ b/sqlite/func.c
@@ -675,7 +675,7 @@ static void guidFromByteFunc(
 
 
 /*
-** Implementation of the comdb2_version() SQL function.  The return
+** Implementation of comdb2_version() SQL function.  The return
 ** value is the same as the stat value for version
 */
 static void comdb2VersionFunc(
@@ -687,6 +687,21 @@ static void comdb2VersionFunc(
   char zBuf[50];
   const char * version = getenv("COMDB2_VERSION");
   sqlite3_snprintf(sizeof(zBuf), zBuf, "%s (%s)", gbl_db_release_name, version);
+  sqlite3_result_text(context, zBuf, -1, SQLITE_TRANSIENT);
+}
+
+/*
+** Implementation of comdb2_gitversion() SQL function. 
+** Uses git describe to get git repository version in makefile
+*/
+static void comdb2GitversionFunc(
+  sqlite3_context *context,
+  int NotUsed,
+  sqlite3_value **NotUsed2
+){
+  UNUSED_PARAMETER2(NotUsed, NotUsed2);
+  char zBuf[50];
+  sqlite3_snprintf(sizeof(zBuf), zBuf, "%s", COMDB2_GITVERSION);
   sqlite3_result_text(context, zBuf, -1, SQLITE_TRANSIENT);
 }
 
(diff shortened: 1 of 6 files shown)


riverszhang89:
Adi, I really appreciate your work. However I can't say I like the patch.

First of all, I could not think of any useful use cases for the function. You said,

> The original intention was to make sure that we are running the same server executable in all the nodes of a cluster.

This is completely unnecessary and even unrealistic: when a cluster is upgraded to a newer executable or downgraded to an older executable, it is certain that not all nodes are running the same executable during the operation.

Second of all, the patch can't even guarantee your goal of making sure all nodes are running the same exectable, as you pointed it out yourself.

Therefore I must revert the patch.

riverszhang89:
On a second thought, it'd be useful if you could embed the information in `comdb2_version()` instead of creating a new function.

nirbhayc:
Not in comdb2_version() please! Sometimes client apps, and even drivers, parse server version to solve compatibility issues. Git hash, as I see it, is exclusive to testing, so let's not add it to server version.

adizaimi:
`when a cluster is upgraded to a newer executable or downgraded to an older executable, it is certain that not all nodes are running the same executable during the operation.` 
Rivers this was only meant to ensure same version during *testing*, not normal db operations. For that i have added m5sum checking of executable. 

git hash (maybe should not call it version) is useful in of itself--whether we make it a function or printed from 'send stat'. Functions are cheap and they don't slow things down.

riverszhang89:
Nirbhay: I agree that git hash should not be added to server version. However the command Adi used in the patch is _`git describe`_. It returns the most recent tag.

`SELECT comdb2_version()` is supposed to return a string in the form of  `<Major_Release> (<Major_Release>.<Tag_Name>)`. However, currently we only get an incomplete version string `"R7.0-pre ()"`. The reason is that we don't actually compile the tag name into the executable.

For R6 and below, an umbrella executable called `comdb2driver` injects the tag name at runtime. Since comdb2driver is not open sourced (and probably should not be as it is a highly customized application for Bloomberg), I highly recommend we compile the information returned by `git describe` into comdb2 executable.

riverszhang89:
(contd.) and make it part of `comdb2_version()`.

riverszhang89:
Adi: if it is for testing only, you don't need to make any changes for that, just use my poor man's `pwhat`:

```
$ readelf -a comdb2 | grep s24 | awk '{print $7}' | xargs readelf comdb2 -p | grep SVNINFO
[ 2ea40]  @(#)plink [comdb2] SVNINFO : git@example.com/placeholder/til/comdb2submit/fix (master) tag  (83a190c)
```

adizaimi:
Well, if you need to ask the running db what version it thinks it is running, that's not useful. 
For better checking of executable version I implemented md5sum check at the beginning of each test. 
At this point the only thing left to is how we want to be querying the db what git version it is running.

riverszhang89:
> Well, if you need to ask the running db what version it thinks it is running, that's not useful.

I agree and I propose we use `comdb2_version()`.

mponomar:
There's applications at Bloomberg that use comdb2_version() as a feature check, so changing the format of that will likely break things for them.  It should really be a different function.  comdb2_build()?

riverszhang89:
I don't think we need a new function for this. `comdb2_version()` is supposed to return the git tag, for instance, `"R6 (R6.R20170725)"`, where `"R6.R20170725"` is a git tag.

PR #307, opened 2017-08-05
check md5sum of comdb2 executable for each test

adizaimi:
Make sure that we run the same executable for all the tests. Changing the comdb2 executable in the src directory will cause the next testcase of the already-running-test to fail.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -12,6 +12,7 @@ export SRCHOME
 export TESTID:=$(shell tools/get_random.sh)
 CURRDIR:=$(shell pwd)
 export TESTDIR:=$(CURRDIR)/test_$(TESTID)
+# get md5sum of executable use check when running every test in a series
 export COMDB2MD5SUM:=$(shell md5sum ${SRCHOME}/comdb2 | cut -d ' ' -f1)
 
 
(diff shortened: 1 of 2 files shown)


tests/setup
```diff
@@ -214,10 +214,22 @@ fi
 #    exit 1
 #fi
 
+compare_md5_sum()
+{
+    loc_comdb2md5sum=$1
```
nirbhayc: I think you can skip loc_comdb2md5sum and use $1 directly.

tests/setup
```diff
@@ -214,10 +214,22 @@ fi
...
+    if [ "$loc_comdb2md5sum" != "$COMDB2MD5SUM" ]; then
+        echo "MD5SUM $COMDB2MD5SUM does not match current $loc_comdb2md5sum"
```
nirbhayc: "comdb2 md5 checksum mismatch, expected: $COMDB2MD5SUM got: $loc_comdb2md5sum" ?

tests/Makefile
```diff
@@ -12,6 +12,7 @@ export SRCHOME
 export TESTID:=$(shell tools/get_random.sh)
 CURRDIR:=$(shell pwd)
 export TESTDIR:=$(CURRDIR)/test_$(TESTID)
+export COMDB2MD5SUM:=$(shell md5sum ${SRCHOME}/comdb2 | cut -d ' ' -f1)
```
nirbhayc: Please add a comment for this.

adizaimi:
thanks Nirbhay.

PR #308, opened 2017-08-05
Revert 305 az comdb2 giversion func

riverszhang89:
See my comment in #305 . Thanks.


riverszhang89:
I don't understand why there're 3 revert commits showing up here. I will work from CLI.

PR #309, opened 2017-08-07
Fix election bug

markhannum:
Address election bug where an out-of-date node can incorrectly become master.


bdb/genid.c
```diff
@@ -1046,14 +1063,10 @@ void bdb_set_commit_genid( bdb_state_type *bdb_state, unsigned long long context
 
     Pthread_mutex_lock(&(bdb_state->gblcontext_lock));
```
markhannum: When will plsn and generation be null?

db/sqlinterfaces.c
```diff
@@ -7397,6 +7393,17 @@ CDB2QUERY *read_newsql_query(struct sqlclntstate *clnt, SBUF2 *sb)
     return query;
 }
 
+int gbl_debug_high_availability_flag = 0;
+static inline void set_high_availability(struct sqlclntstate *clnt, int val)
+{
+    clnt->high_availability = val;
```
markhannum: Maybe rename this to make sure there are no other direct accesses so that you can verify that you caught all of these

markhannum:
Reviewed with Akshat in person.

PR #311, opened 2017-08-08
DDL hardening

nirbhayc:
- Comment unused parser constructs
- Force PKs columns NOT NULL
- Explicit check for multiple PK definitions


nirbhayc:
All failing tests passed on my system except for qcopy (which tries to exec copycomdb2, not in PATH).

sqlite/comdb2build.c
```diff
@@ -2311,6 +2312,13 @@ static char *gen_key_name(struct comdb2_ddl_context *ctx, const char *tabname,
     return keyname;
 }
 
+#define PK "PRIMARY_KEY"
```
akshatsikarwar: Consider using `COMDB2_PK` or similar to avoid name clashes with existing instances.

nirbhayc: Sure

nirbhayc: @akshatsikarwar thought you were suggesting to rename the proprocessor identifier "PK". How about prefixing with a $, "$PRIMARY_KEY"?

PR #312, opened 2017-08-08
Docker configuration modifications

timmui:
Changes to Docker configuration to make it easier to develop/debug on.
* Add Debug utils to `comdb2-dev` (`gdb, vim, iputils, strace`)
* docker-cluster image change to `comdb2-dev` for debug utils
* Remove build containers after they finish building source
* Add [all privileges](https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities) to `docker-cluster` containers
* Isolate build container source from host - allows for the host and docker to run `make` without interfering with each other

### DIFF
--- a/contrib/docker/Dockerfile.dev
+++ b/contrib/docker/Dockerfile.dev
@@ -2,6 +2,8 @@ FROM comdb2-standalone:7.0.0pre
 
 ENV PATH      $PATH:/opt/bb/bin
 
+RUN apt-get update && apt-get install -y vim gdb iputils-ping strace
+
 EXPOSE 5105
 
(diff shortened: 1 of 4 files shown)


contrib/docker/Dockerfile.dev
```diff
@@ -2,6 +2,10 @@ FROM comdb2-standalone:7.0.0pre
 
 ENV PATH      $PATH:/opt/bb/bin
 
+RUN apt-get update && \
+    apt-get install -y vim \
+    gdb iputils-ping strace
```
nirbhayc: All should fit on a single line.

timmui: Good call

PR #313, opened 2017-08-08
Fix running make from a testdir

adizaimi:
Currently, running make from a testdir is cumbersome and one has to provide testid and testrootdir.
But that is unnecessary and i am fixing in this PR. Now one can simply run make from a testdir, and it will automatically set testid and testrootdir. 
In order for those variables to be automatically set (by testcase.mk), a given test's Makefile will need to start with:
```sh
ifeq ($(TESTSROOTDIR),)
   include ../testcase.mk
else
   include $(TESTSROOTDIR)/testcase.mk
endif
```
See `basic.test` for an example.


tests/basic.test/Makefile
```diff
@@ -1,2 +1,6 @@
+ifeq ($(TESTSROOTDIR),)
+  export TESTSROOTDIR=$(shell readlink -f $(PWD)/..)
```
mponomar: Let's just go with ..?

tests/testcase.mk
```diff
@@ -1,6 +1,9 @@
+ifeq ($(TESTSROOTDIR),)
```
mponomar: Maybe it's time we went through all the tests and changed it to include ../testcase.mk?

adizaimi: Ok, can do as as separate PR.

PR #314, opened 2017-08-08
Fix table schema locking in stored procedures

akshatsikarwar:


riverszhang89:
Looks like master is broken again.

akshatsikarwar:
pr needs a bit more work too

akshatsikarwar:
Going to skip drop table test which depends on Dorin's fix. See a couple of additional issues as well.

PR #317, opened 2017-08-09
Documentation for recent DDL additions.

nirbhayc:
#skipbuild


docs/pages/programming/sql.md
```diff
@@ -338,23 +365,45 @@ will no longer be callable from SQL.
...
+section a list of options that may be set for a table.
+
+The second approach, added in version 7.0, supports the usual standard data
+definition language, like other relational database systems. This syntax can
+be used to ```ADD``` a new column or ```DROP``` an existing column from the
+table. Multiple ADD/DROP operations can be used in the same command. In case of
+```DROP``` operation, the references to the column begin dropped will be
+silently removed from the referring keys and constraints definitions.
+
+**Schema changes in Comdb2 are always live**. The database will not acquire
```
riverszhang89: Make it the 1st paragraph of Alter Table section? Superior to other database systems, schema change in Comdb2 was designed with the goals of being live, least disruptive to normal database operations, and resilient to database crashes. In addition, certain types of schema change can even be completed instantly. It is one of the most exciting features of Comdb2. We might want to mention it first.

docs/pages/programming/table_schema.md
```diff
@@ -7,12 +7,26 @@ permalink: table_schema.html
...
+* [Constants](#constants-section)
+
+* [Table definition](#schema-section)
+
+* [Keys](#key-section)
+
+* [Constraints](#constraints-section)
+
+Comments begin with ```//``` and go until the end of the line. Multi-line
+comments are not supported.
```
riverszhang89: This is not true. Multi-line comments are supported. They are started with `/*` and ended with `*/`.

PR #322, opened 2017-08-09
Change net `ack_state_type` free location

timmui:
Change net ack free location so acks don't have to be inline/blocking.
The `ack_state_type` is no longer freed at the end of execution of a userfunc that requires an ack. Instead, the ack-ing functions will do the job of freeing the `ack_state_type`.
This way, threads can be spawned to handle some long task then ack the message at the end rather than blocking other threads.

Added `ack_handle = NULL` or similar to all usages of the two ack functions as a reminder that has been freed.

### DIFF
--- a/bdb/rep.c
+++ b/bdb/rep.c
@@ -3990,6 +3990,7 @@ void receive_start_lsn_request(void *ack_handle, void *usr_ptr, char *from_host,
         logmsg(LOGMSG_ERROR, "%s returning bad rcode because i am not master\n", 
                 __func__);
         net_ack_message(ack_handle, 1);
+        ack_handle = NULL;
         return;
     }
 
@@ -3998,6 +3999,7 @@ void receive_start_lsn_request(void *ack_handle, void *usr_ptr, char *from_host,
         logmsg(LOGMSG_ERROR, "%s returning bad rcode because i don't have enough "
                 "leases\n", __func__);
         net_ack_message(ack_handle, 2);
+        ack_handle = NULL;
         return;
     }
 
@@ -4011,6 +4013,7 @@ void receive_start_lsn_request(void *ack_handle, void *usr_ptr, char *from_host,
                              "durable_gen=%d\n",
                __func__, __LINE__, current_gen, start_lsn.gen);
         net_ack_message(ack_handle, 3);
+        ack_handle = NULL;
         return;
     }
 
@@ -4031,6 +4034,7 @@ void receive_start_lsn_request(void *ack_handle, void *usr_ptr, char *from_host,
     }
 
     net_ack_message_payload(ack_handle, 0, buf, START_LSN_RESPONSE_TYPE_LEN);
+    ack_handle = NULL;
     return;
 }
 
@@ -4220,6 +4224,7 @@ void berkdb_receive_msg(void *ack_handle, void *usr_ptr, char *from_host,
         /* This version of comdb2 shouldn't be getting these messages */
         if (ack_handle) {
             net_ack_message(ack_handle, 0);
+            ack_handle = NULL;
         }
         break;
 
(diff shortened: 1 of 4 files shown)


timmui:
@markhannum

riverszhang89:
Not sure if I'm following this. How come that changing the location of `free()` makes acks non-blocking? In addition, `process_user_message()` could leak memory if you don't free `ack_state` but instead setting it to NUL.

timmui:
Changing the location of where it is freed makes it possible to pass the ack_state into a separate thread , where it will be freed by one of the ack functions when they are called instead of being freed upon completion of the handler on the main thread. Since ack_state is only allocated memory if an ack is required, the free can be handled in the ack methods. Setting them to NULL is just a reminder that it is no longer usable because it was freed in the ack method immediately before it.

riverszhang89:
> to pass the ack_state into a separate thread , where it will be freed by one of the ack functions when they are called

Somehow it sounds even worse. Ack packets are so tiny that we definitely don't want to spawn threads just for that. And even if we did so, the bottle neck would still be our net interface.

riverszhang89:
https://github.com/bloomberg/comdb2/pull/322/files#diff-158d650f8c314f5cf37ed472eed53b5bR3712

This is the leak I was referring to.

timmui:
This change allows for the flexibility to spawn a new thread. No current net handlers actually do that. This is needed as a prerequisite to sequence allocation because bdb locks need to be obtained and we don't want to hold up the main thread. In that new case, we will create a new thread.

Also, there isn't a leak there because [`net_ack_message()` and `net_ack_message_payload()`](https://github.com/bloomberg/comdb2/pull/322/files#diff-158d650f8c314f5cf37ed472eed53b5bR3525) are the ones freeing the handle now. They are the only functions that use the handle and the `NULL` is there as a visual indicator to who ever is using it that the handle has been freed/

markhannum:
Not sure what's right .. updating llmeta  directly from the net handler seems wrong: the thread isn't set up to grab the bdb-lock, and this is running directly from the reader thread.  Dispatching a thread every time we grab a sequence also seems wrong.. 

Maybe signal a dedicated sequence thread?  As long as we're not touching llmeta we could return inline & dispatch to a thread only for the "allocate next chunk" case ..

riverszhang89:
You are right. That isn't a leak. Sorry that I misread the code.

timmui:
What happens is that it is being ack-ed inline for all cases and the one that would need a thread (where ever it comes from) would be the case where the net message needs to get the lock like in the case of sequences.

timmui:
@markhannum I think the "middle ground" of a sequence thread pool would make sense. Regardless, this change would be needed to pass the handle along

mponomar:
Or get it out of the net layer altogether and make a new appsock request type to request sequence numbers.

riverszhang89:
I'm with Mike.

markhannum:
You'd have to create a new connection to the master every time we wanted to allocate a sequence- that's pretty heavy weight.  Even if sockpool fixes that issue perfectly, there could be resource contraints that we'll run up against with having so many additional active sockets against the master (i.e., the number of ephemeral ports).  The resource-contraint issue becomes worse if we're moving to machines that are intended to house substantially more databases than what we currently have.  The upside is that it's a first-step towards the "get-rid-of-rqid" project that we were discussing the other day .. alot of code goes away, and alot of other code becomes cleaner.

mponomar:
Connections can persist for multiple requests, just like they do for SQL.   You can keep a list of active connections per replicant.  My primary motivation for wanting this is getting rid of lots of confusing code.  The net layer fits BerkeleyDB's model well, it's difficult to manage for request/response.

markhannum:
This merits further discussion- if we ultimately decide to go this route we'll include this code as part of a larger pull request that shows the code being used as intended.  Withdrawing this for now.

PR #324, opened 2017-08-10
Prepend all make files of all tests so will run from testdir

adizaimi:


ldennis:
This one made robo return `Regression testing: 0/132 tests failed` instead of a `Success ✓`?

adizaimi:
Hmm...not sure what is the deal with that.

adizaimi:
found it -- one of the makefiles was bad. fixed in master and testing now

PR #327, opened 2017-08-10
Clean intrans flag on clnt to force a new rqid.

riverszhang89:


riverszhang89:
We need to understand what caused the crash.

PR #328, opened 2017-08-11
Port linearizable tests to testsuite

markhannum:
Placed 'linearizable' directory under tests, ported linearizable tests to normal test framework.


riverszhang89:
New roborivers does not have java environment. I will install it tomorrow. However, it is going to be a bit difficult to test paritition healing since it requires sudo to change iptables.

markhannum:
Agreed- I want to smoke test this first for a while before doing that.

markhannum:
This should be ready to merge

berkdb/rep/rep_method.c
```diff
@@ -904,7 +904,8 @@ __rep_elect(dbenv, nsites, priority, timeout, eidp)
 	rep = db_rep->region;
 	dblp = dbenv->lg_handle;
 
-	if (use_committed_gen == dbenv->attr.elect_highest_committed_gen) {
```
akshatsikarwar: sorry

markhannum: NO WORRIES :)

markhannum:
I'm going to annotate these changes to make a code review easier

bdb/cursor.c
```diff
@@ -7686,6 +7686,12 @@ int bdb_osql_trak(char *sql, unsigned int *status)
         *status |= SQL_DBG_SHADOW;
         return 0;
     }
+
+    if (strncasecmp(sql, "OFF", 3) == 0) {
+        *status = 0;
+        return 0;
+    }
+
     return -1;
```
markhannum: I created a set directive ('set cursordebug <level>') which allows an sql session to enable our cursor-debug hooks in the cursor layer.  This allows someone to subsequently disable debug-trace by issuing "set cursordebug off".

bdb/file.c
```diff
@@ -3169,21 +3169,24 @@ if (!is_real_netinfo(bdb_state->repinfo->netinfo))
         }
     }
```
markhannum: This is one of the initial "unable to start a cluster on lxc" issues that Kyle found - on the first day of the first week that he arrived.  The previous code incorrectly assumed that the current node cannot become master inside of this loop.  In a case I observed on my own lxc cluster, a machine managed to become master after the original check (line 3179).  net_send_message fails automatically for self-sent messages, forcing this to sleep and loop 'startup_sync_attempts' times before finally exiting.

bdb/file.c
```diff
@@ -4719,13 +4722,10 @@ static int bdb_downgrade_int(bdb_state_type *bdb_state, int noelect,
         *downgraded = 0;
 
     retries = 0;
```
markhannum: This addresses another "lxc-locked-cluster" error at startup.  A node would win an election under egen x, but get downgraded by a subsequent election before it managed to call upgrade.  The upgrade_allowed flag is set near the end of bdb_open_int.  The retry/poll loop allowed this code (which is holding the bdb-writelock) to prevent the thread which is attempting to call 'upgrade' from getting the bdb writelock.  We can allow it to upgrade even though it is no longer master because none of it's log-writes will be acknowledged by the rest of the cluster, and it's initial 'NEWMASTER' message (called by virtue of the upgrade codepath's __rep_start) will be immediately met with a DUPMASTER from the clusters actual master.

bdb/llmeta.c
```diff
@@ -2655,6 +2655,11 @@ int bdb_add_dummy_llmeta(void)
     }
 
 retry:
```
markhannum: This addresses another locked-startup condition.  The calling code is holding the bdb-lock in read-mode, while a different thread is attempting to upgrade.  The subsequent wait-for-seqnum-from-all code would wait until MIN_REP_TIMEOUTMS as none of the nodes has a valid coherency lease.  An additional check inside of wait-for-seqnum closes the race here.  The lock could be desired AFTER this check- but that's okay, because we catch it inside of the first part of wait-for-seqnum.  Dorin and I might have similar code pr's for the 'inside wait-for-seqnum' case.

bdb/rep.c
```diff
@@ -1724,6 +1724,8 @@ void net_newnode_rtn(netinfo_type *netinfo_ptr, char *hostname, int portnum)
 
 /* Timestamp of when our coherency lease expires on replicant */
 uint64_t coherency_timestamp = 0;
```
markhannum: The 'dump_zero_coherency_timestamp' instrumentation was put in place as an attempt to see and understand why cluster nodes would stay incoherent for semi-extended periods of time after a partition heals.  The changes (mostly bug-fixes) in this PR have reduced the mean time to coherency to about 15 seconds on my laptop computer.  We can do better.  If we choose to work more on this, I suspect election is the next hurdle.

bdb/rep.c
```diff
@@ -3079,6 +3083,13 @@ static int bdb_wait_for_seqnum_from_all_int(bdb_state_type *bdb_state,
             rc = bdb_wait_for_seqnum_from_node_int(bdb_state, seqnum,
                                                    nodelist[i], 1000, __LINE__);
 
+            if (bdb_lock_desired(bdb_state)) {
```
markhannum: Short-circuit getting the base time in wait-for-seqnum if the bdb lock is desired.  This overlaps with a PR from Dorin who solved the same issue in a different way.

berkdb/rep/rep_record.c
```diff
@@ -108,97 +108,119 @@ static inline int wait_for_running_transactions(DB_ENV *dbenv);
     (R) != DB___txn_regop_rowlocks && (R) != DB___txn_regop_gen && (R) != \
     DB___txn_ckp && (R) != DB___dbreg_register)
 
+int gbl_rep_process_msg_print_rc;
```
markhannum: Instrument rep_process_message

berkdb/rep/rep_record.c
```diff
@@ -469,12 +510,22 @@ __rep_process_message(dbenv, control, rec, eidp, ret_lsnp, commit_gen)
 	 * to get in sync.
 	 */
 	if (rp->gen < gen && rp->rectype != REP_ALIVE_REQ &&
-	    rp->rectype != REP_NEWCLIENT && rp->rectype != REP_MASTER_REQ &&
-	    rp->rectype != REP_DUPMASTER) {
+		rp->rectype != REP_NEWCLIENT && rp->rectype != REP_MASTER_REQ) {
```
markhannum: This is a change to the election state machine: a master from a previous generation shouldn't be allowed to downgrade the current master.

berkdb/rep/rep_record.c
```diff
@@ -753,15 +819,25 @@ send:			if (__rep_send_message(dbenv,
...
+		fromline = __LINE__;
 		goto errlock;
 #endif
 	case REP_LOG:
 	case REP_LOG_MORE:
 		CLIENT_ONLY(rep, rp);
 		MASTER_CHECK(dbenv, *eidp, rep);
-		if ((ret =
-			__rep_apply(dbenv, rp, rec, ret_lsnp, commit_gen)) != 0)
+		if (!IN_ELECTION_TALLY(rep)) {
```
markhannum: Prevent a replicant from processing log-messages if it is an election candidate.

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
```diff
@@ -99,6 +99,7 @@
     private int snapshotOffset;
     private boolean isHASql = false;
     private int isRetry;
+    private int nSetsQuerySent;
```
markhannum: Keep track of the number set commands that we have sent to the replicant that we are connected to.

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
```diff
@@ -164,6 +165,7 @@ public Comdb2Handle duplicate() {
     public Comdb2Handle(String dbname, String cluster) {
         super(new ProtobufProtocol(), null);
         sets = new ArrayList<String>();
```
markhannum: JDBC requires statement-level effects, not transaction-level effects.

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
```diff
@@ -1726,7 +1730,7 @@ private boolean reopen(boolean refresh_dbinfo_if_failed) {
          */
 
         // last time we were at dbHostIdx, this time start from (dbHostIdx + 1)
```
markhannum: Fix the issue where the JDBC driver seemingly would not connect.  dbHostIdx was 0, and myDbHosts.size() was 1, which means that prior to this change, both of the next two for loops (which attempt to re-establish a connection with a different replicant) were skipped.

db/appsock_handler.c
```diff
@@ -380,6 +380,8 @@ static void *fstdump_hndlr(void *arg_)
     return NULL;
 }
 
+int gbl_allow_incoherent_sql = 0;
```
markhannum: Provide an lrl tunable that allows us to interact with a database even though it is incoherent.  It must be set in the lrl file to be useful as you cannot set this or any other tunable for an already incoherent database.  In addition to this setting, a user must issue 'set ignorecoherency on' for this to have an effect.

db/sqlinterfaces.c
```diff
@@ -7570,6 +7649,15 @@ static int process_set_commands(struct sqlclntstate *clnt)
                 } else {
                     clnt->verifyretry_off = 1;
                 }
+            } else if (strncasecmp(sqlstr, "queryeffects", 6) == 0) {
```
markhannum: Toggle statement-level or transaction level query effects

tests/linearizable/ctest/nemesis.c
```diff
@@ -0,0 +1,277 @@
+#include <stdlib.h>
```
markhannum: Nemesis functionality from a C program.  I use this utility to study the effect of network partitions (and to measure recovery time) against a running cluster.

linearizable/jepsen/src/comdb2/core.clj
```diff
@@ -1,614 +0,0 @@
-(ns comdb2.core
```
markhannum: Alot of what follows is Kyle correcting my terrible clojure

tests/linearizable/ctest/breakloop.c
```diff
@@ -0,0 +1,308 @@
+#include <alloca.h>
```
markhannum: Run nemesis in a loop.  This utility was created to study a cluster's recovery time after a partition is healed.

berkdb/rep/rep_record.c
markhannum: This closes a race that Kyle pointed out:
"... if you allow log messages to hit candidates, then I think you can wind up in a situation like... n1, the old primary, tries to repllcate write w1. n2, isolated from n1, declares candidacy for leadership, without w1. It asks n3, n4 to vote for it, and neither of them have w1, and they grant their votes to n2. w1 arrives on n3 and n5, and n1 now knows w1 is committed. n2 becomes master, and n3 rolls *back* w1."

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
markhannum: Kyle was managing to coerce the JDBC driver into *failing* to send set directives.  The result was that alot of his failing tests were failing because the database never received 'set transaction serializable'.  It seemed like this was occurring because the code would issue these set commands before reading the results of a previous operation.  When the code eventually *did* read the results, it got a LASTROW, and set nSetsSent to sets.size(), even though the just issued set commands had not ever been sent.

The fix is to introduce a new variable, nSetsQuerySent.  This is set to be the current sets.size() when we issue a sendQuery (and actually SEND the set commands).  When we receive a LASTROW, the code updates nSetsSent to the most recent value of nSetsQuerySent (what we've actually sent to the database) rather than the current sets.size().

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
markhannum: The JDBC driver forces the database into statement-level query effects via the new "set queryeffects statement" directive.  A user can force it back into the old behavior by issuing a "set queryeffects transaction"

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
```diff
@@ -604,6 +606,7 @@ in cdb2jdbc to make executeUpdate() work. */
                 queryList.add(new QueryItem(payload, sql, isRead));
             }
             tdlog(Level.FINEST, "sendQuery returns a good rcode");
```
markhannum: Set the nSetsQuerySet variable after a successful send-query.

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
```diff
@@ -769,6 +772,13 @@ private int runStatementInt(String sql, List<Integer> types) {
         sql = sql.trim();
         String lowerSql = sql.toLowerCase();
```
markhannum: Changed this to mimic what we do in cdb2api.c: every statement we consume everything from the database and clear responses before processing set directives .. this alone might be sufficient to fix cdb2jdbc (I'm testing that now).

berkdb/rep/rep_util.c
```diff
@@ -704,31 +721,6 @@ void
 __env_rep_enter(dbenv)
 	DB_ENV *dbenv;
 {
```
markhannum: This code has been dead for a few years.

bdb/file.c
markhannum: We don't entirely understand this .. may not need the upgrade_allowed flag at all.

markhannum:
Thanks for the help everyone!

riverszhang89:
:clap: :clap::clap: :heart: :heart: :heart:

PR #331, opened 2017-08-13
Fix some memory leaks.

nirbhayc:


nirbhayc:

PR #333, opened 2017-08-14
Sctests timeout

ldennis:
This should fix three failing tests: sc_fieldlarger, sc_swapfields and sc_timepart
sc_fieldlarger and sc_swapfields were failing because concurrent inserts in the tests were running too fast and the schema change wasnt able to catch up.
sc_timepart's failure was due to newly added check_table_version in ddl. Fixed in the second commit of this PR.


ldennis:
Thanks for reviewing.

PR #334, opened 2017-08-14
Fix constraints after schema change

mponomar:


db/osqlcomm.c
```diff
@@ -6436,6 +6436,7 @@ int osql_process_packet(struct ireq *iq, unsigned long long rqid, uuid_t uuid,
             return conv_rc_sql2blkop(iq, step, -1, ERR_NO_SUCH_TABLE, err,
                                      tablename, 0);
         }
+        iq->origdb = iq->usedb;
```
ldennis: Can we instead set `iq->usedbtablevers = iq->usedb->tableversion` in constraints.c? Using `origdb` is a bit confusing to me.

akshatsikarwar: 'usedtableversion' is supposed to be the version the replicant sent. Not the version currently on the master, which 'struct db' will provide.

ldennis: Yes, but I think this is just a "hack" to pass the version test in record.c for cascade deletes/updates. (We do this for usedb too)

ldennis: In the cases of cascade deletes/updates, replicants didnt send any versions. Unless, we make replicants check for constraints.

mponomar: Neither way is particularly clean.  I'm ok with either.  If origdb is confusing, let's set usedtableversion before calling into record.c.  Either way works.

ldennis: I am ok with origdb too, though I slightly prefer usedbtablevers.

ldennis: FYI, I did a similar usedb-version-related fix (/hack) in #333 for the sc_timepart.test.

mponomar:
> What where the other tests in constraints testing?
There's a constraint test there.  The failure we were seeing was due to the table with constraints being recreated several times, and thus having a different version number from the referencing table.  That's why there's a `rebuild` in the new test case.

PR #336, opened 2017-08-15
Discard the pending transaction when receiving RESET from the sockpool.

riverszhang89:
@dorinhogea


akshatsikarwar:
whoah

db/sqlinterfaces.c
```diff
@@ -7229,6 +7229,19 @@ CDB2QUERY *read_newsql_query(struct sqlclntstate *clnt, SBUF2 *sb)
...
         goto retry_read;
     } else if (hdr.type == FSQL_RESET) { /* Reset from sockpool.*/
+
+        if (clnt->ctrl_sqlengine == SQLENG_INTRANS_STATE) {
+            /* Discard the pending transaction when receiving RESET from the
+               sockpool. We reach here if
+               1) the handle is in a open transaction, and
+               2) the last statement is a SELECT, and
+               3) the client closes the handle and donates the connection
+                  to the sockpool, and then,
```
nirbhayc: IIUC, I believe should we reset the connection first, before moving it back to the cache instead of handing/resetting its stale state on reuse.

tests/snapisol.test/t6_01.req
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE newsqlresetbug (i INTEGER)$$
+SET TRANSACTION SNAPSHOT ISOLATION
+BEGIN
+UPDATE newsqlresetbug SET i = i WHERE 1=1
+COMMIT
+BEGIN
+SELECT * FROM newsqlresetbug
```
nirbhayc: Just curious, shouldn't these test req files be kept logically independent? So that they create objects and tear them down at the end.

db/sqlinterfaces.c
riverszhang89: I assume by "cache" you mean the sockpool? And by "we" you mean the client API?

nirbhayc: cache == sockpool, yes. "we" ==> the server code should reset the thread before adding it to the pool.

riverszhang89: Sockpool lives on client side. You might confuse sockpool with the SQL engine thread pool.

nirbhayc: Oh right, I misunderstood.

cdb2api/cdb2api.c
```diff
@@ -1557,7 +1557,7 @@ static int newsql_disconnect(cdb2_hndl_tp *hndl, SBUF2 *sb, int line)
     if ((hndl->firstresponse &&
          (!hndl->lastresponse ||
           (hndl->lastresponse->response_type != RESPONSE_TYPE__LAST_ROW))) ||
-        (!hndl->firstresponse)) {
+        (!hndl->firstresponse) || hndl->in_trans) {
```
mohitkhullar: server side changes should be enough to fix this.

riverszhang89: Mohit, you're right: the API change is not required in order to fix the bug. However, I'd prefer to keep the change to cover all the bases..

riverszhang89:
Reviewed internally.

PR #337, opened 2017-08-15
fix hashing and compare for query_path_component

akshatsikarwar:


db/sqlglue.c
```diff
@@ -591,6 +591,49 @@ void sql_get_query_id(struct sql_thread *thd)
...
+        // mismatch
+        return 1;
+    } else {
+        // both remote
+        n1 = fdb_table_entry_tblname(q1->fdb);
+        n2 = fdb_table_entry_tblname(q2->fdb);
+    }
+    // check name + ix
+    if (strcmp(n1, n2) == 0) {
+        if (q1->ix == q2->ix) {
```
riverszhang89: Combine two if-s into one? Also I suggest you compare ix first then string compare as comparing 2 integers is a lot cheaper than comparing 2 strings.

db/sqlglue.c
```diff
@@ -6227,7 +6268,7 @@ const void *sqlite3BtreeKeyFetch(BtCursor *pCur, u32 *pAmt)
 }
 
 /* add the costs of the sorter to the thd costs */
-void addVbdeToThdCost(int type)
+void addVdbeToThdCost(int type)
```
riverszhang89: Whoah it took me 10 seconds to find out what's changed.

db/sqlglue.c
akshatsikarwar: Will do

db/sqlglue.c
```diff
@@ -611,8 +611,11 @@ static unsigned int query_path_component_hash(const void *key, int len)
 
 static int query_path_component_cmp(const void *key1, const void *key2, int len)
 {
-    const char *n1, *n2;
     const struct query_path_component *q1 = key1, *q2 = key2;
+    if (q1->ix != q2->ix) {
```
riverszhang89: Awesome.. this is even better..

PR #341, opened 2017-08-16
Invalidate cursor pgno before releasing the pagelock

ldennis:
The bug was introduced in my new snapshot code in 2015. The original idea was to keep the last visited page number even though bam_search returns not found. Instead of keeping the page number in cp->pgno, we could use dbc->lastpage for the same purpose. Since we no longer hold page lock on the last page, the cursor itself should not be pointing to the page (i.e. I shouldn't have touched cp->pgno).

This should fix the scindex.test crashes.


berkdb/btree/bt_cursor.c
```diff
@@ -3161,8 +3161,10 @@ split:	ret = stack = 0;
 		 */
 
         /* Invalidate the cursor before releasing the pagelock */
-		cp->pgno = PGNO_INVALID;
-		cp->indx = 0;
+		if (own == 0) {
+			cp->pgno = PGNO_INVALID;
+			cp->indx = 0;
+		}
```
ldennis: This change was discussed with @markhannum and `if (own == 0)` was what berkdb originally had. Also see Mark's last change [here](https://github.com/bloomberg/comdb2-old/pull/723/files#diff-276bb96720dfac5f10f48b22dca4cb37R3159).

riverszhang89:
Can you please enable siasofupdate test in your branch?

markhannum:
Looks great- good to merge .. tx Lingzhi!

ldennis:
Ran siasofupdate ~300 times locally without errors.

PR #344, opened 2017-08-18
Using thread_safe readdir_r() on all platforms.

riverszhang89:
We memcpy the directory entry returned from `readdir()` to our own buffer. However it is not strictly thread-safe, for instance: http://comdb2.s3-website-us-east-1.amazonaws.com/tests/3610e435345d711ad2a59daf931401634b424915/detail.txt


riverszhang89:
Thank you.

PR #349, opened 2017-08-27
Change to use standardised intrinsics for vector operations

mthorpe7:
Replace gcc intrinsics with standardised intrinsics.

This works towards allowing clang builds instead of gcc.

### DIFF
--- a/crc32c/crc32c.c
+++ b/crc32c/crc32c.c
@@ -1,16 +1,16 @@
 /*
    Copyright 2015 Bloomberg Finance L.P.
-  
+
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


crc32c/crc32c.c
```diff
@@ -77,6 +77,10 @@ uint32_t crc32c_software(const uint8_t* buf, uint32_t sz, uint32_t crc)
 
 #ifdef __x86_64__
 
+#include <emmintrin.h>
```
riverszhang89: Thank you for the patch. `smmintrin.h` is a super set of `emmintrin.h` hence you don't need to `#include <emmintrin.h`.

mthorpe7: Ah, thanks. Corrected.

PR #350, opened 2017-08-27
Remove dead code from bb_oscompat

mthorpe7:
None of this code is currently in use.


mponomar:
Looks good.  Thanks Michael.

PR #356, opened 2017-08-30
SQL: Support for START TRANSACTION

nirbhayc:


cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
```diff
@@ -758,6 +758,14 @@ private boolean isClientOnlySetCommand(String sql) {
         return false;
     }
 
+    private boolean isBeginQuery(String sql) {
+        if (sql.equalsIgnoreCase("begin") ||
+            sql.ReplaceAll("\\s+"," ").equalsIgnoreCase("start transaction")) {
```
riverszhang89: This is wrong. It does not capture `START TRANSACTION AS OF`.

akshatsikarwar:
Purpose?

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
nirbhayc: I missed that, pushed a fix. I wonder how ```if (lowerSql.equals("begin"))``` worked in case of AS OF? Probably, the scenario was never tested with jdbc driver?

nirbhayc:
> Purpose?

SQL standard

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
riverszhang89: Yes. That is a bug too. Nice catch.

riverszhang89:
Which version of SQL standard are you referring to?

riverszhang89:
Okay. It is part of SQL-99.

adizaimi:
Can we please move the test into a new (or another existing sql test), so that basic remains a fast test to run?
If other DBs support START TRANSACTION (and part of the standard), there seems no harm to add it to comdb2. This PR looks good otherwise.

PR #357, opened 2017-08-30
Remove tautological comparison

mthorpe7:
```
./db/types.c:9579:49: error: comparison of constant 128 with expression of type 'char' is always true [-Werror,-Wtautological-constant-out-of-range-compare]
        if ((highnibble & 0x080) && (highnibble != 0x080)) {
                                     ~~~~~~~~~~ ^  ~~~~~
```

### DIFF
--- a/db/types.c
+++ b/db/types.c
@@ -9576,7 +9576,7 @@ int dec64_exponent_is_outrageous(server_decimal64_t *pdec64, char *decimals)
         }
         /* check range for positive exponents; first nibble is 8 for new format!
          */
-        if ((highnibble & 0x080) && (highnibble != 0x080)) {
+        if (highnibble & 0x080) {
             return 0;
         }
         return 1;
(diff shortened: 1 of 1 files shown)


mthorpe7:
Note: I entered this as a single pull request because I'm not sure of the intention of the code. the alternative fix is `highnibble` should be declared as an `unsigned char` instead

ldennis:
I think we should instead declare the variable as `unsigned char` to preserve the logic. @dorinhogea can you confirm this?

mthorpe7:
@ldennis @dorinhogea - corrected.

dorinhogea:
The update is correct, the condition is correct (bit 080 set, but other bits set as well).

dorinhogea:
Thank you for fixing this.  
NOTE: this piece of code refers to old versions of R6, and it is only meaningful for rolling out R7 internally.  We will remove the function after the internal rollout.

nirbhayc:
https://github.com/bloomberg/comdb2/commit/c6c2a75b1e071b7ef851f278bd1bf1ac457b81b3

PR #366, opened 2017-09-06
cherrypick sqlite [d1ba2002]

dorinhogea:
Present already in latest sqlite and R6.


dorinhogea:
thank you

adizaimi:
Does this result in less vdbe steps? If yes, would it make sense to add a silly testcase that runs this and makes sure that the vdbe steps we take are less than X?

PR #370, opened 2017-09-07
Rootpages

dorinhogea:
already in R6


dorinhogea:
remsql format changed (since a new range of rootpage values are assigned to remote)

PR #372, opened 2017-09-08
Fix negative ds intervals with non-0 ms

mponomar:


riverszhang89:
Awesome- good to merge. Thank you Mike!

PR #376, opened 2017-09-12
sc_live races

ldennis:
Two commits:
- Added rwlock to prevent races between live_sc_post*() and live_sc_off()
- get_hostname_with_crc32 could return NULL if db was moved to another machine and we should check for NULL

### DIFF
--- a/schemachange/schemachange.c
+++ b/schemachange/schemachange.c
@@ -182,7 +182,7 @@ int start_schema_change_tran(struct ireq *iq, tran_type *trans)
         logmsg(
             LOGMSG_INFO,
             "Resuming schema change: fetched seed 0x%llx, original node %s\n",
-            seed, node);
+            seed, node ? node : "(unknown)");
         if (stopsc) {
             errstat_set_strf(&iq->errstat, "Master node downgrading - new "
                                            "master will resume schemachange");
--- a/schemachange/schemachange.c
+++ b/schemachange/schemachange.c
@@ -534,9 +534,10 @@ int live_sc_post_delete(struct ireq *iq, void *trans, unsigned long long genid,
                         const void *old_dta, unsigned long long del_keys,
                         blob_buffer_t *oldblobs)
 {
-
+    int rc = 0;
+    pthread_rwlock_rdlock(&sc_live_rwlock);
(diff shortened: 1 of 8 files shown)


akshatsikarwar:
Why do we need broadcast_sc_start?

schemachange/schemachange.c
```diff
@@ -534,76 +534,89 @@ int live_sc_post_delete(struct ireq *iq, void *trans, unsigned long long genid,
                         const void *old_dta, unsigned long long del_keys,
                         blob_buffer_t *oldblobs)
 {
-
+    int rc = 0;
+    pthread_rwlock_rdlock(&sc_live_rwlock);
```
akshatsikarwar: I prefer to wrap in function so we don't have to goto unlock. Following is more robust and easier to read:
```
int live_sc_post_delete() {
    lock()
    live_sc_post_delete_int() // this function is now _int
    unlock()
}
```

ldennis:
`broadcast_sc_start` is probably for stat displaying? R6 has it so that `send stat` will show sc in progress on all nodes. For `do_ddl`, we broadcast but dont care about the rcode and will continue even broadcast fails.

schemachange/schemachange.c
ldennis: sure, thanks

akshatsikarwar:
Can we remove broadcast? `send stat` should read `scseed` from `llmeta` to show sc in progress.

ldennis:
I am ok with removing it. @adizaimi , I think you added the broadcast, is there any other usage for the function? Can you confirm if we can remove it?

adizaimi:
At some point it was setting a global to avoid analyze from running. Also it would stop any other SC from coming in. Things may be changed since. I'll look.

ldennis:
I am going to leave `broadcast_sc_start` for another PR.

adizaimi:
Yes, no need to add to this.

dorinhogea:
Certain sc types require all nodes up, due to limited recovery ability.  broadcast_sc_start  tried to do that, albeit poorly.

riverszhang89:
Thanks for fixing the bugs Lingzhi!

PR #378, opened 2017-09-12
Eliminate conflicts in maccparse.y from keystruct empty rhs rule

adizaimi:


tests/basic.test/runit
```diff
@@ -193,7 +193,7 @@ delete_records()
             continue
         fi
         let j=j+1
-        sleep 0.01
+        sleep 0.001
```
ldennis: Why did we need sleep here?

adizaimi: I dont remember. I can try removing.

PR #385, opened 2017-09-14
Fix bb build

markhannum:
Fixes to allow building on BB Solaris, AIX and linux machines


dorinhogea:
what is the extra commits?

markhannum:
Hmmm .. lemme try this again ..

PR #386, opened 2017-09-15
fixes for time partition, schema changes, and so

dorinhogea:
Resubmit after rebase.

### DIFF
--- a/db/views.h
+++ b/db/views.h
@@ -50,6 +50,16 @@ enum views_trigger_op {
     VIEWS_TRIGGER_UPDATE = 3
 };
 
+typedef struct timepart_sc_arg 
+{
+    struct schema_change_type *s;
+    const char *view_name;
+    int indx;
+    int nshards;
+    int rc;
+    void *tran; /*remove?*/
+} timepart_sc_arg_t; 
+
 /**
  * Initialize the views
  *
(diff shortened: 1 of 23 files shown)


db/views.h
```diff
@@ -261,12 +270,27 @@ int timepart_is_shard(const char *name, int lock);
 int timepart_is_timepart(const char *name, int lock);
 
 /**
- * Alter a timepart 
+ * Time partition schema change resume
+ *
+ */
+int timepart_resume_schemachange(int check_llmeta(const char *));
```
ldennis: I don't see this function being called. Why do we need this?

db/views.h
```diff
@@ -261,12 +270,27 @@ int timepart_is_shard(const char *name, int lock);
...
+
+/**
+ * During resume, we need to check at if the interrupted alter made any
+ * progress, and continue with that shard
+ *
+ */
+int timepart_schemachange_get_shard_in_progress(const char *view_name,
```
ldennis: same as above. Why do we need this? I think `resume_schema_change` already loops through all tables in the `thedb->num_dbs` array.

ldennis:
Does this also resolve timepart sc resume issue?
And does this fix races you found for views on replicants?

dorinhogea:
No, the resume will need review and another PR.

db/views.h
dorinhogea: I need these for partition resume, it is not done.

dorinhogea:
This PT is just meant to refactor time partition schema changes (i.e. eliminate lower level handling) , and align the code with the new refactored tran ddl changes.

db/views.h
ldennis: I think resume_schema_change already takes care of resuming multiple schema changes?

ldennis:
The refactored code looks good.

dorinhogea:
thank you

Issue #330, opened 2017-08-12
table-schema lock is acquired per row inserted/deleted/updated


akshatsikarwar:
Insert 1000 rows, get tablelock 1000 times.
Better:
As bplog is streaming in, add table to list (dedup'd) and get all required table-schema locks once.

PR #387, opened 2017-09-15
Improve schema and tbl-name locking

akshatsikarwar:


akshatsikarwar:
Idea is to unify code to prepare+lock tablename between offload, sql, lua.

db/sql.h
```diff
@@ -813,13 +808,35 @@ int release_locks_on_emit_row(struct sqlthdstate *thd,
 void clearClientSideRow(struct sqlclntstate *clnt);
 void comdb2_set_tmptbl_lk(pthread_mutex_t *);
 void clone_temp_table(sqlite3 *dest, const sqlite3 *src, const char *sql,
-                      int rootpg); //, pthread_mutex_t *lk);
-void sqlengine_prepare_engine(struct sqlthdstate *, struct sqlclntstate *);
-int check_thd_gen(struct sqlthdstate *, struct sqlclntstate *);
+                      int rootpg);
+int sqlengine_prepare_engine_int(struct sqlthdstate *, struct sqlclntstate *,
+                                 int recreate);
```
dorinhogea: Any chance we rename this to something not _int if exposed through header? If there is a good reason to share this across modules, it should have a better prototype :)

schemachange/sc_logic.c
```diff
@@ -718,16 +714,7 @@ int resume_schema_change(void)
...
-            /*
-            **   _
-            **  | |_ ___ _ __ ___  _ __
-            **  | __/ _ \ '_ ` _ \| '_ \
-            **  | ||  __/ | | | | | |_) |
-            **   \__\___|_| |_| |_| .__/
-            **                    |_|
-            */
             if (scabort) {
-                system("figlet force scabort");
```
dorinhogea: thanks

db/sqlinterfaces.c
```diff
@@ -1436,71 +1436,61 @@ void sql_dump_hist_statements(void)
...
+            reqlog_set_rqid(logger, clnt->osql.uuid, sizeof(uuid_t));
+    }
+
+    LISTC_T(struct sql_hist) lst;
     listc_init(&lst, offsetof(struct sql_hist, lnk));
 
-    h = calloc(1, sizeof(struct sql_hist));
-    if (thd->sqlclntstate && thd->sqlclntstate->sql)
-        h->sql = strdup(thd->sqlclntstate->sql);
+    struct sql_hist *h = calloc(1, sizeof(struct sql_hist));
```
dorinhogea: Any time  you feel the need to cluster variables and associated code together, there is a chance to put everything in its own encapsulating function, just saying

db/sqlinterfaces.c
```diff
@@ -3786,18 +3753,12 @@ static void clear_stmt_record(struct sql_state *rec)
...
+                       struct sql_state *rec, int outrc)
 {
     sqlite3_stmt *stmt = rec->stmt;
-    if (stmt || ((stored_proc == 1) && (rec->status & CACHE_HAS_HINT))) {
-        if ((gbl_enable_sql_stmt_caching == STMT_CACHE_ALL) ||
-            ((gbl_enable_sql_stmt_caching == STMT_CACHE_PARAM) && clnt->tag)) {
-            if (stmt)
-                sqlite3_reset(stmt);
-
+    if (stmt) {
```
dorinhogea: do we cache the requests that have no parameters if only STMT_CACHE_PARAM is set with this chance? do we want to do that?

db/sqlinterfaces.c
```diff
@@ -5712,80 +5699,67 @@ static void sqlengine_work_lua_thread(struct thdpool *pool, void *work,
...
+    } else {
         abort();
+    }
 
     thr_set_user(clnt->appsock_id);
 
     clnt->osql.timings.query_dispatched = osql_log_time();
     clnt->deque_timeus = time_epochus();
 
-    /* make sure we have an sqlite engine sqldb */
```
dorinhogea: thanks!

db/sqlinterfaces.c
```diff
@@ -6229,6 +6150,17 @@ static void sqlengine_thd_end(struct thdpool *pool, void *thddata)
     backend_thread_event(thedb, COMDB2_THR_EVENT_DONE_RDWR);
 }
 
+static void thdpool_sqlengine_start(struct thdpool *pool, void *thd)
```
dorinhogea: Why aliases?

db/sqlinterfaces.c
```diff
@@ -8530,181 +8462,34 @@ static int execute_sql_query_offload_inner_loop(struct sqlclntstate *clnt,
...
+    }
+    sqlite3 *sqldb = poolthd->sqldb;
+    sqlite3_stmt *stmt = rec.stmt;
+    thrman_wheref(poolthd->thr_self, "%s", rec.sql);
     user_request_begin(REQUEST_TYPE_QTRAP, FLAG_REQUEST_TRACK_EVERYTHING);
-
     if (gbl_dump_sql_dispatched)
         logmsg(LOGMSG_USER, "BLOCKSQL mode=%d [%s]\n", clnt->dbtran.mode,
                 clnt->sql);
-
```
dorinhogea: thank you following up on this, I have not get to it after execute_sql_query refactoring!

lua/sp.c
```diff
@@ -962,59 +962,23 @@ static int send_col_data(Lua lua, SP sp, sqlite3_stmt *stmt, int nargs)
...
-
-#define no_active_stmt(...)
-#define no_active_stmt_but_me(...)
-
 static void donate_stmt(SP sp, dbstmt_t *dbstmt)
 {
     sqlite3_stmt *stmt = dbstmt->stmt;
     if (stmt == NULL) return;
 
-    if (!gbl_enable_sql_stmt_caching || !dbstmt->has_cmpl_stmt) {
```
dorinhogea: nice

lua/sp.c
```diff
@@ -1076,22 +1032,41 @@ static const char *create_temp_table(Lua lua, pthread_mutex_t *lk)
...
-    char *err = NULL;
-    comdb2_set_tmptbl_lk(lk);
-    sqlite3 *db = getdb(sp);
-    db->force_sqlite_impl = 1;
-    rc = sqlite3_exec(db, strbuf_buf(sql), NULL, NULL, &err);
-    db->force_sqlite_impl = 0;
+    // Following can throw exception which may leak strbuf.
+    // Copy DDL string onto stack instead.
+    int len = strbuf_len(sql) + 1;
+    char *ddl = alloca(len);
```
dorinhogea: What is the maximum sp length?

db/osqlcomm.c
```diff
@@ -6416,22 +6416,23 @@ int osql_process_packet(struct ireq *iq, unsigned long long rqid, uuid_t uuid,
         char *tablename;
 
         tablename = (char *)osqlcomm_usedb_type_get(&dt, p_buf, p_buf_end);
+        bdb_lock_tablename_read(thedb->bdb_env, tablename, trans);
```
dorinhogea: If we get the table lock here, do we get it again later in record.c?

dorinhogea: (i.e., it is great to do it  here, but I would prefer if this would be the only place:) )

db/sqlinterfaces.c
```diff
@@ -5712,80 +5699,67 @@ static void sqlengine_work_lua_thread(struct thdpool *pool, void *work,
...
+    if (rc == SQLITE_DONE) {
+        return 0;
+    }
+    return rc;
+}
+
+static int execute_verify_indexes(struct sqlthdstate *thd,
+                                  struct sqlclntstate *clnt)
+{
+    rdlock_schema_lk();
```
ldennis: For `verify_indexes` queries (i.e. partial/functional indexes' internal queries), there is no need to get schemalk or to re-prepare engine. We can just simply run them as these queries can only be common table expressions. see [here](https://github.com/bloomberg/comdb2/pull/387/files#diff-4932981bbf2c2bd6bf0267d1a6d39173L5741)

ldennis: Maybe I should take this whole logic out of sqlinterfaces.c?

akshatsikarwar: Perhaps that is a better way to do it.

db/osqlcomm.c
akshatsikarwar: We get tablelock for each row - This is the correct place to get it. `record.c` is too late. Cleanup will be a different PR. Already have issue #330 to track. Will have to reconcile with how we support tag requests.

This is bare minimum required for correctness at the moment.

db/sql.h
akshatsikarwar: Agree.

db/sqlinterfaces.c
akshatsikarwar: I embrace C99 whole heartedly :-p

db/sqlinterfaces.c
akshatsikarwar: Yes this was a fun one, as I am sure it was for you before!

akshatsikarwar:
Thanks guys for looking. I know this is large changeset in one PR. It was hard to break into individual commits which passed tests so I got lazy and lumped them all together.

db/sqlinterfaces.c
akshatsikarwar: Good catch.

I don't like the nesting in `put_prepared_stmt()`. I will simplify (so its atleast easy to read)

akshatsikarwar:
I've rebased and fixed conflicts with views - plz review `prepare_engine`.
I've refactored `put_prepared_stmt` to remove 7(!!) levels of nested if stmts.

akshatsikarwar:
Oh wow!

akshatsikarwar:
run robo, run!

PR #390, opened 2017-09-16
illustrate rootpage bug, fixed in rootpages branch

dorinhogea:
testcase; fixed is in https://github.com/bloomberg/comdb2/pull/370


adizaimi:
Test looks ok, not sure what the failure of sc_remsql is. BTW, just wondering if pre #370 builds fail the test.

PR #393, opened 2017-09-17
Add printf attribute to logmsg, and fix errors

mthorpe7:


adizaimi:
Code looks ok to me (needs formatting apparently). Not sure what the Smoke testing error is.

mthorpe7:
@adizaimi - not sure what the smoke test error is either. maybe @riverszhang89 can help?

just pushed the clang format fixes now.

riverszhang89:
```
berkdb/btree/bt_prefix.c:113:23: error: zero-length gnu_printf format string [-Werror=format-zero-length]
   logmsg(LOGMSG_USER, "");
```

riverszhang89:
I have no idea why the error was only reported in your branch.. :(

nirbhayc:
@mthorpe7 Smoke test uses docker to build and install comdb2 and jdbc driver on rpm/deb. roborivers does not print much about the failure at the moment, but you should be able to check it by running it in your local environment (check out the Makefile under tests/smoketest).

mthorpe7:
@riverszhang89 - because I changed logmsg to check printf style arguments. The real question is why didn't the regression testing pick this up :)

riverszhang89:
@mthorpe7 Smoke test runs on an EC2 instance. Regression runs on a dedicated Bloomberg machine. My guess is that the GCC on the EC2 instance is stricter than ours.

adizaimi:
Trying to get this merged--submitted for testing.

adizaimi:
One test timed out -- client disconnect -- will try again.

adizaimi:
This is looking good now, and tests passed.

PR #394, opened 2017-09-18
Handle wrong db by invalidating current hosts/ports and doing requery on comdb2db

mohitkhullar:
...


mohitkhullar:
reviewed in bbgithub

PR #395, opened 2017-09-18
Fix a memory leak in client context handling.

riverszhang89:


adizaimi:
Looks good.

riverszhang89:
Thank you very much Adi.

Issue #396, opened 2017-09-18
Logging remote sql components access freed memory


dorinhogea:
The "struct query_path_component" contains a pointer to a fdb object describing a remote btree, which will be freed if a schema change require refreshing the locally cached schema.

dorinhogea:
Resolved in https://github.com/bloomberg/comdb2/pull/397

PR #397, opened 2017-09-18
use direct fields instead of object pointer for fdb entry

dorinhogea:
addresses https://github.com/bloomberg/comdb2/issues/396


dorinhogea:
Thank you Akshat

PR #398, opened 2017-09-18
Improve make setup operations for tests

adizaimi:
* Dump all exports, runit, clean cmds in one script
* Cleanup svninfo git path, compile warning, use echo instead of print


adizaimi:
FTR sc_remsql is failing in master too.

adizaimi:
Thanks Akshat

dorinhogea:
thank you

PR #400, opened 2017-09-19
reqlog detailed blob bound prameter converted to hex

adizaimi:
* convert to hex blob content so we don't crash
* move tohex function to util file


db/db_util.c
```diff
@@ -632,3 +632,33 @@ char *get_full_filename(char *path, int pathlen, enum dirtype type, char *name,
...
+    mem = malloc((2 * keylen) + 2);
+    output = util_tohex(mem, key, keylen);
+
+    *buf = output;
+}
+
+/* Return a hex string */
+char *util_tohex(char *output, char *key, int keylen)
+{
+    output[0] = '\0';
```
mponomar: output[keylen] = 0?

db/eventlog.c
```diff
@@ -109,9 +110,8 @@ static cson_output_opt opt = {.indentation = 0,
 void eventlog_params(struct reqlogger *logger, sqlite3_stmt *stmt,
                      struct schema *params, struct sqlclntstate *clnt)
 {
-    if (eventlog == NULL) return;
-    if (!eventlog_enabled) return;
-    if (!eventlog_detailed) return;
+    if (eventlog == NULL || !eventlog_enabled || !eventlog_detailed)
+        return;
```
mponomar: Looks like we don't handle NULL?

db/eventlog.c
```diff
@@ -232,9 +232,19 @@ void eventlog_params(struct reqlogger *logger, sqlite3_stmt *stmt,
                 if (rc == 0)
                     blobno++;
             }
-            if (rc == 0)
+            if (rc == 0) {
+                datalen = min(datalen, 1024); /* cap the datalen logged */
```
mponomar: attr for later?

adizaimi:
The attrs, noleak, and handling null will come later.

PR #402, opened 2017-09-19
Significantly improving SQLite sort performance on large datasets with BLOB or TEXT: Genid-based Sort.

riverszhang89:
When sorting rows, SQLite writes the complete row content into the sorter. After all rows are written into the sorter, SQlite simply walks the sorter to get the sorted rows. It is the most efficient implementation on small datasets.

However the implementation may suffer from high memory usage and poor performance on large datasets, especially those with blob and text. Take the following schema and query as an example:
```SQL
CREATE TABLE T1 (i INTEGER, b BLOB);

SELECT i, b FROM T1 ORDER BY i;
```
Suppose the table contains 1M rows and each row has a 1MB blob. The SQLite implementation requires more than 1TB RAM to run the select query completely in memory, which makes it almost impossible to accomplish. We then have to use external sorting with high disk I/O cost.

Moreover, the SQLite implementation will unnecessarily fetch all the blobs for LIMIT queries where only a handful are actually needed, for instance:
```SQL
SELECT i, b FROM T1 ORDER BY i LIMIT 10;
```

The genid-base sort, when applicable, only writes the genid-s of the blobs to the sorter. Those blobs are fetched after all the partial rows are sorted and only when they are needed. It uses far less RAM and is much faster than the SQLite implementation for ORDER-BY and GROUP-BY queries with blobs.

### DIFF
--- a/sqlite/sqliteInt.h
+++ b/sqlite/sqliteInt.h
@@ -3137,6 +3137,9 @@ struct AuthContext {
 #ifdef SQLITE_ENABLE_PREUPDATE_HOOK
 #define OPFLAG_ISNOOP        0x40    /* OP_Delete does pre-update-hook only */
 #endif
+/* vvv COMDB2 MODIFICATION vvv */
+#define OPFLAG_GENID         0x20    /* OP_Column only fetches genids */
+/* ^^^ COMDB2 MODIFICATION ^^^ */
 #define OPFLAG_LENGTHARG     0x40    /* OP_Column only used for length() */
 #define OPFLAG_TYPEOFARG     0x80    /* OP_Column only used for typeof() */
 #define OPFLAG_BULKCSR       0x01    /* OP_Open** used to open bulk cursor */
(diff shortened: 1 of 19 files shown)


sqlite/sqliteInt.h
```diff
@@ -3137,6 +3137,9 @@ struct AuthContext {
 #ifdef SQLITE_ENABLE_PREUPDATE_HOOK
 #define OPFLAG_ISNOOP        0x40    /* OP_Delete does pre-update-hook only */
 #endif
+/* vvv COMDB2 MODIFICATION vvv */
+#define OPFLAG_GENID         0x20    /* OP_Column only fetches genids */
+/* ^^^ COMDB2 MODIFICATION ^^^ */
```
ldennis: Nice comment 👍 . Can you put this kind of comments before/after other SQLite changes? It would make future SQLite upgrade easier.

akshatsikarwar:
Seems like a solid idea. Upstream didn't want to do this for us? Better for future engine upgrades.

ldennis:
If I do
```SQL
SELECT i, b, length(b) FROM T1 ORDER BY i;
```
how many times do we fetch blob `b` with genid sort? Can genid sort be used in this case?

sqlite/sqliteInt.h
riverszhang89: Thanks. Will do.

riverszhang89:
@akshatsikarwar I didn't talk to Dr. Hipp as the problem is a bit comdb2-specific: We serve large datasets, and we store blobs separately. Also I wasn't quite sure about the idea so initially I just wanted to work out a proof of concept. After spending some time on it, I thought it'd be nice to take it 1 one step farther to make it a real project.

riverszhang89:
@ldennis Yeah. Thanks for finding the bug.

akshatsikarwar:
I think we should try to get upstream to implement this.

akshatsikarwar:
What happens if engine latches genid, but by the time fetches blob, row has been updated/deleted?

riverszhang89:
@akshatsikarwar  :( Sadly it will be broken.

riverszhang89:
@akshatsikarwar You are right it is the best to have Dr. Hipp do this for us.

akshatsikarwar:
As always this will need more work even after upstream implements it. Rows don't go missing in SQLite's concurrency model.

PR #404, opened 2017-09-21
reqlog print null bound parameter

adizaimi:
* print null bound parameter
* avoid leak
* reduce tohex definitions


adizaimi:
Failure is from setup:

PR #410, opened 2017-09-25
fix bad rules in maccparse.y

adizaimi:


adizaimi:
Dorin updated in chat:
11:49:55 DORIN HOGEA : csc2 LGTM, though it's been years since yacc&bison

nirbhayc:
@adizaimi This change is potentially backwards incompatible. From the docs:
```
A key can point at more than one key by specifying multiple "REFERENCED_TABLE_NAME":"REFERENCED_KEY_NAME" items on the same line delimited by white-space. A key can point to a key within its own table by specifying its own table name in REFERENCED_TABLE_NAME.
```
The following commands would have exec'ed successfully without this patch:
```
test> create table t1(i int primary key) $$
[create table t1(i int primary key)] rc 0
test> create table t2(i int primary key) $$
[create table t2(i int primary key)] rc 0
test> create table t3 { schema
 ...> {
 ...>     int i
 ...> }
 ...> keys
 ...> {
 ...>     "COMDB2_PK" = i
 ...> }
 ...> constraints
 ...> {
 ...>     "COMDB2_PK" -> <"t1":"COMDB2_PK"> <"t2":"COMDB2_PK"> on update cascade on delete cascade
 ...> }
 ...> } $$
[create table t3 { schema
{
    int i
}
keys
{
    "COMDB2_PK" = i
}
constraints
{
    "COMDB2_PK" -> <"t1":"COMDB2_PK"> <"t2":"COMDB2_PK"> on update cascade on delete cascade
}
}] failed with rc 304 ERROR at line  10: syntax error: <
```
Consider updating relevent docs iff you think none of the existing installations use this (obscure?) construct, I'd suggest reverting it otherwise.

adizaimi:
Nirbhay, nice catch. It seems that the picture of the grammar for constraints in https://bloomberg.github.io/comdb2/table_schema.html, is not the same as the wording few lines later. I don't know at this point if we should or should not support multiple foreign keys--obviously if we should support, this patch will be reverted. Thanks.

riverszhang89:
We have this usage on production.

riverszhang89:
I know this because I fixed a related bug a while ago. See https://github.com/bloomberg/comdb2-old/commit/453182026a757016459a0df7f2a14598cf697da2

adizaimi:
Note that the way that it was allows for incorrect constraint:
constraints
{
    "d" -> <"t1":"a"> on delete cascade
    "f" ->
}

PR #414, opened 2017-09-27
Add gradle build system to JDBC driver

aibolit:
I've introduced gradle build scripts for the jdbc
To install dependency `./gradlew install` 
Having the protoc executable is also no longer required for this to build (the correct version will be downloaded; I've had issues with these mismatching in the maven build)

Tested on WSL (Ubuntu) and Windows (cmd)

> **The non-generated files are** 
> - cdb2jdbc/build.gradle
> - cdb2jdbc/settings.gradle
> - .gitignore


riverszhang89:
This is awesome. Only Java is required to build cdb2jdbc. All other dependencies will be automatically resolved by gradlew (on *nix) or gradlew.bat (on Windows).

riverszhang89:
@aibolit Thank you for the patch!

PR #419, opened 2017-09-28
Use plain malloc for log region.

mponomar:


berkdb/os/os_region.c
```diff
@@ -116,7 +116,10 @@ __os_r_attach(dbenv, infop, rp)
 	dbenv->set_use_sys_malloc(dbenv, 1);
 
 	if (!gbl_largepages || rp->size < MB_2) {
-		ret = __os_calloc(dbenv, 1, rp->size, &infop->addr);
+        if (rp->size != 0)
+            ret = __os_calloc(dbenv, 1, rp->size, &infop->addr);
+        else
+            ret = 0;
```
markhannum: Do you think it's worth setting infop->addr to NULL here?

markhannum:
Ok to merge

PR #421, opened 2017-09-28
allow multiple table entries for foreign key constraints

adizaimi:
* this fixes issue introduced in https://github.com/bloomberg/comdb2/pull/410: this will now allow  multiple foreign keys per constraint if foreign keys are enclosed in <> : 
```
constraints
{
    "COMDB2_PK" -> <"t1":"COMDB2_PK"> <"t2":"COMDB2_PK"> on update cascade on delete cascade
}
```
* add testcase with multiple table:index entries as foreign key constraint
* updated docs


akshatsikarwar:
oh man - these gif don't animate

docs/src/sqlitegen/bubble-generator-data.tcl
```diff
@@ -491,13 +491,21 @@ set all_graphs {
...
-              {opt on delete cascade}
+          {line /keyname -> 
+               {or 
+                    {line /ref-table-name : /ref-keyname }
+                    {line {loop {line < /ref-table-name : /ref-keyname > } } }
+               }
+          }
+          {opt 
+            {loop 
+               {line on {or update delete} {or {line cascade} {line restrict} }}
```
nirbhayc: Please use upper-case letters for keywords.

tests/sc_lotsoftables.test/runit
```diff
@@ -159,6 +159,29 @@ for i in `seq 1 $tblcnt`; do
...
+  cdb2sql ${CDB2_OPTIONS} $dbnm --host $node "exec procedure sys.cmd.send('sqlenginepool stat')" | grep Current
+  sleep 7 #wait for sql threads to go away
+  cdb2sql ${CDB2_OPTIONS} $dbnm --host $node "exec procedure sys.cmd.send('sqlenginepool stat')" | grep Current
+done
+#get memory utilization
+ps -p $(cat ${TMPDIR}/${DBNAME}.pid) -o vsz=
+
+
+
+
```
nirbhayc: This seems unrelated. Please move this to a separate PR/commit.

docs/src/sqlitegen/bubble-generator-data.tcl
adizaimi: these keywords seem to be case sensitive, not sure if we want to confuse by putting in upper case.

PR #424, opened 2017-10-02
Fix queuedb deadlock handling.

riverszhang89:
This looks like a typo: When c_close returns DB_LOCK_DEADLOCK, we should return the error immediately to let the caller abort the transaction.

This was internally reviewed by Mike, but I'd like roborivers to eyeball as well.


akshatsikarwar:
Good

riverszhang89:
Thank you Akshat.

PR #428, opened 2017-10-03
Add eventlog_status and other cleanup

adizaimi:


db/eventlog.c
```diff
@@ -72,11 +73,19 @@ void eventlog_init(const char *dbname)
...
+}
+
 static gzFile eventlog_open()
 {
     char *fname = eventlog_fname(thedb->envname);
+    gbl_eventlog_fname = fname;
     gzFile f = gzopen(fname, "2w");
-    free(fname);
     if (f == NULL) {
         eventlog_enabled = 0;
```
mohitkhullar: free_gbl_eventlog_fname() will never be called, as eventlog will be null in this case.

adizaimi: Good catch. Not sure under what condition (and how often) could gzopen() fail, but fixing so when it does, we free.
Thanks Mohit.

PR #431, opened 2017-10-05
rootpgleaktest: instead of let, use posix math to compute mem diff

adizaimi:
let a=0  causes script to exit as it thinks failure (set -e), so when diff is zero test would fail.
Using posix math fixes this issue:
13:05:30> (max(rootpage)=103)
13:05:30> numberbefore 17027072
13:05:45> numberafter 17108992
13:05:45> numberdelta 81920
13:05:45> Passed.


riverszhang89:
TIL from [the man page](https://linux.die.net/man/1/bash):

> let arg [arg ...]
> Each arg is an arithmetic expression to be evaluated (see ARITHMETIC EVALUATION above).  If the last arg evaluates to 0, let returns 1; 0 is returned otherwise.

Thank you Adi!

PR #433, opened 2017-10-05
 make generation of cnonce pthread safe

adizaimi:
We use a random number as part of cnonce, which for multithreaded functions could result in same cnonce from different threads. This patch ensures thread safety for generating cnonce.


adizaimi:
Tried 1000 iterations of creating 10000 threads and using val_combine() with same tv value, and the 48bit portion of the hash did not create a dup. Also tried over 160k iterations of 100 threads -- no dups were observed.

dorinhogea:
looks good

adizaimi:
Thank you Dorin.

PR #440, opened 2017-10-09
Have pmux close "reg" connections when database goes away

mponomar:
Fixes file descriptor leak when a database exits.  pmux will reject "reg" connections unless database makes a "hello" connection first.  Both are closed when database exits.


db/db_tunables.h
```diff
@@ -1285,4 +1285,13 @@ REGISTER_TUNABLE("poll_in_pgfree_recover", "Poll pgfree recovery handler.",
                  TUNABLE_BOOLEAN, &gbl_poll_in_pg_free_recover,
                  EXPERIMENTAL | INTERNAL, NULL, NULL, NULL, NULL);
 
+REGISTER_TUNABLE("portmux_port", "Specifies portmux port", TUNABLE_INTEGER,
```
adizaimi: there already is a portmux_port (remove?):
https://github.com/mponomar/comdb2/blob/b0560811f6811fa9e9e10fcec3cd5cdff3834818/db/db_tunables.h#L937
There also is a portmux_bind_path which may be similar to the "portmux_sock" you added next.

adizaimi: we use the two i mentioned above in the testsuite (see setup), if the user defines a PMUXPORT.

tools/pmux/pmux.cpp
```diff
@@ -171,6 +180,14 @@ int client_func(int fd)
     char *sav;
     char *cmd = strtok_r(service + 4, " \n", &sav);
     if (strncasecmp(service, "reg", 3) == 0) {
+        active_services_mutex.lock();
+        if (active_services.find(cmd) == active_services.end()) {
+            fprintf(stderr, "reg request from %s, but not an active service?\n",
```
markhannum: syslog :)

mponomar: Heh.  Thanks Mark.

db/db_tunables.h
mponomar: Thanks Adi, I didn't think these existed.  No reason to have 2 sets of options for the same thing.

PR #443, opened 2017-10-10
verify indexes queries dont need schemalk

ldennis:


adizaimi:
Looks good to me. Did master fail the test you are adding here?

ldennis:
Yes, after adding the same sleep to master.

db/sqlinterfaces.c
```diff
@@ -5740,11 +5740,16 @@ static int execute_verify_indexes(struct sqlthdstate *thd,
 {
     int rc;
     if (thd->sqldb == NULL) {
-        rdlock_schema_lk();
-        rc = sqlengine_prepare_engine(thd, clnt, 1);
-        unlock_schema_lk();
-        if (rc) {
-            return rc;
+        /* open sqlite db without copying rootpages */
+        rc = sqlite3_open_serial("db", &thd->sqldb, thd);
```
akshatsikarwar: Won't this fail if a schemachange is updating schema variables which is used to bootstrap sqlite vm during open?

ldennis: I guess only copying rootpages needs schemalk.

akshatsikarwar: Specifically, code in create_sqlmaster_record. You can latch on to one of those pointers which can get freed by the time you deref.

ldennis: I dont think `sqlite3_open_serial` derefs sqlmaster, no?

akshatsikarwar: umm - it should.

akshatsikarwar: `select * from sqlite_master` gets it from there (or cached rootpage info). Either way lock is required during copy.

ldennis: Yes, `select * from sqlite_master` needs to load schema from rootpage. However, schema is not loaded for `select 1` or CTEs. I dont copy rootpage here so I dont think schemalk is required.

akshatsikarwar: hmm - ok. This is fine then!

ldennis: I have concern on stack if we call vm inline given that we are already deep in the stack in blockprocessor. And other than bp threads, convert record thd in sc_records.c also needs the vm. So I would keep the dispatch sql code for now. I can time the performance diff later.

PR #454, opened 2017-10-16
Osql restart keeprqid

ldennis:
PORTED: osql_send_commit_logic restarts sock with same rqid for retries


ldennis:
Ported from R6

PR #455, opened 2017-10-17
DDL: ALTER TABLE + default 'CURRENT_TIMESTAMP'

nirbhayc:
A datetime column loses the default 'CURRENT_TIMESTAMP' on ALTER.


nirbhayc:
Reported by @mohitkhullar

mohitkhullar:
Thanks, this looks good.
Can we add drop index test over here too?

nirbhayc:
Thanks @mohitkhullar

PR #464, opened 2017-10-23
when durable_lsns is off, berkdb doesn't set durable lsn

dorinhogea:
this stops add_dummy trickle when there are no incoherent nodes but durable_lsn is disabled.


adizaimi:
Looks good.

dorinhogea:
thank you Adi

Issue #465, opened 2017-10-23
Downgrading master deadlocks


dorinhogea:
Cluster of two nodes, bring down replicant; master heartbeats, then watchdogs (downgrade never gets the bdb lock in write mode).

dorinhogea:
fixed

PR #467, opened 2017-10-23
when downgrading, respect bdblock desired to avoid watchdog

dorinhogea:
Fixes issue: https://github.com/bloomberg/comdb2/issues/465
In a nutshell, when downgrading, transactions on master node, which are about to wait for data replication, don't wait for replicants anymore.  Transactions caught in phase 1 (determining a minimum wait time) will loop continuously for rep_timeout_maxms (which is 5 minutes by default);  while looping the bdb lock is kept, so the downgrade cannot make progress.  In 60 seconds the node aborts due to watchdog trap.

The fix is to detect this scenario and jump out of the initial phase.
The jepsen test should catch this scenario.

### DIFF
--- a/bdb/rep.c
+++ b/bdb/rep.c
@@ -2994,6 +2994,7 @@ static int bdb_wait_for_seqnum_from_all_int(bdb_state_type *bdb_state,
     uint32_t nodegen;
     int num_successfully_acked = 0;
     int total_connected;
+    int lock_desired = 0;
 
     /* if we were passed a child, find his parent */
     if (bdb_state->parent)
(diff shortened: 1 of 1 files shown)


ldennis:
block processor has bdb locks?

dorinhogea:
There is one bdb lock, and any access to underlying berkdb subsystem acquire it in read-only mode.

ldennis:
add_dummy?

dorinhogea:
yes; but it can be any pending transaction; add_dummy will do for a quiet db

PR #469, opened 2017-10-23
cache columns row to avoid malloc/free penalty

adizaimi:
* Avoiding call to newsql_alloc_row() gets 1.5% savings from not mallocing and deallocing row. 
* Not mallocing but using alloca in make_retrow() gives 0.5% savings.


db/sqlinterfaces.c
```diff
@@ -1185,11 +1185,36 @@ int newsql_send_last_row(struct sqlclntstate *clnt, int is_begin,
                          malloc, 1);
 }
 
+static int newsql_realloc_row(int ncols, CDB2SQLRESPONSE__Column ***columns, int *columns_count)
+{
+    if(ncols <= *columns_count)
+        return 0; /* already have more columns set up; nothing to do */
+
+printf("AZ: realloc from %d to %d\n", *columns_count, ncols);
```
ldennis: remove trace

db/sqlinterfaces.c
```diff
@@ -1185,11 +1185,36 @@ int newsql_send_last_row(struct sqlclntstate *clnt, int is_begin,
...
+    int newtotsize = arrsz + ncols * sizeof(CDB2SQLRESPONSE__Column);
+
+    CDB2SQLRESPONSE__Column **loc_cols = *columns;
+    char *ptr = realloc(loc_cols, newtotsize);
+    if (!ptr) {
+        return -1;
+    }
+
+    loc_cols = (CDB2SQLRESPONSE__Column **) ptr;
+    for (int i = 0; i < ncols; i++) {
```
ldennis: The loop can start from `*columns_count` instead?

adizaimi: We realloc, so unfortunately we have to reassign new pointer locations for the individual colums.

ldennis: Sorry, you were right.

adizaimi:
thanks Lingzhi.

PR #473, opened 2017-10-24
rename dbtable parameter dbname to tablename

adizaimi:


nirbhayc:
Can we rename it to "name" instead? tbl->name looks equally good + less typing.

adizaimi:
Sure. I think in a second iteration, when we rename all the db->   objects to tbl-> ....
Thanks

PR #476, opened 2017-10-26
add test which exercises queing sql

adizaimi:


tests/queueing_sql.test/runit
```diff
@@ -0,0 +1,43 @@
...
+
+
+echo "create t1"
+cdb2sql ${CDB2_OPTIONS} $dbnm default "create table t1 (a int) "
+
+
+cnt=0
+while [ $cnt -lt 100 ] ; do
+   cdb2sql -s ${CDB2_OPTIONS} $dbnm default 'insert into t1 values(1)' &>> insert1.txt &
+   cdb2sql -s ${CDB2_OPTIONS} $dbnm default 'select sleep(10)' &>> output1.txt &
```
nirbhayc: I don't see output1.txt/insert1.txt being processed? What exactly are you trying to test?

adizaimi: As of now, we have no test that queues sql. This is a first pass, we can potentially enhance this. 
The idea is to have quantifiable and predictable size for the queue and the sql that executes immediately and that which is queued. 
Insert1.txt and output1.txt should at the least have size of 100. Maybe I can already test for that.

nirbhayc: Ok, got it. Will it be reasonable to also check/diff the contents captured via awk?

PR #477, opened 2017-10-26
remove persistent_info variable from thdpool as it has redundant info

adizaimi:


dorinhogea:
This is not redundant.  Sql string is freed by a different thread concurrently, you cannot simply print it here.  The persistent_info saves a copy so we don't need to synchronize with other (sql) threads.

adizaimi:
Thank you Dorin. Closing this.

PR #480, opened 2017-10-26
Rename a table

dorinhogea:


sqlite/parse.y
```diff
@@ -750,8 +750,14 @@ sortlist(A) ::= expr(Y) sortorder(Z). {
...
 }
 
-%type sortorder {int}
+cmd ::= rename_comdb2table.
 
+rename_comdb2table ::= dryrun(D) ALTER TABLE nm(X) RENAME TO nm(Y). {
+    comdb2WriteTransaction(pParse);
+    sqlite3AlterRenameTable(pParse,&X,&Y,D);
+}
+
```
nirbhayc: I'd move this to COMDB2 ALTER TABLE STATEMENT section. Also, RENAME could be added as an 'action' to ALTER TABLE like ADD/DROP COLUMN.

PR #483, opened 2017-10-26
Add SHA256 SQL function

mthorpe7:
Open to using a different hashing function. 

sha256 has a good tradeoff between performance and odds of clashes.

### DIFF
--- a/sqlite/func.c
+++ b/sqlite/func.c
@@ -17,6 +17,7 @@
 #include <stdlib.h>
 #include <assert.h>
 #include "vdbeInt.h"
+#include <openssl/sha.h>
 #include <unistd.h>
 #include <uuid/uuid.h>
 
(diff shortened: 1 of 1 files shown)


sqlite/func.c
```diff
@@ -1155,6 +1156,53 @@ static void nullifFunc(
...
+    case SQLITE_DECIMAL:
+    case SQLITE_INTEGER:
+    case SQLITE_FLOAT:
+    case SQLITE_BLOB:
+    case SQLITE_TEXT: {
+      SHA256_CTX sha256;
+      SHA256_Init(&sha256);
+
+      SHA256_Update(&sha256, data, len);
+      SHA256_Final(hash, &sha256);
```
c-bb: These 3 functions return an `int` return value, which should be checked.

Or use the `SHA256()` convenience function.

PR #486, opened 2017-10-27
Rename4

dorinhogea:
trying again


mponomar:
I was able to break the db by renaming t->t1, t1->t in a loop:

`[ERROR] bdb_rename_file: dbrename failed XXX.t_59f8c0db000c0000.datas7->XXX.t_59f8c0db00120000.datas7 -30998 DB_FILEOPEN: Rename or remove while file is open.
[ERROR] upgrade: open_dbs as master failed
[ERROR] get_csc2_file had an error for csc2 version 0 table t, rc: 15
[ERROR] get_csc2_file failed /home/mike/comdb2/master/db/tag.c:7061
[ERROR] rc: -1 len: 6 csc2:
(null)
[ERROR] Failed to load new .ONDISK
[alter table t rename to t1] failed with rc 203 
[alter table t1 rename to t] failed with rc -3 New table name exists
Retrieved 14 version for t
[alter table t rename to t1] failed with rc 240 stale version for table:t master:14 replicant:15
[alter table t1 rename to t] failed with rc -3 New table name exists
`
DB is not recoverable at this point:

I AM NEW MASTER NODE xps
[ERROR] get_csc2_file had an error for csc2 version 0 table t, rc: 15
[ERROR] get_csc2_file failed /home/mike/comdb2/master/db/comdb2.c:2021
[FATAL] could not load tables from the low level meta table
[FATAL] failed to start

berkdb/rep/rep_record.c
```diff
@@ -2477,9 +2477,10 @@ gap_check:		max_lsn_dbtp = NULL;
 
 		__memp_sync_out_of_band(dbenv, &rp->lsn);
 		break;
+	case DB___txn_regop_rowlocks:
+        printf("Here\n");
```
mponomar: Debug trace?

ldennis:
Why do we use op bpfunc instead of schemachange for this? When running in a tran, how do we abort a rename?

dorinhogea:
@mponomar : we don't drop those, neither for rename, nor for drop.  It prevents reuse of old version numbers when the same old table name is recreated.

bdb/file.c
```diff
@@ -917,10 +940,14 @@ int bdb_start_file_versioning_table(bdb_state_type *bdb_state,
         }
 
         /*update version*/
+        bdb_state->name = newtblname;
         if (bdb_new_file_version_index(bdb_state, tran, ixnum, version_num,
```
ldennis: new_version_num ?

dorinhogea: good catch, submitted a fix

ldennis:
oh.. the bpfunc code is not being used? Can you remove that as well?

dorinhogea:
@mponomar : what schema did you use to run the loop test?

mponomar:
```
cdb2sql mikedb "create table t(a integer)"
while :; do cdb2sql mikedb "alter table t rename to t1"; cdb2sql mikedb "alter table t1 rename to t"; done
```

dorinhogea:
Note: bpfunc was cut 0, but it was too limited mechanism to support transactional table rename.  I'll remove it.

dorinhogea:
I have removed bpfunc code, updated instrumentation, and fixed the loop test.

dorinhogea:
Interesting enough, closing files without flushing was the cause of the loop test failure.   A closed file that was not flushed doesn't stay closed if any of the dirty pages are flushed out; if the latter happens, berkeley db happily re-opens the closed file and renaming fails.

mponomar:
> Interesting enough, closing files without flushing was the cause of the loop test failure. A closed file that was not flushed doesn't stay closed if any of the dirty pages are flushed out; if the latter happens, berkeley db happily re-opens the closed file and renaming fails.

That explains a bug we were seeing in dropping/recreating queues as well.  Tables don't have that problem since we create files with new names.

dorinhogea:
Thanks for the review!

PR #492, opened 2017-10-30
oqslshadtbl uses tablename and tableversion

ldennis:


dorinhogea:
Why do we need extra cached tbl->tableversion, when it is available in db->tableversion, and it is protected by a table read lock?

ldennis:
db->tableversion is protected by table read lock for each sql statement but tableversion could change between two SQLs in the same client tran. In the case where table is alter'd in between, the format of the data written to shadow table may not be compatible. I was trying to detect that.
For example (in a stepper):
```
1 create table t1 {schema {int a int b}}
1 set transaction read committed
1 begin
1 insert into t1(a, b) values (1, 10)  // tbl->tableversion is 0
2 alter table t1 {schema {int b int a}}
1 insert into t1(a, b) values (2, 20) // tbl->tableversion is 0 but db->tableversion is 1
```

dorinhogea:
I thought you wanna protect against recover deadlock (where this check would move), when the table lock is lost.
The version check will stop all writes after a schema change, which seems a bit too conservative.  
I think that the sql engine used after an in transaction schema change need to be updated to reflect the new schema for the following n-transaction statements. This will automatically detect incompatibilities during parsing phase.  You might dedicate one engine for the rest of the transaction to minimize impact on the sql engine pool.

ldennis:
shadow table has no knowledge of the schema. sql engine wasnt the problem. The problem was we could interpret shadow table data written in old schema using the new schema for subsequent in-tran reads. (i.e. if T1 does a `select * from t1` in the above example, wrong result `(b=1, a=10), (b=20, a=2)` is returned)

dorinhogea:
I am arguing that is too low level and what has to happen is:

```
1 create table t1 {schema {int a int b}}
1 set transaction read committed
1 begin
1 insert into t1(a, b) values (1, 10)  // tbl->tableversion is 0
2 alter table t1 {schema {int b}}
```
Sqlite engine here needs to update schema!
```
1 insert into t1(a, b) values (2, 20) 
```
This is parsed with with new schema and should fail.
But:
```
1 insert into t1(b) values (20) 
```
this should succeed.  In the current PR it is failing.  This doesn't seem correct.

ldennis:
IMO in your case, the whole client tran is failing anyways and I want to make sure subsequent reads are correct. Also, I want to reject `1 insert into t1(a, b) values (1, 10)` whereas now the following is allowed and (b=1) is inserted for the 1st insert stmt instead of 10:
```
1 create table t1 {schema {int a int b}}
1 set transaction read committed
1 begin
1 insert into t1(a, b) values (1, 10)
2 alter table t1 {schema {int b}}
1 insert into t1(b) values (20)
1 commit
```
Therefore, I think allowing your query above doesnt seem correct either.

My concern was that if we allow mix of schema data in shadow table, it might fool the whole cursor layer (e.g. cached indexes in shadow table). This not only affects read committed but also snapshot and would make subsequent reads kinda tricky. But we can leave it for further discussion.

dorinhogea:
I think my example should work.  It might not work currently, but that doesn't look right, in read committed.
Can we at least have a simplified check in sqlinterfaces.c, that will simply return error after parsing if the table was altered and the statement is a write? We don't really need to execute it and fail while writing to shadow btrees.

ldennis:
> We don't really need to execute it and fail while writing to shadow btrees.
If it is the first time writing to that table, I think that should be allowed as there is no mix of schema in shadow. Also, if an unrelated table was schema changed, we dont need to care.

db/osqlshadtbl.c
```diff
@@ -418,7 +418,11 @@ static shad_tbl_t *create_shadtbl(struct BtCursor *pCur,
 
     tbl->seq = 0;
     tbl->env = env;
-    tbl->db = db;
+    tbl->tablename = strdup(db->tablename);
+    tbl->tableversion = db->tableversion;
```
dorinhogea: where do we free this?

ldennis: good catch, fixing

dorinhogea:
looks good otherwise.

PR #494, opened 2017-11-01
unlock mutex in pmux before returning

adizaimi:
Pmux was left in an unresponsive state, the bug is we should unlock before returning.


adizaimi:
Thank you.

PR #497, opened 2017-11-01
Cdb2sql min retries

markhannum:
Quick bug-fix: expose '-R' in cdb2sql.  This is the short-opt flag for setting minretries.


mponomar:
This change looks good.  But `cdb2_set_max_retries`/`cdb2_set_min_retries` really should have been set commands.

PR #501, opened 2017-11-02
make thread_type_key a thread local variable

adizaimi:


riverszhang89:
I disagree. pthread key is standard and portable.

adizaimi:
To summarize discussion from yesterday, we already use __thread variables elsewhere in comdb2, and they also are portable. According to valgrind this saves about 0.7% of cpu from avoiding the function call (ftr. non optimized build).
All in all, it does not harm anything.

riverszhang89:
On my Linux box, __thread is 2x faster than pthread key with and without GCC optimization. __thread is not standard but is widely supported. So portability isn't a concern anymore.

adizaimi:
Thanks for testing that Rivers.

PR #503, opened 2017-11-03
Maven mirror not working. Use apache central repository instead.

riverszhang89:
This should fix the smoke test.


riverszhang89:
I'm merging this so folks can test their branches.

PR #509, opened 2017-11-04
Generate cnonce faster by caching a prefix for it

adizaimi:
* cache a prefix (machine and PID) of the cnonce
* always print cnonce if it is set in the event logger


mohitkhullar:
looks good

adizaimi:
Thanks Mohit.

PR #510, opened 2017-11-06
Testcase for the SET statement bugfix (commit 4841d63).

riverszhang89:
@markhannum Thank you for fixing it!


riverszhang89:
Thanks a lot Adi.

PR #511, opened 2017-11-06
Make Comdb2 server build without SSL/TLS support.

riverszhang89:
The pull request fixes server compilation when `-DWITH_SSL=0`. `WITH_SSL` is also made a CMake option to allow easy configuration.

Two symbols, `CLIENT_ssl_new_ctx` and `CLIENT_ssl_init`, are removed from `cdb2api.o`. Instead they're compiled into `ssl_support.o`.


adizaimi:
This looks good to me.

CMakeLists.txt
```diff
@@ -10,6 +10,13 @@ endif()
 #-g
 #endif()
 
+option(WITH_SSL "Turn OFF to compile without SSL" ON)
```
nirbhayc: The check below is redundant. OPTION() should take care of enabling/disabling the option.

riverszhang89: I don't think `option()` automatically adds a compile definition. The check is actually necessary.

nirbhayc: I stand corrected. Thank you.

riverszhang89:
@adizaimi @nirbhayc Thank you for the review.

PR #512, opened 2017-11-06
 tackle more clang warnings

adizaimi:


bdb/cursor.c
```diff
@@ -2587,7 +2584,7 @@ static int bdb_update_pglogs_fileid_queues(
         }
 
         // Sanity
-        if (chk = fileid_queue->queue_keys.bot)
+        if ((chk = fileid_queue->queue_keys.bot))
```
akshatsikarwar: Although this shuts up the compiler just as well, I prefer exlicit check e.g. `if ((chk = fileid_queue->queue_keys.bot) != 0)` for added readability (like berk code does).

adizaimi: I'm actually ambivalent about which is more readable: 
 If (rc) 
Or
 If (rc != 0)

As such I think I will initially refrain from changing it one way or the other (there are  instances of the other in this PR). But if we agree on making such change, I can certainly do so.

bdb/bdb_verify.c
```diff
@@ -93,8 +93,8 @@ int bdb_verify(
 
     BDB_READLOCK("bdb_verify");
 
-    if (rc = bdb_state->dbenv->lock_id_flags(bdb_state->dbenv, &lid,
-                                             DB_LOCK_ID_READONLY)) {
+    if ((rc = bdb_state->dbenv->lock_id_flags(bdb_state->dbenv, &lid,
+                                             DB_LOCK_ID_READONLY))) {
```
akshatsikarwar: By itself `if(rc)` is more readable. We're not discussing that case though. Its about stuff like this one.. By the time the expression finishes, there are 3 parens closing. I prefer if this were: `if ((rc = bdb_state->dbenv->lock_id_flags(bdb_state->dbenv, &lid, DB_LOCK_ID_READONLY)) != 0)`.

bdb/bdb_osqlcur.c
```diff
@@ -193,6 +193,7 @@ int bdb_osql_shadow_is_bkfilled(bdb_cursor_ifn_t *pcur_ifn, int *bdberr)
         if (shadow->tables[cur->dbnum].ix_shadows)
             return shadow->tables[cur->dbnum].ix_shadows[cur->idx].bkfill;
         break;
+    default: break;
```
nirbhayc: I'd assert(0) this and similar cases below as such cases should not go unnoticed.

bdb/llmeta.c
```diff
@@ -6684,8 +6684,7 @@ static int __llmeta_preop_alias(struct llmeta_tablename_alias_key *key,
     strncpy(key->tablename_alias, tablename_alias,
             sizeof(key->tablename_alias));
 
-    if (llmeta_tablename_alias_key_put(key, key_buf, key_buf + key_buf_len) ==
-        NULL) {
+    if (llmeta_tablename_alias_key_put(key, (uint8_t *)key_buf, (uint8_t *)(key_buf + key_buf_len)) == NULL) {
```
nirbhayc: Line too long?

bdb/bdb_osqlcur.c
adizaimi: Ok, changing.

bdb/bdb_verify.c
adizaimi: ok, applying.

riverszhang89:
Could you please fix up the format?

bdb/cursor.c
dorinhogea: Please don't write this as ((a=b)).  Please either use a=b;if(a), or  if((a=n)!=0).

akshatsikarwar:
Looks good

adizaimi:
Thanks Akshat and Nirbhay.

PR #516, opened 2017-11-08
Preventing tag thread from using freed resources.

riverszhang89:
A hardware controller issue can cause a tag request to take more than the time limit (1000 seconds) to finish. The parent appsock thread will then free resources, but leave the tag thread running. The tag thread will be at risk of using those freed resources.

The `reply_done` flag in `struct buf_lock_t` is extended to address the issue. If the child tag thread has timed out, the parent appsock thread marks the flag DISCARD without freeing those resources which are likely used by the tag thread. The tag thread later frees the resources.

### DIFF
--- a/db/handle_buf.c
+++ b/db/handle_buf.c
@@ -414,7 +414,7 @@ int free_bigbuf(uint8_t *p_buf, struct buf_lock_t *p_slock)
 {
     if (p_slock == NULL)
         return 0;
-    p_slock->reply_done = 1;
+    p_slock->reply_state = REPLY_STATE_DONE;
     LOCK(&buf_lock) { pool_relablk(p_bufs, p_buf); }
     UNLOCK(&buf_lock);
     pthread_cond_signal(&(p_slock->wait_cond));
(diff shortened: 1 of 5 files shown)


db/handle_buf.c
```diff
@@ -991,12 +990,23 @@ int handle_socketrequest(SBUF2 *sb, int *keepsocket, int wrongdb)
...
+             * be DONE and we won't go inside loop again.
+             * If the worker thread give signal, after we check for reply_state
+             * the condition will timeout after 1 sec and variable will be
              * checked again. */
-            while (p_slock->reply_done == 0) {
+            while (p_slock->reply_state != REPLY_STATE_DONE) {
+                /* Assuming here that a tag request can't be 1000 second long.
+                   We have to exit in cases of error scenarios of
+                   offloading block requests. */
+                if (num_sec > TAG_MAX_WAITS) {
```
nirbhayc: TAG_MAX_WAITS: turn this into a tunable with default 1000 secs?

riverszhang89: This is tag code. I'd rather not put too much effort into it.

riverszhang89:
This is ported from R6.

PR #519, opened 2017-11-09
enable asserts

adizaimi:


CMakeLists.txt
```diff
@@ -17,6 +17,10 @@ if(COMDB2_BUILD_STATIC)
   SET(CMAKE_FIND_LIBRARY_SUFFIXES .a ${CMAKE_FIND_LIBRARY_SUFFIXES})
 endif()
 
+if(${CMAKE_BUILD_TYPE} STREQUAL Debug)
```
akshatsikarwar: Lets not do this - posted proposal in IB..

adizaimi: As discussed offline, we will enable SQLITE_DEBUG so that all correctness checks will occur, and for build type Release, we will define NDEBUG so NO asserts will occur.

akshatsikarwar: Move this along with other SQLite add_definitions. No need for this noise at top level.

adizaimi: Asserts in comdb2.c do not work if i move this in sqlite/CMakeLists.txt. I guess it applies to only sqlite/ ?

akshatsikarwar: Put it in: `sqlite/definitions.cmake`. It'll work.

mohitkhullar:
comdb2expert test should work now

adizaimi:
Thank you Mohit and Akshat.

PR #468, opened 2017-10-23
Network failure test

mponomar:
Test Comdb2 with 2 redundant networks where one fails.

PR #520, opened 2017-11-09
Write limit

markhannum:
Provide a tunable that limits writes on the master


tests/max_wr_rows.test/runit
```diff
@@ -0,0 +1,169 @@
...
+
+[[ "$debug" == 1 ]] && set -x
+db=$1
+
+if [[ -z "$stage" ]]; then
+    echo "Setting stage to 'default' (set 'stage' var to override)."
+    stage="default"
+fi
+
+if [[ -n "$CLUSTER" ]]; then
```
riverszhang89: You could export CLUSTER to empty in the Mikefile so the database will come up standalone. For instance,
https://github.com/bloomberg/comdb2/pull/468/files#diff-7091688b541d5294bd463281cf3ae3fdR10

markhannum: Oh .. yeah .. that's a great idea .. tx Rivers!

PR #521, opened 2017-11-10
use rwlock instead of mutex

dorinhogea:


adizaimi:
Looks good.

dorinhogea:
thank you Adi

PR #523, opened 2017-11-10
fix logic

dorinhogea:


adizaimi:
Looks great. This change makes local cached stmts work in my test. Thank you Dorin.

dorinhogea:
per Adi, timeout not related to the change here.

dorinhogea:
thank you for the review

PR #528, opened 2017-11-11
Cleanup some more clang warnings

adizaimi:
Down to about 530 warnings (from about 1400) after this.

### DIFF
--- a/bbinc/fwd_types.h
+++ b/bbinc/fwd_types.h
@@ -14,13 +14,22 @@
    limitations under the License.
  */
 
-#ifndef __bdb_net_h__
-#define __bdb_net_h__
+/*
+ * forward declared types
+ */
 
-#include "fwd_types.h"
+#ifndef __fwd_types_h__
+#define __fwd_types_h__
 
-char *print_addr(struct sockaddr_in *addr, char *buf);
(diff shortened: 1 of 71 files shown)


bbinc/fwd_types.h
```diff
@@ -0,0 +1,35 @@
...
+/*
+ * forward declared types
+ */
+
+#ifndef __fwd_types_h__
+#define __fwd_types_h__
+
+typedef struct bdb_state_tag bdb_state_type;
+typedef struct UnpackedRecord UnpackedRecord;
+typedef struct Vdbe Vdbe;
```
riverszhang89: It feels wrong to place SQLite definitions in `bbinc` and have SQLite include the file.

adizaimi: Yeah, maybe i'll move it to sqlite/

adizaimi:
2017/11/11 18:26:02 [ERROR] udp_bind:bind: Address already in use
for both the failed tests above.

adizaimi:
Thank you Mike.

adizaimi:
> Aren’t these in db/sql.h
Yes, but i had to remove it from there because of the warning of duplicate definition. Moving them to one place seems to be the only solution.

akshatsikarwar:
That's cool. Didn't realize you did that.

PR #530, opened 2017-11-13
baddb remsql test and fix

dorinhogea:


dorinhogea:
thanks for the review!

Issue #315, opened 2017-08-08
Memory leak in eventlog


nirbhayc:
Apparently, memory alloced in eventlog_context never gets freed.
```
static void eventlog_context(cson_object *obj, const struct reqlogger *logger)
{
    cson_value *contexts = cson_value_new_array();
.. snip..
}  
```
```
==24249== 120 bytes in 3 blocks are definitely lost in loss record 379 of 505
==24249==    at 0x667BB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==24249==    by 0x94F25D: cson_malloc (cson_amalgamation_core.c:1797)
==24249==    by 0x94F25D: cson_value_new (cson_amalgamation_core.c:2298)
==24249==    by 0x94F25D: cson_value_array_alloc (cson_amalgamation_core.c:2366)
==24249==    by 0x94F25D: cson_value_new_array (cson_amalgamation_core.c:2401)
==24249==    by 0x42D34B: eventlog_context (eventlog.c:348)
==24249==    by 0x42DC46: eventlog_add_int (eventlog.c:480)
==24249==    by 0x42DD9A: eventlog_add (eventlog.c:511)
==24249==    by 0x5067C8: reqlog_end_request (reqlog.c:1723)
==24249==    by 0x4D5FFC: sql_statement_done (sqlinterfaces.c:1510)
==24249==    by 0x4E0BB4: sqlite_done (sqlinterfaces.c:5362)
==24249==    by 0x4E1115: handle_sqlite_requests (sqlinterfaces.c:5483)
==24249==    by 0x4E126D: execute_sql_query (sqlinterfaces.c:5532)
==24249==    by 0x4E1F8A: sqlengine_work_appsock (sqlinterfaces.c:5847)
==24249==    by 0x4E2155: sqlengine_work_appsock_pp (sqlinterfaces.c:5904)
```

mponomar:
Taking this opportunity to clean up this and other blocks reported by valgrind on exit.  This one is a legitimate leak, others aren't.

adizaimi:
Fixed in https://github.com/bloomberg/comdb2/pull/540/

PR #534, opened 2017-11-13
 call reqlog_reset_logger at the end of reqlog_end_request

adizaimi:
Currently we reset logger at the beginning of a request, thus such cleanup is in the critical path of real work to be done.
By resetting logger at the end of reqlog_end_request, we are doing such cleanup after we send last row to client, thus decreasing the delay for client getting back a response.


adizaimi:
Thank you Mike.

nirbhayc:
@adizaimi : Could you check if this also fixes https://github.com/bloomberg/comdb2/issues/315 ?

PR #535, opened 2017-11-13
cleanup clnt when done, issue was context from sp

adizaimi:
Cleanup clnt object when clnt is about to go out of scope. 
Currently we are leaking by not performing such cleanup, ex. context member from SP execution.

### DIFF
--- a/db/sqlinterfaces.c
+++ b/db/sqlinterfaces.c
@@ -6344,6 +6344,51 @@ static inline int tdef_to_tranlevel(int tdef)
     }
 }
 
+void cleanup_clnt(struct sqlclntstate *clnt)
+{
+    if (clnt->saved_errstr) {
+        free(clnt->saved_errstr);
+        clnt->saved_errstr = NULL;
+    }    
+
+    if (clnt->context) {
+        for (int i = 0; i < clnt->ncontext; i++) {
+            free(clnt->context[i]);
+        }
+        free(clnt->context);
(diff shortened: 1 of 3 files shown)


db/sqlinterfaces.c
```diff
@@ -6344,6 +6344,51 @@ static inline int tdef_to_tranlevel(int tdef)
...
+        currangearr_free(clnt->selectv_arr);
+        clnt->selectv_arr = NULL;
+    }
+
+    if (clnt->arr) {
+        currangearr_free(clnt->arr);
+        clnt->arr = NULL;
+    }
+
+    if (clnt->spversion.version_str) {
```
akshatsikarwar: Not sure we should do this.. If next run matches sp + version, then there is no need to fetch sp code again.

adizaimi: reset_clnt() does the following:
```
clnt->spversion.version_num = 0;
free(clnt->spversion.version_str);
clnt->spversion.version_str = NULL;
```
So i think that right before clnt destruction this should be not an issue. Doing so in reset_clnt() may need another look?

akshatsikarwar: OK to do if clnt is getting destroyed.

akshatsikarwar: Sorry! This is OK to do (even during reset) - the version check is done against clnt->sp.

PR #536, opened 2017-11-14
send cached buffer for last row for requests with no effect

adizaimi:


db/sqlinterfaces.c
```diff
@@ -1156,9 +1151,56 @@ static int fill_snapinfo(struct sqlclntstate *clnt, unsigned int *file,
...
+    pthread_mutex_unlock(&clnt->write_lock);
+    sbuf2flush(sb);
+}
+
+
 int newsql_send_last_row(struct sqlclntstate *clnt, int is_begin,
                          const char *func, int line)
 {
+    if (!(get_high_availability(clnt)) &&
+        clnt->effects.num_selected == 1 && clnt->effects.num_updated == 0 &&
```
nirbhayc: Why is num_selected 1?

PR #537, opened 2017-11-15
Schema change for remote sql: fixing adding a column to remote db

dorinhogea:


ldennis:
What was the problem? You might also want to add a testcase for that?

PR #538, opened 2017-11-16
pmux.test: Assign a port range that is not reserved by portmux.tsk.

riverszhang89:


riverszhang89:
I am merging the pull request so that I don't have to fix portmux on the test machine again.

PR #539, opened 2017-11-16
Rep verify

markhannum:
Only grab the bdb_writelock if this replication message will result in a match which will run recovery


markhannum:
Reviewed with Mike (& made some changes)

PR #541, opened 2017-11-17
SQLite: Fix broken parser.

riverszhang89:
SQLite parser is broken when compiled with GCC 5.4.0 on Ubuntu 16.04. For instance:

```
$ gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.5) 5.4.0 20160609
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ cdb2sql mydb 'insert into t values(1)'
[insert into t values(1)] failed with rc -3 near "into": syntax error
```

It can be fixed by moving the `RENAME` rule to the end of the grammar file (See https://github.com/bloomberg/comdb2/pull/237).

### DIFF
--- a/sqlite/parse.y
+++ b/sqlite/parse.y
@@ -756,13 +756,6 @@ sortlist(A) ::= expr(Y) sortorder(Z). {
   sqlite3ExprListSetSortOrder(A,Z);
 }
 
-cmd ::= rename_comdb2table.
-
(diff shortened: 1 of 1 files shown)


riverszhang89:
@ldennis Thanks for the review!

PR #545, opened 2017-11-20
fix reqlog dump invalid read

ldennis:


adizaimi:
looks good, vsnprintf returns *not* the number of chars written. 
Thanks for finding this.

ldennis:
Thanks Adi.

PR #551, opened 2017-11-28
Add hibernate dialect

aibolit:
This is essentially a copy of https://github.com/gwenn/sqlite-dialect/blob/master/src/main/java/org/hibernate/dialect/SQLiteDialect.java (unlicense)
Except, in order to make DDL work, I had to change the IdentityColumnSupport to be the default IdentityColumnSupportImpl because comdb2 doesn't currently support auto-increment indentity columns.

```
./gradlew install
```

___
I am currently making use of this in https://github.com/aibolit/Comdb2Test
It requires a local comdb2 named 'testdb'
```
./gradlew bootRun
```
to see the sql output from hibernate. It will create the tables and insert some rows.

The classes are in grails-app/domain
The entities are created at grails-app/init/BootStrap.groovy


riverszhang89:
Thank you for the patch! I'm a bit concerned about the copyright, although the repository is unlicensed. @kpfleming Could you please comment on this?

kpfleming:
The author of that code has indicated that they do not wish to retain copyright on it; while it's not ideal to do it that way (since there are places in the world where copyright ownership cannot be relinquished) and they should have chosen a suitable statement like CC-0, as long as the new file retains this notice and suitable links/references to the source of the file and the author, it should be fine for inclusion.

aibolit:
Added a reference at the top of the file to the original commit.

cdb2hibernatedialect/src/main/java/org/hibernate/dialect/Comdb2Dialect.java
```diff
@@ -0,0 +1,352 @@
...
+  /* not case insensitive for unicode characters by default (ICU extension needed)
+	public boolean supportsCaseInsensitiveLike() {
+    return true;
+  }
+  */
+
+	@Override
+	public boolean doesReadCommittedCauseWritersToBlockReaders() {
+		// TODO Validate (WAL mode...)
+		return true;
```
riverszhang89: Should be false.

cdb2hibernatedialect/src/main/java/org/hibernate/dialect/Comdb2Dialect.java
```diff
@@ -0,0 +1,352 @@
...
+	}
+
+	@Override
+	public boolean doesRepeatableReadCauseReadersToBlockWriters() {
+		return true;
```
riverszhang89: Ditto.

cdb2hibernatedialect/src/main/java/org/hibernate/dialect/Comdb2Dialect.java
```diff
@@ -0,0 +1,352 @@
...
+		//registerColumnType(Types.DOUBLE, "double");
+		registerColumnType( Types.DECIMAL, "decimal" );
+		registerColumnType( Types.CHAR, "char" );
+		registerColumnType( Types.LONGVARCHAR, "longvarchar" );
+		registerColumnType( Types.TIMESTAMP, "datetime" );
+		registerColumnType( Types.BINARY, "blob" );
+		registerColumnType( Types.VARBINARY, "blob" );
+		registerColumnType( Types.LONGVARBINARY, "blob" );
+
+		registerFunction( "concat", new VarArgsSQLFunction( StandardBasicTypes.STRING, "", "||", "" ) );
```
riverszhang89: Comdb2 support a lot more functions than the list.

cdb2hibernatedialect/src/main/java/org/hibernate/dialect/Comdb2Dialect.java
```diff
@@ -0,0 +1,314 @@
...
+        registerColumnType(Types.LONGVARBINARY, "blob");
+        registerColumnType(Types.BIGINT, "longlong");
+
+        registerFunction("concat", new VarArgsSQLFunction(StandardBasicTypes.STRING, "", "||", ""));
+        registerFunction("mod", new SQLFunctionTemplate(StandardBasicTypes.INTEGER, "?1 % ?2"));
+        registerFunction("quote", new StandardSQLFunction("quote", StandardBasicTypes.STRING));
+        registerFunction("random", new NoArgSQLFunction("random", StandardBasicTypes.INTEGER));
+        registerFunction("round", new StandardSQLFunction("round"));
+        registerFunction("substr", new StandardSQLFunction("substr", StandardBasicTypes.STRING));
+        registerFunction("trim", new AbstractAnsiTrimEmulationFunction() {
```
riverszhang89: I can complete the function list. Would you mind if I push a few commits to your branch?

aibolit: yes; thank you. I have added you as a collaborator

aibolit:
Looks good to me. The gradle wrapper wasn't working so I updated it with 
`$ gradle wrapper --gradle-version 4.2`

riverszhang89:
Thanks @aibolit !

PR #558, opened 2017-12-04
fix live_sc datacopy indexes

ldennis:
convert ondisk record to ".NEW.ONDISK" before calling into live_sc index add code if table has datacopy index.


adizaimi:
good to merge. Thanks Lingzhi.

PR #565, opened 2017-12-05
fix cost and remote sql naming

dorinhogea:
Currently reprepared statements miscount cost.
Also, remote tables show as temporary indexes.


adizaimi:
looks good. (this scenario seems like it could use a test case)
Thanks

dorinhogea:
Thank you Adi

PR #570, opened 2017-12-08
Misc updates

akshatsikarwar:

PR #572, opened 2017-12-09
Test with db name with max length

adizaimi:
Have a test with dbname with max length
Fix cdb2sql to connect to portmux and work ok with this name length


tests/reallylongtestnametomakesurethatthedbworksfinewithlongname.test/runit
```diff
@@ -0,0 +1,314 @@
+#!/usr/bin/env bash
```
riverszhang89: I think an empty `runit` file is good enough to test your change. It'd be great if we could keep tests simple and easy to understand.

tools/cdb2sql/cdb2sql.c
```diff
@@ -111,6 +111,11 @@ void dumpstring(FILE *f, char *s, int quotes, int quote_quotes)
         fprintf(f, "'");
 }
 
+static inline void verbose_print(const char *str)
```
riverszhang89: I'd prefer to keep these changes out of this patch. These are irrelevant and might need more work. For example, the `verbose_print` function does not support variable length arguments which makes it not quite useful.

cdb2api/cdb2api.c
```diff
@@ -4432,13 +4461,18 @@ static int cdb2_get_dbhosts(cdb2_hndl_tp *hndl)
...
+            return rc;
+
         num_retry++;
         poll(NULL, 0, 250); // Sleep for 250ms everytime and total of 5 seconds
         rc = 0;
-    } else if (rc) {
-        return rc;
     }
+    if (hndl->debug_trace)
+        fprintf(stderr, "td %d %s line %d: get blah num_retry=%d\n",
```
riverszhang89: What is `blah`?

tests/reallylongtestnametomakesurethatthedbworksfinewithlongname.test/runit
adizaimi: We need to test client connecting to long named db.
This is a copy of basic.test ... i can put that in the readme.

cdb2api/cdb2api.c
adizaimi: function name, obviously :) 
clearly I need to clean it up.
thanks

tests/reallylongtestnametomakesurethatthedbworksfinewithlongname.test/runit
riverszhang89: `setup` will issue `SELECT 1` to ping the database. It already tests your change in a way.

bb/portmuxusr.c
```diff
@@ -206,7 +206,7 @@ int portmux_register(const char *app, const char *service, const char *instance)
 int portmux_deregister(const char *app, const char *service,
                        const char *instance)
 {
-    char name[NAMELEN];
+    char name[NAMELEN * 2];
```
nirbhayc: Why?

adizaimi: Same as all the other instances of this fix: 
rc = snprintf(name, sizeof(name), "%s/%s/%s", app, service, instance);   
is definitely > 64 chars. Just making it easy and doing *2 rather than measuring the real app and service lengths.

nirbhayc:
https://github.com/bloomberg/comdb2/pull/72

adizaimi:
Actually Akshat fixed some things in https://github.com/bloomberg/comdb2/pull/570 similar to #72.
This is different because it deals with the db name and registering with pmux.

tests/reallylongtestnametomakesurethatthedbworksfinewithlongname.test/runit
adizaimi: I added a comment that runit is copied from basic.test, maybe i can in Makefile do the cp so there would be no runit in that directory.

adizaimi:
i left this piece which appears to break formatting because i like the setup of the struct better the way we had it. 
Thanks for looking at this.

PR #574, opened 2017-12-09
Handle dbstore in vutf8 correctly

mohitkhullar:


adizaimi:
Looks good. 
Can you please modify the testcase sc_add_vutf8.test to catch this?
Thanks

mohitkhullar:
already handled with alter t1_3 in that test, will change the comment there.

PR #577, opened 2017-12-12
fix compile error

adizaimi:


riverszhang89:
You'd need `#include <limits.h>` only to bring in `PATH_MAX`.

PR #579, opened 2017-12-12
fix running clustered: getting subnet segfaults

adizaimi:
* return if we can't get host name (bug was parameter he is null)
* use getnameinfo instead of gethostbyaddr

### DIFF
--- a/net/net.c
+++ b/net/net.c
@@ -5224,6 +5224,7 @@ static void get_subnet_incomming_syn(host_node_type *host_node_ptr)
     size_t lcl_len = sizeof(lcl_addr_inet);
     struct hostent *he = NULL;
 
+    /* get local address of connection */
     int ret =
         getsockname(host_node_ptr->fd, &lcl_addr_inet, (socklen_t *)&lcl_len);
     if (ret != 0) {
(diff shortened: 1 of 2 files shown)


net/net.c
```diff
@@ -5222,25 +5222,37 @@ static void get_subnet_incomming_syn(host_node_type *host_node_ptr)
...
+    }
+    
+    char host[NI_MAXHOST], service[NI_MAXSERV];
+    int s = getnameinfo((struct sockaddr *) &lcl_addr_inet,
+            lcl_len, host, NI_MAXHOST,
+            service, NI_MAXSERV, NI_NUMERICSERV);
+
+    if (s != 0) {
+        logmsg(LOGMSG_ERROR, "Error from getaddrinfo: %s\n", gai_strerror(s));
+        logmsg(LOGMSG_INFO, "Incoming connection from name: %s (%s:%u)\n",
```
dorinhogea: C&P of below, extra %s

dorinhogea:
Also format needed

net/net.c
adizaimi: I'm pretty sure i had it printf 'unknown' in some iteration of the fix, and then i switched to logmsg. Thanks for catching.

Issue #559, opened 2017-12-04
Serializability issue when running Concurrent TPCC 


saatviks:
I've got ComDB2 running the TPCC benchmark on OLTPBench with submitting a single transaction at a time. But when trying to run concurrently by submitting multiple transactions at a time, there are errors in the results returned. 
Transactions are running in `READ COMMITTED` isolation level.
Currently 2(`OrderStatus` and `StockLevel`) of 5 procedures in the TPCC benchmark are running successfully when submitting transactions concurrently - The other 3 fail.
Is there any setting I need to tune for Concurrency?

I tried using the `SERIALIZABLE` isolation level. Now depending on the transaction I get different types of errors. Here are some:-
1. `java.sql.SQLException: [RC = 2] Table 'new_order' unable to delete genid =5a23671f80890002 rc=4`
2. `java.sql.SQLIntegrityConstraintViolationException: [RC = -103] selectv constraints`
3. `java.sql.SQLException: [RC = 2] unable to update record rc = 4`
4. `java.sql.SQLException: [RC = 230] transaction is not serializable`

akshatsikarwar:
I hope you reviewed the transaction docs I sent you earlier.

There really is only one way to address this: retry the transaction.

Depending on how you run, either the system can automatically retry for you, or your application will have to. Consider following:
```
create table t (i int)
insert into t values(1)
```
Consider also that the rowid for above is 0.

Now 2 transactions run concurrently: `update t set i = i + 1`. Both statements will identify rowid 0 and ask master to set the new value. First transaction to commit will update new value and bump rowid to 1. Second transaction will fail to find rowid 0 and fail with a _verify_ error. In this case comdb2 will rerun your update stmt and find rowid 1 and then update it. This is called _verifyretry_.

Running under snapshot or serializable isolation disables verifyretry (rerun of queries will not see new rows -- thats the whole point of snapshot.)

Running selectv stmt will also disable verifyretry -- the whole point is for application to be notified of verify error.

It is also possible to disable verifyretry. Run stmt: `set verifyretry off`. If you do this, the second transaction from our example above will fail with verify error.

Verifyretry is also disabled if you want to see per stmt effects (as opposed to per transaction effects). That may be default for JDBC or maybe that is how you are running it.

For most usecases,  it is sufficient to run just the default isolation level and without selectv (i.e. no need to selectv the same rows that update stmt will operate on.) In default isolation level, a transaction can't see its own side-effects.

[...]

PR #583, opened 2017-12-13
cdb2jdbc: Providing a JDBC URL parameter to toggle verifyretry.

riverszhang89:
Addresses #559.


riverszhang89:
I don't quite understand the crash. I did not change any server side code however the crash seems steadily reproducible. I've rebased my branch onto master. Let's see if the test passes this time.

PR #586, opened 2017-12-14
comdb2ar and cdb2api fix: correctly check return code from pmux

adizaimi:
currently clustered tests are not working because copycomdb2 which
relies on comdb2ar fails--failure is from return code from portmux:
ar requests rte and portmux returns -1 which means failure but
ar does not treat it as such but rather continues operating on the
fd which is connection to portmux and requests to hold logs on that
fd fail. By checking for return code, ar will proceed correctly
(-1 would mean can't forward fd thus the db is most likely down).

### DIFF
--- a/cdb2api/cdb2api.c
+++ b/cdb2api/cdb2api.c
@@ -1485,7 +1485,7 @@ static int cdb2portmux_route(const char *remote_host, char *app, char *service,
     sbuf2flush(ss);
     res[0] = 0;
     sbuf2gets(res, sizeof(res), ss);
-    if (res[0] == 0) {
+    if (res[0] != '0') {
         sbuf2close(ss);
         return -1;
     }
--- a/tests/basic.test/runit
+++ b/tests/basic.test/runit
@@ -226,12 +226,12 @@ for node in $cluster ; do
 
     port=`cdb2sql --tabs ${CDB2_OPTIONS} $dbnm --host $node "select comdb2_port()"`
     echo "connect to db with direct port info"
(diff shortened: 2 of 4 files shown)


adizaimi:
We were discussing with Mike about the fallback (which i had written but ommitted here): fallback is not needed because if rte does not work, get will not work either. If rte returns -1, either the db is down or  something is really broken. 
FTR, portmux supports rte just fine, issue here is that ar does not continue successfully because rte returns -1 but ar does not interpret it as a failure to communicate to the db, thus continues operating erroneously on the fd.

akshatsikarwar:
Can you confirm that with portmux, comdb2ar is able to hold logfiles when db is up.

adizaimi:
Ok. I shall do that. It looked like portmux handles rte just fine.

adizaimi:
RTE from portmux does seem to work and hold logfiles:
In a running DB I send the following command manually:
```
copycomdb2 -x /bb/comdb2a/adi/master_branch/comdb2 /bb/comdb2a/adi/master_branch/tests/test_47073/basic47073/basic47073.lrl pnj-comdb2btda2:
```
and see the following output:
```
FULL CLUSTER CACHE COHERENCY
LOG DELETE ENABLED only up to and including log.0000000000
LOG DELETE POLICY: delete all eligible log files
Disabling log file deletion
sent recovery options: -recovery_lsn 1:2893780
Reenabling log file deletion
FULL CLUSTER CACHE COHERENCY
LOG DELETE ENABLED
LOG DELETE POLICY: delete all eligible log files
```

PR #589, opened 2017-12-14
fix analyze and userschema with stmt caching

adizaimi:


db/sqlinterfaces.c
```diff
@@ -3884,6 +3884,16 @@ static void clear_parameters_to_bind(struct sql_state *rec)
     }
 }
 
+static int dont_cache_sql(struct sqlclntstate *clnt, const char *sql)
+{
+    //if (clnt->ddlflag) return 1;
+    if (strncmp(sql, "analyze", 7) == 0) return 1;
```
mponomar: May want to do strncasecmp, and skip any leading spaces.

db/sql.h
```diff
@@ -55,6 +55,8 @@ enum transaction_level {
  * of appsock threads with small stacks. */
 
 #define MAX_HASH_SQL_LENGTH 8192
+#define MAX_USERNAME_LEN 17
+#define MAX_PASSWORD_LEN 19
```
nirbhayc: Would it be good to add them to cdb2_constants.h and expose via comdb2_limits table?

adizaimi: It would. We will address in a subsequent PR.

db/sqlinterfaces.c
adizaimi: I will check for case. Will consume whitespace when we initially set clnt->sql in a separate PR.
Testcase with uppercase ANALYZE you mean?

mponomar: I meant a test that the cache is properly cleared on user change.

adizaimi:
Whitespace consumed in https://github.com/bloomberg/comdb2/pull/591

Issue #580, opened 2017-12-12
Impossible to add many to one foreign key using DDL


aibolit:
Since the only way to add foreign keys is using the create statement, it should be possible to add dup keys using the create statement as well. This is currently not the case.

PR #590, opened 2017-12-14
DDL: Syntax to specify duplicate key and index name in CREATE TABLE

nirbhayc:
Addresses #580


ldennis:
Looks good to me

nirbhayc:
@ldennis thank you for having a look at the patch.

nirbhayc:
Will update the docs after I am done with some more DDL work (currently in progress).

PR #591, opened 2017-12-15
strip whitespace from clnt->sql string

adizaimi:


db/sqlinterfaces.c
```diff
@@ -1664,7 +1664,7 @@ void sql_set_sqlengine_state(struct sqlclntstate *clnt, char *file, int line,
 }
 
 /* skip spaces and tabs, requires at least one space */
-static char *skipws(char *str)
+static inline char *skipws(char *str)
 {
     if (str && *str && isspace(*str)) {
         while (*str && isspace(*str))
```
nirbhayc: Unnecessary caution? :)
```
    if (str) {
        while (isspace(*(str)))
            ++str;
    }
    return str;
```

db/sqlinterfaces.c
```diff
@@ -7677,24 +7665,24 @@ static int process_set_commands(struct sqlclntstate *clnt)
     num_commands = sql_query->n_set_flags;
     for (int ii = 0; ii < num_commands && rc == 0; ii++) {
         sqlstr = sql_query->set_flags[ii];
-        sqlstr = cdb2_skipws(sqlstr);
```
nirbhayc: I think it'd be better to offload this whole parsing of SET command to the parser, so that we will no longer have to worry about spaces, cases, etc. Not something related to this PR obviously, just a note.

db/sqlinterfaces.c
adizaimi: yes. adjusting : )

db/sqlinterfaces.c
adizaimi: Could not agree more.

PR #592, opened 2017-12-16
fix userschema issues

adizaimi:
* fix segfault when authentication is not on--return empty table
* check and return error when user and password exceed proper length
* add testcases

### DIFF
--- a/db/sqlinterfaces.c
+++ b/db/sqlinterfaces.c
@@ -7679,7 +7679,7 @@ static int process_set_commands(struct sqlclntstate *clnt)
         sqlstr = sql_query->set_flags[ii];
         sqlstr = cdb2_skipws(sqlstr);
         if (strncasecmp(sqlstr, "set", 3) == 0) {
-            char err[256]; 
+            char err[256];
             err[0] = '\0';
             if (gbl_extended_sql_debug_trace) {
                 logmsg(LOGMSG_ERROR,
@@ -7757,34 +7757,28 @@ static int process_set_commands(struct sqlclntstate *clnt)
                     rc = ii + 1;
                 }
             } else if (strncasecmp(sqlstr, "user", 4) == 0) {
-                char *savptr = sqlstr;
                 sqlstr += 4;
                 sqlstr = cdb2_skipws(sqlstr);
                 sqlite3Dequote(sqlstr);
                 if (strlen(sqlstr) >= sizeof(clnt->user)) {
                     snprintf(err, sizeof(err), "set user: '%s' exceeds %d characters",
                            sqlstr, sizeof(clnt->user) - 1);
-                    sqlstr = savptr;
                     rc = ii + 1;
                 } else {
                     clnt->have_user = 1;
-                    strncpy(clnt->user, sqlstr, sizeof(clnt->user) - 1);
-                    clnt->user[sizeof(clnt->user) - 1] = '\0';
+                    strcpy(clnt->user, sqlstr);
                 }
             } else if (strncasecmp(sqlstr, "password", 8) == 0) {
-                char *savptr = sqlstr;
                 sqlstr += 8;
                 sqlstr = cdb2_skipws(sqlstr);
                 sqlite3Dequote(sqlstr);
                 if (strlen(sqlstr) >= sizeof(clnt->password)) {
                     snprintf(err, sizeof(err), "set password: '%s' exceeds %d characters",
                            sqlstr, sizeof(clnt->password) - 1);
-                    sqlstr = savptr;
                     rc = ii + 1;
                 } else {
                     clnt->have_password = 1;
-                    strncpy(clnt->password, sqlstr, sizeof(clnt->password) - 1);
-                    clnt->password[sizeof(clnt->password) - 1] = '\0';
+                    strcpy(clnt->password, sqlstr);
                 }
             } else if (strncasecmp(sqlstr, "spversion", 9) == 0) {
                 clnt->spversion.version_num = 0;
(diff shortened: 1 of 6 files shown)


db/sqlinterfaces.c
```diff
@@ -7755,16 +7755,36 @@ static int process_set_commands(struct sqlclntstate *clnt)
...
                 }
             } else if (strncasecmp(sqlstr, "user", 4) == 0) {
+                char *savptr = sqlstr;
                 sqlstr += 4;
                 sqlstr = cdb2_skipws(sqlstr);
-                clnt->have_user = 1;
-                strncpy(clnt->user, sqlstr, sizeof(clnt->user));
+                if (strlen(sqlstr) >= sizeof(clnt->user)) {
+                    sqlstr = savptr;
+                    logmsg(LOGMSG_USER,
```
nirbhayc: I’d expect a test that fails with this error.

tests/userschema.test/t09.expected
```diff
@@ -0,0 +1,19 @@
...
+(t=2)
+(tablename='forecast@dcba', username='abcd', READ='N', WRITE='N', DDL='N')
+(tablename='forecast@dcba', username='dcba', READ='Y', WRITE='Y', DDL='Y')
+(tablename='forecast@dcba', username='default', READ='N', WRITE='N', DDL='N')
+(tablename='forecast@dcba', username='mohit', READ='Y', WRITE='Y', DDL='Y')
+(tablename='test@dcba', username='abcd', READ='N', WRITE='N', DDL='N')
+(tablename='test@dcba', username='dcba', READ='Y', WRITE='Y', DDL='Y')
+(tablename='test@dcba', username='default', READ='N', WRITE='N', DDL='N')
+(tablename='test@dcba', username='mohit', READ='Y', WRITE='Y', DDL='Y')
+[select * from test] failed with rc -3 Invalid set command 'user 12345678901234567'
```
nirbhayc: Why do we get this error on SELECT?

db/sqlinterfaces.c
```diff
@@ -7755,16 +7755,36 @@ static int process_set_commands(struct sqlclntstate *clnt)
...
+                           "set password parameter can not exceed length of %d "
+                           "characters\n", sizeof(clnt->password) - 1);
+                    rc = ii + 1;
+                } else {
+                    clnt->have_password = 1;
+                    strncpy(clnt->password, sqlstr, sizeof(clnt->password) - 1);
+                    clnt->password[sizeof(clnt->password) - 1] = '\0';
+                    sqlite3Dequote(clnt->password);
```
nirbhayc: I think the password should be dequoted first before validating its length. Ditto for username.

tests/userschema.test/t09.req
```diff
@@ -0,0 +1,26 @@
+select * from test
+select * from comdb2_tablepermissions order by tablename
+
+set user dcba
```
nirbhayc: Can you also add test for quoted user/password as well?

tests/userschema.test/t09.expected
adizaimi: We don't send anything to server until the select.

tests/userschema.test/t09.req
adizaimi: ok.

db/sqlinterfaces.c
adizaimi: good point.

db/sqlinterfaces.c
adizaimi: This is printed on the .db log file. It would be nice to return that to user.

db/sqlinterfaces.c
adizaimi: Good idea. Will apply.

db/sqlinterfaces.c
adizaimi: Changing errstr to return that.

db/sqlinterfaces.c
```diff
@@ -7755,16 +7757,35 @@ static int process_set_commands(struct sqlclntstate *clnt)
...
-                strncpy(clnt->user, sqlstr, sizeof(clnt->user));
+                sqlite3Dequote(sqlstr);
+                if (strlen(sqlstr) >= sizeof(clnt->user)) {
+                    snprintf(err, sizeof(err), "set user: '%s' exceeds %d characters",
+                           sqlstr, sizeof(clnt->user) - 1);
+                    sqlstr = savptr;
+                    rc = ii + 1;
+                } else {
+                    clnt->have_user = 1;
+                    strncpy(clnt->user, sqlstr, sizeof(clnt->user) - 1);
```
nirbhayc: strcpy should be safe here, we have already checked the size. That makes the below line redundant too. Ditto for password.

sqlite/ext/comdb2/tablepermissions.c
```diff
@@ -98,6 +98,8 @@ static int permissionsOpen(sqlite3_vtab *p, sqlite3_vtab_cursor **ppCursor){
   bdb_user_get_all(&pCur->ppUsers, &pCur->nUsers);
   unlock_schema_lk();
   *ppCursor = &pCur->base;
+  if (pCur->nUsers == 0)
+      return SQLITE_EMPTY;
```
nirbhayc: This isn’t right. IMHO, an empty table is perfectly normal, and shouldn’t be treated as error.

tests/userschema.test/t09.expected
nirbhayc: Huh..? On failure, I’d expect the SET command to report an error. But, it should probably be dealt in a separate PR.

sqlite/ext/comdb2/tablepermissions.c
adizaimi: Well, this table can only have content when user schema is turned on. This was meant to be like an indicator for that. But I agree that this looks not too good.

adizaimi:
Thank you Nirbhay. Appreciate your input.

PR #596, opened 2017-12-19
Fix: first query with cdb2api always goes to same node

mohitkhullar:


cdb2api/cdb2api.c
```diff
@@ -108,6 +108,7 @@ static void do_init_once(void)
     }
     _PID = getpid();
     _MACHINE_ID = gethostid();
+    srandom(time(0));
```
nirbhayc: How does this exactly help?

mohitkhullar: without this random https://github.com/bloomberg/comdb2/blob/master/cdb2api/cdb2api.c#L1648 is called with seed 1 and it will give same value for every new run.

mponomar: random/srandom aren't threadsafe.  srandom also alters process state that other calls may depend on.   We should consider replacing them with something like

https://github.com/bloomberg/comdb2/blob/master/cdb2api/cdb2api.c#L2496

nirbhayc: Okay

adizaimi:
This looks great. Thanks.

mohitkhullar:
Thanks

PR #597, opened 2017-12-19
Remove log of master dropping connection

mohitkhullar:
Master sends dbinfo back, and client should handle it correctly.


mohitkhullar:
thanks

PR #599, opened 2017-12-19
PORTED: osql repository duplicate sessions handling

ldennis:


ldennis:
This is in R6 already and it passes the socksql_master_swings.test..

PR #600, opened 2017-12-19
DDL: ALTER TABLE ADD/DROP INDEX/PK/FK

nirbhayc:


ldennis:
you may want to update the docs too

nirbhayc:
@ldennis Thanks for the review.

PR #606, opened 2017-12-22
Pmux fix

markhannum:
Mike fixed actually - I typed.

### DIFF
--- a/bdb/bdb.c
+++ b/bdb/bdb.c
@@ -157,6 +157,12 @@ char *bdb_whoismaster(bdb_state_type *bdb_state)
         return NULL;
 }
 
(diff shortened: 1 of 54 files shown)


adizaimi:
we can add to pmux.test and send commands to pmux to add, rem, stat, etc. I don't know what this fixes or if might be testable by sending commands.

adizaimi:
Also, please add description of what the problem was.

markhannum:
Ha .. yeah .. it was a terrible one .. pmux crashed when you telnet'd to 5105 & issued a 'used' command if there were no ports used.  So a pmux test would have absolutely helped this.

PR #607, opened 2017-12-22
try cleanup some more before exit

adizaimi:


bb/object_pool.c
```diff
@@ -312,6 +310,7 @@ int comdb2_objpool_destroy(comdb2_objpool_t op)
 
     /* clear access history */
     hash_for(op->history, hash_elem_free_wrapper, NULL);
+    hash_clear(op->history);
```
riverszhang89: You don't need to call `hash_clear()` as `hash_free()` clears entries as well.

adizaimi: good point, removing, thanks.

riverszhang89:
Wait.. You didn't remove the redundant code I pointed out..

adizaimi:
Doing it in https://github.com/bloomberg/comdb2/pull/620 -- changed in local branch but forgot to push. 
Thanks.

PR #609, opened 2017-12-22
Pmux hello init try2

markhannum:
A second fix for pmux. The is_hello flag wasn't being cleared for a new incoming connection. This bug allowed a non-is-hello message which were assigned the same file descriptor as a previous is-hello-message to be incorrectly erased from the active-services list. When this occurred, a client which is using rte routing will fail to connect against that node.

I'm putting in the pull request immediately, but I will not merge until I write a test for this and the last issue.

### DIFF
--- a/bdb/bdb.c
+++ b/bdb/bdb.c
@@ -157,6 +157,12 @@ char *bdb_whoismaster(bdb_state_type *bdb_state)
         return NULL;
 }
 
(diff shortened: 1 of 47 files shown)


markhannum:
pmux_hello tests both of the pmux bugs that were fixed today, and will break against versions of the database which do not have the fix.  This is ready for a review.

adizaimi:
looks good to me.

PR #616, opened 2017-12-27
Use policy for dbinfo and randomroom should downgrade to random

mohitkhullar:


adizaimi:
this looks good, however it would be great if we could have a test for this.
thanks

mohitkhullar:
Thanks

mohitkhullar:
Will add it once we run our tests in cluster

PR #617, opened 2017-12-28
more refactor cdb2api

adizaimi:
* clear few unnecessary gotos
* don't free buffer that we know its null
* remove sirace under test/tools
* fix error code check


riverszhang89:
Could you please fix the format.

cdb2api/cdb2api.c
```diff
@@ -4067,12 +4040,15 @@ int cdb2_bind_param(cdb2_hndl_tp *hndl, const char *varname, int type,
 int cdb2_bind_index(cdb2_hndl_tp *hndl, int index, int type,
                     const void *varaddr, int length)
 {
-    int rc = 0;
     pthread_once(&init_once, do_init_once);
+    if (log_calls)
```
mohitkhullar: We should log it at end and print rc

cdb2api/cdb2api.c
```diff
@@ -1976,38 +1975,32 @@ static int cdb2_effects_request(cdb2_hndl_tp *hndl)
...
-
     if (type ==
         RESPONSE_HEADER__SQL_RESPONSE) { /* This might be the error that
                                             happened within transaction. */
         hndl->firstresponse =
             cdb2__sqlresponse__unpack(NULL, len, hndl->first_buf);
-        hndl->error_in_trans =
-            cdb2_convert_error_code(hndl->firstresponse->error_code);
+        int error_code = hndl->firstresponse->error_code;
+        hndl->error_in_trans = cdb2_convert_error_code(error_code);
```
mohitkhullar: Why an extra variable?

adizaimi: left over from previous refactoring, will fix.

cdb2api/cdb2api.c
adizaimi: ok, was trying to avoid the goto, but will do.
Thanks

adizaimi: hah, there is no rc for this one, there might be for the other one i removed.

adizaimi: I don't think there is any difference if we put at function entry.

mohitkhullar: rc was removed from this function, we usually log rc when the function gives non-zero returns too.

PR #620, opened 2017-12-28
fix some clustered tests

adizaimi:
* sockpool_msg vers0 can only be 48 characters--dont talk to sockpool if msg lengh is larger
* fix verify to get table lock so db does not go away when its processing
* fix tunables i_am_master setting to ignore exact val


cdb2api/cdb2api.c
```diff
@@ -1180,11 +1180,13 @@ static int cdb2_socket_pool_get_ll(const char *typestr, int dbnum, int *port)
         }
     }
 
-    struct sockpool_msg_vers0 msg;
+    struct sockpool_msg_vers0 msg = {0};
     /* Please may I have a file descriptor */
-    bzero(&msg, sizeof(msg));
     msg.request = SOCKPOOL_REQUEST;
     msg.dbnum = dbnum;
+    if (strlen(typestr) >= sizeof(msg.typestr)) {
```
mohitkhullar: Add the logic for donate too, otherwise it will lead to lot of idle appsock connections.

adizaimi: Good thinking, thanks.

db/verify.c
```diff
@@ -268,30 +268,49 @@ int verify_table(const char *table, SBUF2 *sb, int progress_report_seconds,
              int attempt_fix, 
              int (*lua_callback)(void *, const char *), void *lua_params)
 {
-    struct dbtable *db;
-    blob_buffer_t blob_buf[MAXBLOBS];
     int rc = 0;
 
-    db = get_dbtable_by_name(table);
-    bzero(blob_buf, sizeof(blob_buf));
+    struct dbtable *db = get_dbtable_by_name(table);
```
mponomar: dbtable can go away right after we look it up - use schema_lk

adizaimi:
So it turns out that verify grabbing lock on table is not playing well with several other tests that attempt to write at the same time:
```
2018/01/05 16:01:33 db_comdb_verify: verify table 't6'
transaction 0 6f8fbb52-6bed-47f4-89bd-d520bf8b3d48 failed 500 times with verify errors
Replayed too many times, too high contention, failing id=0 rc=2 retries=500 [global retr=1211]
```

PR #621, opened 2017-12-29
Go serial for every regop which gets the schema-lock in write mode.

markhannum:
A replicant will abort after a pthread_rwlock_wrlock against the schema-lock returns "deadlock".  I've made a simple fix here, and forced any commit which requires the schema-lk to go serial.

I'll be adding a test in a bit: there's code in another branch that should allow me to reproduce this pretty trivially.


adizaimi:
This might be the issue with clustered tests that i see at this point.

adizaimi:
Looks good. Thanks Mark.

markhannum:
Awesome- reviewed also w Lingzhi & Mike.

markhannum:
merge-ed

PR #623, opened 2017-12-29
Dbthread join recover deadlock

markhannum:
recover_deadlock in dbthread_join


lua/sp.c
```diff
@@ -3227,6 +3227,14 @@ static int dbthread_join(Lua lua1)
         struct timespec ts;
         clock_gettime(CLOCK_REALTIME, &ts);
         ts.tv_sec += 1;
+        if (bdb_lock_desired(thedb->bdb_env)) {
+            int rc;
+            if ((rc = recover_deadlock(thedb->bdb_env, lt->sp->thd->sqlthd, NULL, 0)) !=
+                    0) {
+                luabb_error(lua1, lt->sp, "recover deadlock failed");
+                return -3;
```
adizaimi: Can we look into why recover_deadlock could fail here, and what is best to do when it does?

markhannum: Discussed with Akshat- he is going to take care of this ..

akshatsikarwar: I will "take care" of this.

markhannum:
Handing to Akshat.  He is going to change the behavior to force the lua program to stop with a failure condition if recover_deadlock fails.

PR #626, opened 2018-01-02
Sc resume race

ldennis:


adizaimi:
This looks good. Thanks for fixing.

PR #631, opened 2018-01-04
Port lock-ordering to R7

markhannum:
Modify order in which we acquire locks in sqlengine_prepare: on upgrade we hold the bdb-lock in write mode, and then attempt to acquire the views-lock in write mode from newmaster->timepart_resume_schema_change.  If we want to get the views-lock and bdb-lock in sqlengine_prepare, we must get them in the same order.


markhannum:
Dorin reviewed & approved the changes under the R6 branch.

PR #634, opened 2018-01-05
fix timepart: reload table version on replicants

ldennis:
fix timepart: reload table version on replicants
This should fix sc_timepart and simple_timepart clustered tests

### DIFF
--- a/db/views.c
+++ b/db/views.c
@@ -648,6 +648,7 @@ int views_handle_replicant_reload(const char *name)
             view = NULL;
             goto done;
         }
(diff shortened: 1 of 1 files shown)


adizaimi:
looks good. not sure if info level will spew too much but will see from tests.

ldennis:
I wanted to be able to debug failing tests easier. But I can change it back.

adizaimi:
This seems to fix the timepart testcases in clustered. 
Thanks Lingzhi.

PR #640, opened 2018-01-09
Remove redundant sbuf2setttimeout call

mohitkhullar:
We call sbuf2setttimeout and set it to max idle time, just after this call


mohitkhullar:
Thanks

PR #642, opened 2018-01-10
Move test execs to test specific directory

adizaimi:
In this PR we move comdb2 and cdb2sql executables to the test specific directory:
* Now we will be able to run two tests (clustered) simultaneously (say from two different terminal windows). Note that we make only one copy of the two executables for all the tests running from make test command.
* Without this change, issuing a second make test will fail because the comdb2 executable is in use and is not overwritable. 
* This also allows us to retain the executable with the test dir for further investigating issues.


tests/testcase.mk
```diff
@@ -35,17 +35,13 @@ export CDB2_OPTIONS=--cdb2cfg $(CDB2_CONFIG)
...
-export CDB2SQL_EXE?=$(SRCHOME)/cdb2sql
-export COPYCOMDB2_EXE?=$(SRCHOME)/db/copycomdb2
-export CDB2_SQLREPLAY_EXE?=$(SRCHOME)/cdb2_sqlreplay
-export PMUX_EXE?=$(SRCHOME)/pmux
-
-ifeq ($(COMDB2MD5SUM),)
-  # record md5sum so we can verify from setup of each individual test
-  export COMDB2MD5SUM:=$(shell md5sum ${COMDB2_EXE} | cut -d ' ' -f1)
-endif
+#also defined in Makefile -- needed here because we can run tests from .test/
```
nirbhayc: It should be possible to move them into a separate file.
```
include execs.mk
```

tests/testcase.mk
```diff
@@ -35,17 +35,13 @@ export CDB2_OPTIONS=--cdb2cfg $(CDB2_CONFIG)
...
+export COMDB2_EXE?=$(TESTDIR)/comdb2
+export COMDB2AR_EXE?=$(TESTDIR)/comdb2ar
+export CDB2SQL_EXE?=$(TESTDIR)/cdb2sql
+export COPYCOMDB2_EXE?=$(TESTDIR)/db/copycomdb2
+export CDB2_SQLREPLAY_EXE?=$(TESTDIR)/cdb2_sqlreplay
```
nirbhayc: Where is copycomdb2 & cdb2_sqlreplay being copied? executables target above doesn't seem to copy them.

tests/testcase.mk
```diff
@@ -35,17 +35,13 @@ export CDB2_OPTIONS=--cdb2cfg $(CDB2_CONFIG)
...
+export PMUX_EXE?=$(SRCDIR)/pmux
```
nirbhayc: Why leave it in SRCDIR?

tests/runtestcase
```diff
@@ -3,7 +3,7 @@ TEST_TIMEOUT=${TEST_TIMEOUT:-5m}
 SETUP_TIMEOUT=${SETUP_TIMEOUT:-2m}
 
 export CLEANUPDBDIR=${CLEANUPDBDIR:-1}
-export PATH="${SRCHOME}/:${PATH}"
+export PATH="${TESTDIR}/:${PATH}"
```
nirbhayc: I see that we are exporting PATH at a couple of places. How about doing it in testcase.mk (top-level) after we have generated a random test directory name?

adizaimi: putting it in testcase.mk gave error about recursive referencing PATH, so make does not like it there.

tests/testcase.mk
adizaimi: Yep, needs fixing.

tests/testcase.mk
adizaimi: srcdir is incorrect, fixing to srchome. 
Only comdb2, cdb2sql and comdb2ar are needed in the testdir, the others are fine in the srchome.

PR #643, opened 2018-01-10
Fix userschema tests for cluster

mohitkhullar:
Add user at the time of setting authentication

### DIFF
--- a/sqlite/comdb2build.c
+++ b/sqlite/comdb2build.c
@@ -30,6 +30,7 @@
 extern pthread_key_t query_info_key;
 extern int gbl_commit_sleep;
 extern int gbl_convert_sleep;
+extern int gbl_check_access_controls;
 /******************* Utility ****************************/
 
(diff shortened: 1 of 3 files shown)


sqlite/comdb2build.c
```diff
@@ -1425,6 +1426,7 @@ void comdb2enableAuth(Parse* pParse, int on)
     arg->auth = auth;
     arg->type = BPFUNC_AUTHENTICATION;
     auth->enabled = on;
+    gbl_check_access_controls = 1;
```
ldennis: what if we fail on master?

mohitkhullar: This is just to check do we need access or not. it will be reset to 0 after next query.
https://github.com/bloomberg/comdb2/blob/master/db/sqlinterfaces.c#L5743

mohitkhullar:
thanks

PR #646, opened 2018-01-11
Fix for bad comdb2 dbinfo parsing, affects connection to local

adizaimi:
* Connection to comdb2 local is broken because we rely on node number to determine which node is the master. This PR fixes that by strcmp the node name
* further refactor some functions in cdb2api and add more verbose trace
* turn to milliseconds delay switches (rather than microseconds).


mohitkhullar:
looks good,

PR #648, opened 2018-01-16
fix clustered test: analyze and pg_free_recovery

ldennis:
- fix analyze cluster test
    - analyze doesnt have to get schema_lk, use ATOMIC_ADD on gbl_analyze_gen instead
    - call memp_sync on flush on replicants

- disable parallel rep for pg_free_recovery.test


adizaimi:
ll_checkpoint() on a replicant is a noop, thus flush or replicant did nothing. 
This looks good. Thanks Lingzhi.

PR #649, opened 2018-01-17
DDL: use 'tag ondisk' in the generated csc2

nirbhayc:
This is required to preserve the backwards compatibility
of the generated csc2 with older comdb2 versions.


akshatsikarwar:
This is better done under 'legacy_tunables`. No need to update testcases and will work properly when deployed internally.

akshatsikarwar:
And stock R7 can keep using better names :)

nirbhayc:
It was decided that the r6 csc2 parser will be updated to accept "schema".

PR #652, opened 2018-01-18
remove berk atomic.h and use comdb2_atomic.h

adizaimi:
We have multiple occurences of atomic operations, so I am removing one of them, berkeley atomic.h and dependences on it, in favor of comdb2_atomic.h


adizaimi:
Thank you Mike.

PR #653, opened 2018-01-18
Fix random 'Error in sending back data' from server

mohitkhullar:
Server would randomly give error: 'Error in sending back data' on sp trace/debug


mohitkhullar:
Thanks Lingzhi

nirbhayc:
https://gist.github.com/nirbhayc/286de66f15bde2dedcb8c5dae9fea989

PR #654, opened 2018-01-18
Block on active processor threads

markhannum:
This is a series of changes which exposes and addresses a race condition.  Before this change, parallel-rep worker-threads could corrupt a database if they are allowed to persist during an upgrade.   The fix is to force parallel rep to go serial if the bdb writelock is desired, and to force the thread acquiring the bdb writelock to block until the number of processor threads has dropped to 0.  The halt_processor_tds test reproduces the corruption if the 'force_serial_writelock' tunable is turned off.


ldennis:
1308e761c2899b1ae352cb77297e7bb0a045b4a5 disabled parallel rep for pg_free_recovery.test. Can you try enabling parallel rep and if pg_free_recovery can pass clustered with your fix?

akshatsikarwar:
Should re-enable in this PR

markhannum:
No- doesn't fix

markhannum:
Doesn't look related to parllel-rep- it's failing because a node it's running a select against is running recovery.  I'm wondering if this test isn't too aggressive.

akshatsikarwar:
Disabling ||rep makes test pass though.

How would query run on node which is in recovery?

markhannum:
Test forces a request against every node in the cluster & fails if one of them doesn't work

akshatsikarwar:
Server was crashing in rep worker thread.

markhannum:
The abort after dispatching?  Yeah- that's one of the ways I was able to get this to break before these changes..

ldennis:
I tested your branch and it passed pg_free_recovery in clustered mode. (I ran the test with 6 nodes on internal btda machines.) Can you rebase and re-enable parallel rep for the test?

akshatsikarwar:
Whats up with the PR? Why so many unrelated commits?

markhannum:
I think this fixes what you guys were seeing.  It still fails, but it's getting watchdog'd because of the artificial latency that we've added in the pgfree recovery handler.  Yes, I'll rebase & reenable.

ldennis:
it passed for me with `make pg_free_recovery CLUSTER='pnj-comdb2btda1 pnj-comdb2btda2 pnj-comdb2btda3 pno-comdb2btda2 pno-comdb2btda3 pno-comdb2btda4'`

PR #655, opened 2018-01-18
Az refactor more

adizaimi:
* Refactor newsql_connect to not have confusing goto jump
* error check @host capability when port keyword is missing
* return error when can't parse the @host line
*  add test for --host node:port which is supported


markhannum:
Looks fine- okay to merge

adizaimi:
Thank you Mark.

PR #661, opened 2018-01-23
Test: Fix cdb2api.test

nirbhayc:
ORDER BY over datetime dbstore="CURRENT_TIMESTAMP" column could return rows in different order.


ldennis:
`select i from t1 order by d1, i desc` instead? so we can at least check the datetime "CURRENT_TIMESTAMP" is non-decreasing

nirbhayc:
> select i from t1 order by d1, i desc instead? so we can at least check the datetime "CURRENT_TIMESTAMP" is non-decreasing

Point taken!

nirbhayc:
Thanks @ldennis

PR #662, opened 2018-01-23
Cdb2 setinfo clean

markhannum:
Send some additional information to the server: the client's argv0, and the calling backtrace of the thread which called 'cdb2_open'.


markhannum:
don't understand why there are 19 commits instead of 1 .. maybe i need to re-fork?

markhannum:
Yup- you're right .. we should add to jdbc

markhannum:
Yup- you're right .. we should add this to jdbc

adizaimi:
Is there argv0 for orange?

markhannum:
Reviewed with Mike today.  Fixed plenty- this should be okay to merge now.

markhannum:
It was embarrassing that the review was as useful as it was ..

PR #663, opened 2018-01-25
cdb2api unit test

adizaimi:
This checkin adds a unit test for cdb2api.
It currently only checks a small number of the functions in cdb2api for behavior, much more work is needed to have the other many functions checked. 
It is a starting point where we can add incrementally tests independently for each function.


cdb2api/cdb2api_priv.h
```diff
@@ -0,0 +1,141 @@
+/*
```
riverszhang89: You don't need to do this. Just `#include "cdb2api.c"` in your unit tests.

adizaimi: Well, the point is to stub out functions that any given function needs. If I include cdb2api.c, there would be duplicate function defined.
Coincidentally, in this case (get_comdb2db_hosts) I dont use hndl so i could have done without this change, but other functions will need to verify variables in it.

mohitkhullar: Can't we have whatever is needed for unit tests in the actual cdb2api.c functions #ifdefs?

PR #664, opened 2018-01-25
fix print statement

markhannum:
Fix a simple print-format error


markhannum:
ADI SAYS THIS IS COOL

akshatsikarwar:
IS IT THOUGH

PR #665, opened 2018-01-25
Destroy durable lsn mutex in __dbenv_close instead of __dbenv_refresh

mohitkhullar:
__dbenv_refresh is called in __dbenv_open before init mutex is done


mohitkhullar:
thanks

PR #666, opened 2018-01-25
Pmux close unix socket

markhannum:
Fixes to pmux designed to address security issues found in SDSK 843638697 (an internal Bloomberg ticket).


markhannum:
Thanks Nirbhay!

PR #667, opened 2018-01-25
pmux rte fix

adizaimi:
rte request will cause fd to be unwatched and next request can incorrectly access memory causing a pmux crash. This pr fixes that by checking for closed fd.


markhannum:
Reviewed with Adi yesterday - okay to merge

PR #669, opened 2018-01-26
cdb2api:have a function that only reads config

adizaimi:
Instead of calling get_comdb2db_hosts() with all null parameters, just call new function that loads config only_read_config(). This way our intention is clear, but also we decrease chance of a bug in case we modify the former.

### DIFF
--- a/cdb2api/cdb2api.c
+++ b/cdb2api/cdb2api.c
@@ -966,7 +966,7 @@ static int get_comdb2db_hosts(cdb2_hndl_tp *hndl, char comdb2db_hosts[][64],
                 __LINE__);
     }
 
-    if (get_config_file(dbname, filename, sizeof(filename)) != 0)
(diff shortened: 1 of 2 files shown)


cdb2api/cdb2api.c
```diff
@@ -4467,6 +4467,17 @@ static int cdb2_dbinfo_query(cdb2_hndl_tp *hndl, char *type, char *dbname,
     return -1;
 }
 
+static inline void only_read_config()
+{
+    FILE * fp = fopen(CDB2DBCONFIG_NOBBENV, "r");
+    if (fp != NULL) {
+        read_comdb2db_cfg(NULL, fp, NULL, NULL, NULL,
```
mohitkhullar: this isn't falling back to /bb/bin or other location of comdb2db.cfg..

adizaimi: you are right, i was not thinking about  those other locations.

PR #671, opened 2018-01-26
check and return if query is empty

adizaimi:
In this checkin read_newsql_query() checks that either dbinfo or sqlquery is set for a client query. 
This ensures that the db wont crash on bad input that results in sqlquery to be null.
Added testcase with bad input that previously crashed db.

### DIFF
--- a/plugin/newsql/newsql.c
+++ b/plugin/newsql/newsql.c
@@ -583,13 +583,13 @@ retry_read:
         fdb_heartbeats(clnt);
         pthread_mutex_unlock(&clnt->wait_mutex);
     }
+    free(p);
 
     if (pre_enabled) {
         pthread_mutex_lock(&clnt->wait_mutex);
         clnt->ready_for_heartbeats = 0;
         pthread_mutex_unlock(&clnt->wait_mutex);
     }
-    free(p);
 
     if (query && query->dbinfo) {
         if (query->dbinfo->has_want_effects &&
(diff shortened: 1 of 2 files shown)


mohitkhullar:
can it be validated in read_newsql_query function?

adizaimi:
Possibly. I'll take a closer look. Thanks.

plugin/newsql/newsql.c
```diff
@@ -583,13 +583,13 @@ static CDB2QUERY *read_newsql_query(struct dbenv *dbenv,
         fdb_heartbeats(clnt);
         pthread_mutex_unlock(&clnt->wait_mutex);
     }
+    free(p);
```
nirbhayc: I’d check errno here and not worry about checking if(query).

plugin/newsql/newsql.c
```diff
@@ -627,6 +627,11 @@ static CDB2QUERY *read_newsql_query(struct dbenv *dbenv,
         cdb2__query__free_unpacked(query, &pb_alloc);
         goto retry_read;
     }
+    else if (query && !query->sqlquery) {
```
nirbhayc: And now this becomes if(unlikely(!query->sqlquery)).

tests/basic.test/runit
```diff
@@ -241,6 +241,9 @@ for node in $cluster ; do
     res=`cdb2sql --tabs $dbnm --host $node:$port "select comdb2_dbname()"`
     assertres $res $dbnm
 
+    #check for empty sql stmt
+    echo -e "newsql\n\x0\x0\x0\x1\x0\x0\x0\x0\x0\x0\x0\x0\x0\x0\x0\n\x0\x0\x0\x1\x0\x0\x0\x0\x0" | nc -w 1 $node $port
+
```
nirbhayc: A test for empty query.
> cdb2sql <opts> “”

PR #673, opened 2018-01-30
Block on active processor threads rb

markhannum:
This is a series of changes which exposes and addresses a race condition. Before this change, parallel-rep worker-threads could corrupt a database if they are allowed to persist during an upgrade. The fix is to force parallel rep to go serial if the bdb writelock is desired, and to force the thread acquiring the bdb writelock to block until the number of processor threads has dropped to 0. The halt_processor_tds test reproduces the corruption if the 'force_serial_writelock' tunable is turned off.


bdb/tran.c
```diff
@@ -1567,6 +1568,7 @@ static int bdb_tran_commit_with_seqnum_int_int(
                 outrc = -1;
                 atexit(abort_at_exit);
                 failed_to_log = 1;
+                bdb_osql_trn_repo_unlock();
```
mponomar: This is definitely right, though looks like a different problem.  What problem was this?

berkdb/rep/rep_method.c
```diff
@@ -1539,42 +1539,7 @@ __rep_lockout(dbenv, db_rep, rep, msg_th)
...
-		__os_sleep(dbenv, 1, 0);
-#ifdef DIAGNOSTIC
-		if (++wait_cnt % 60 == 0)
-                      __db_err(dbenv,
-"Waiting for handle count to run replication recovery/backup for %d minutes",
-			    wait_cnt / 60);
-#endif
-		MUTEX_LOCK(dbenv, db_rep->rep_mutexp);
-	}
+	return;
```
mponomar: Only called in one place, may as well remove.

bdb/tran.c
markhannum: Didn't occur during testing- I just noticed it was broken.

berkdb/rep/rep_record.c
```diff
@@ -654,6 +658,10 @@ __rep_process_message(dbenv, control, rec, eidp, ret_lsnp, commit_gen)
 			MUTEX_UNLOCK(dbenv, db_rep->rep_mutexp);
 		} else if (rp->rectype != REP_NEWMASTER) {
 
+			if (gbl_verbose_master_req) {
+				logmsg(LOGMSG_ERROR, "%s line %d sending REP_MASTER_REQ\n",
```
mponomar: Should be LOGMSG_USER (user requested this to be verbose, should always be emitted).  Minor nitpick.

berkdb/rep/rep_record.c
```diff
@@ -905,6 +918,12 @@ send:			if (__rep_send_message(dbenv,
 								   commit_gen)) != 0)
 				goto errlock;
 		} else {
+
+			if (gbl_verbose_master_req) {
+				logmsg(LOGMSG_ERROR, "%s line %d sending REP_MASTER_REQ\n",
```
mponomar: LOGMSG_USER

berkdb/rep/rep_record.c
```diff
@@ -2573,8 +2595,14 @@ gap_check:		max_lsn_dbtp = NULL;
 				    NULL);
 			} else {
 				MUTEX_UNLOCK(dbenv, db_rep->rep_mutexp);
+
+				if (gbl_verbose_master_req) {
+					logmsg(LOGMSG_ERROR, "%s line %d sending REP_MASTER_REQ\n",
```
mponomar: LOGMSG_USER

berkdb/rep/rep_record.c
```diff
@@ -2988,6 +3016,10 @@ logical_record_file_affinity(int rectype)
 
 #include <stdlib.h>
 
+int gbl_processor_thd_poll;
+
+/* On second thought i think this is wrong if i have the order correct */
```
mponomar: What's wrong?

markhannum:
Thanks Mike!

PR #674, opened 2018-01-30
cleanup freeing code and ensure sql has whitespace stripped out

adizaimi:
* cleanup freeing code in newsql_write_response
* dont clean whitespace unneccessarily of the sql string since was
  already done once in dispatch request

### DIFF
--- a/db/sqlinterfaces.c
+++ b/db/sqlinterfaces.c
@@ -778,7 +778,7 @@ done:
         exit(1);
     }
 
-    if (dta && alloc == malloc)
+    if (dta)
         free(dta);
 
     return rc;
--- a/db/sqlinterfaces.c
+++ b/db/sqlinterfaces.c
@@ -702,12 +702,9 @@ int newsql_write_response(struct sqlclntstate *clnt, int type,
                           int line)
 {
     struct newsqlheader hdr;
-    int rc;
-    SBUF2 *sb;
-    int len;
-    void *dta;
-
-    sb = clnt->sb;
+    int rc = 0;
+    int len = 0;
+    void *dta = NULL;
 
     if (gbl_dump_fsql_response) {
         char cnonce[256] = {0};
@@ -740,9 +737,6 @@ int newsql_write_response(struct sqlclntstate *clnt, int type,
         len = cdb2__sqlresponse__get_packed_size(sql_response);
         dta = (*alloc)(len + 1);
         cdb2__sqlresponse__pack(sql_response, dta);
-    } else {
-        len = 0;
-        dta = NULL;
     }
 
     /* header */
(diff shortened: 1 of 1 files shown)


db/sqlinterfaces.c
```diff
@@ -752,57 +746,42 @@ int newsql_write_response(struct sqlclntstate *clnt, int type,
...
+    if (flush)
         sbuf2flush(sb);
-    }
 
-    rc = pthread_mutex_unlock(&clnt->write_lock);
-    if (rc != 0) {
-        logmsg(LOGMSG_FATAL, "couldnt get clnt->write_lock\n");
+done:
+    if (pthread_mutex_unlock(&clnt->write_lock) != 0) {
+        logmsg(LOGMSG_FATAL, "couldnt put clnt->write_lock\n");
```
nirbhayc: "Failed to unlock clnt->write_lock"

db/sqlinterfaces.c
```diff
@@ -3319,49 +3294,48 @@ int release_locks_on_emit_row(struct sqlthdstate *thd,
...
         }
     }
-    len = 6; // strlen "pragma"
+    len = sizeof("pragma") - 1;
     if (strncasecmp(sql, "pragma", len) == 0) {
         sql += len;
-        if (!isspace(*sql)) {
-            return 0;
+        if (isspace(*sql)) {
+            goto error;
```
nirbhayc: Why?

db/sqlinterfaces.c
```diff
@@ -752,57 +746,42 @@ int newsql_write_response(struct sqlclntstate *clnt, int type,
...
         sbuf2flush(sb);
-    }
 
-    rc = pthread_mutex_unlock(&clnt->write_lock);
-    if (rc != 0) {
-        logmsg(LOGMSG_FATAL, "couldnt get clnt->write_lock\n");
+done:
+    if (pthread_mutex_unlock(&clnt->write_lock) != 0) {
+        logmsg(LOGMSG_FATAL, "couldnt put clnt->write_lock\n");
         exit(1);
```
nirbhayc: UNRELATED: Just a thought, logmsg() should trigger exit(1)/abort() on LOGMSG_FATAL.

db/sqlinterfaces.c
```diff
@@ -1699,9 +1677,6 @@ static void sql_update_usertran_state(struct sqlclntstate *clnt)
     if (!sql)
         return;
 
-    while (isspace(*sql))
-        ++sql;
-
```
nirbhayc: UNRELATED: I strongly feel, we should offload all the processing of user commands to the parser instead.

nirbhayc: Did you test this change?

db/sqlinterfaces.c
adizaimi: Hmm, that does not sound bad. I wonder if there is any circumstance where we don't exit after logmsg_fatal.

db/sqlinterfaces.c
adizaimi: We make copy of the sql string and use that for hash lookup and so on. Since this only cleans up the leading whitespace this is not too bad, however any difference in spacing inside the query will not match (thus we may need to s/  / /g and that is more of a parser's job).

adizaimi: 1697     const char *sql = clnt->sql;
We are imposing clnt->sql does not contain leading whitespace from all paths.

adizaimi:
thank you Nirhbay.

db/sqlinterfaces.c
nirbhayc: > I wonder if there is any circumstance where we don't exit after logmsg_fatal.

In such cases it shouldn't be fatal.

PR #676, opened 2018-01-30
Initialize lastuser

mohitkhullar:


db/sqlinterfaces.c
```diff
@@ -6153,6 +6153,7 @@ void sqlengine_thd_start(struct thdpool *pool, struct sqlthdstate *thd,
     thd->offsets = NULL;
     thd->sqldb = NULL;
     thd->stmt_caching_table = NULL;
+    bzero(thd->lastuser, sizeof(thd->lastuser));
```
nirbhayc: Better to use memset().

adizaimi:
I'm afraid i fixed this unknowingly in a more recent checkin.

PR #677, opened 2018-01-30
Timepart systables and fixes

dorinhogea:
1) Systables to expose time partition information:
'select * from comdb2_timepartitions':
(name='tv', period='daily', retention=4, nshards=4, version=0, shard0name='t', starttime=1516856400, source_id='31438f5e-dd6d-4ccf-aff2-c193221027fe')
...
'select shardname, cast(start as datetime) as start, cast(end as datetime) as end from comdb2_timepartshards where name="tv"':
(shardname='$1_E5EE49E6', start="2018-01-30T000000.000 America/New_York", end="2038-01-18T221407.000 America/New_York")
....
2) port from R6: multiple master swings during rollout generate multiple events

### DIFF
--- a/db/views.c
+++ b/db/views.c
@@ -2636,10 +2636,10 @@ static int _view_update_table_version(timepart_view_t *view, tran_type *tran)
 }
 
 /**
- * Returned a malloced string for the "iRowid"-th view, column iCol 
+ * Returned a malloced string for the "iRowid"-th timepartition, column iCol 
  * NOTE: this is called with a read lock in views structure
  */
-void timepart_systable_get_column(sqlite3_context *ctx, int iRowid, enum systable_columns iCol)
+void timepart_systable_column(sqlite3_context *ctx, int iRowid, enum systable_columns iCol)
 {
    timepart_views_t  *views = thedb->timepart_views;
    timepart_view_t   *view;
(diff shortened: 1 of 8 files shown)


db/views.h
```diff
@@ -59,6 +59,24 @@ typedef struct timepart_sc_arg {
     void *tran; /*remove?*/
 } timepart_sc_arg_t;
 
+
+enum systable_columns {
+    VIEWS_NAME
+    ,VIEWS_PERIOD
+    ,VIEWS_SHARDNAME=VIEWS_PERIOD
```
nirbhayc: All this to save one extra enum definition? :) Please separate them.

db/views.c
```diff
@@ -2631,25 +2635,123 @@ static int _view_update_table_version(timepart_view_t *view, tran_type *tran)
     return rc;
 }
 
-#if 0
-static int bdb_lock_table_write_pp(const char *tblname, timepart_sc_arg_t *arg)
+/**
```
nirbhayc: Any specific reason for having these systbl changes done here than in sqlite/ext/comdb2/timepartitions.c.

db/views.h
nirbhayc: and move them to timepartitions.c.

db/views.c
dorinhogea: Abstraction isolation: I do not want to expose time partition implementation to sqlite library.  
I am also not thrilled about the location of comdb2 sqlite extension modules; most of these are close users of stuff living in db/ folder.

db/views.h
dorinhogea: See above.

dorinhogea: Test coming.

PR #678, opened 2018-01-30
Incremental copy

mponomar:
Copy only the pages that changed, when copying on top of an existing image.

### DIFF
--- a/db/copycomdb2
+++ b/db/copycomdb2
@@ -141,6 +141,10 @@ if [[ ! -x $comdb2ar ]]; then
    comdb2ar=$(which comdb2ar 2>/dev/null)
 fi
 
+if [[ -z "$comdb2ar" ]]; then
+   comdb2ar=/home/mponomar/comdb2/comdb2ar
+fi
+
 if [[ ! -x $lz4 ]]; then
    lz4=$(which lz4 2>/dev/null)
    if [[ -z "$lz4" ]]; then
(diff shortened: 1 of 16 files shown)


tools/comdb2ar/comdb2ar.cpp
```diff
@@ -236,6 +241,14 @@ int main(int argc, char *argv[])
                 mode = CREATE_MODE;
                 break;
 
+            case 'p':
+                mode = PARTIAL_CREATE_MODE;
+                break;
+
+            case 'P':
```
nirbhayc: p & P could be confusing. I'd go with 'i' for incremental create and 'r' restore.

mponomar: We already have -I.  I was trying to avoid confusion there.

db/copycomdb2
```diff
@@ -139,14 +141,18 @@ if [[ ! -x $comdb2ar ]]; then
    comdb2ar=$(which comdb2ar 2>/dev/null)
 fi
 
+if [[ -z "$comdb2ar" ]]; then
```
mponomar: I sure hope this fails review :)

tools/comdb2ar/comdb2ar.cpp
mponomar: FWIW, I don't expect comdb2ar to be used outside of copycomdb2

markhannum:
This is a really fantastic idea .. it'll allow us to reduce the size of a copy to the current logical differences between the source and the destination.  Wow!

PR #679, opened 2018-01-30
cdb2api: auto consume remaining rows with timeouts before cdb2_close

ldennis:
cdb2api: auto consume remaining rows with timeouts before cdb2_close

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,7 @@ tests/tools/register
 tests/tools/verify_atomics_work
 tests/tools/cdb2_open
 tests/tools/cdb2api_unit
+tests/tools/cdb2_close_early
 
 
 tests/test_*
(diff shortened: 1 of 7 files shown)


mohitkhullar:
The code looks fine, can we have a test case where we run a long query and do cdb2_close after reading just 1 row, just so that it triggers this code logic.

mohitkhullar:
looks good

riverszhang89:
What problem does the pull request address?

riverszhang89:
Cdb2jdbc consumes all rows on close. The patch does not need to be ported to cdb2jdbc.

akshatsikarwar:
For jdbc - it doesn't really have to - there is no connection pool.

ldennis:
Thanks.

PR #680, opened 2018-01-31
PORTED: system table for clients stats

ldennis:
PORTED: system table for clients stats

### DIFF
--- a/db/reqlog.c
+++ b/db/reqlog.c
@@ -2276,7 +2276,7 @@ void process_nodestats(void)
     cpl.i = 0;
     cpl.n = nclnts;
     hash_for(clientstats, collect_clientstats, &cpl);
-    for (i == 0; i < nclnts; i++) {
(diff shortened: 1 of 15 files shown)


db/reqlog.c
```diff
@@ -1990,88 +1988,333 @@ void reqlog_set_truncate(int val)
...
+    list = calloc(nclnts, sizeof(nodestats_t *));
+    if (list == NULL) {
+        logmsg(LOGMSG_ERROR, "%s: no memory\n", __func__);
+        goto done;
+    }
+    cpl.list = list;
+    cpl.i = 0;
+    cpl.n = nclnts;
+    hash_for(clientstats, collect_clientstats, &cpl);
+    for (i == 0; i < nclnts; i++) {
```
markhannum: i = 0

markhannum:
We've discussed ways to make the interface more intuitive, but I think we'll do that from a different PR.  Approved.

PR #681, opened 2018-01-31
Throttle log-puts against incoherent nodes

markhannum:
Current log-traffic can prevent re-transmission requests from being queued to a node which has fallen behind.  This revised version of throttle_updates_to_incoherent_nodes addresses this by only allowing log-puts to incoherent nodes which are within 1 mb of the master.

### DIFF
--- a/net/net.c
+++ b/net/net.c
@@ -2021,6 +2021,51 @@ int net_get_stack_flush_threshold(void) { return stack_flush_min; }
 
 void net_set_stack_flush_threshold(int thresh) { stack_flush_min = thresh; }
 
+int gbl_dump_full_net_queue = 0;
+
+static void dump_queue(netinfo_type *netinfo_ptr, host_node_type *host_node_ptr)
+{
+    int now, cnt = 0, logput_cnt = 0, non_logput_cnt = 0;
+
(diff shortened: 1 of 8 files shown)


markhannum:
This branch adds three new tunables.  'time_rep_apply' periodically prints timing information for the __rep_apply thread.  'logput_window' allows fine-grained control over how far behind an incoherent node will start receiving normal log-traffic.  'dump_full_netqueue' will dump the logput LSNs of a host's netqueue should a net-send fail with a 'queue-full' (-8) error.

net/net.c
```diff
@@ -2021,6 +2021,51 @@ int net_get_stack_flush_threshold(void) { return stack_flush_min; }
 
 void net_set_stack_flush_threshold(int thresh) { stack_flush_min = thresh; }
 
+int gbl_dump_full_net_queue = 0;
+
+static void dump_queue(netinfo_type *netinfo_ptr, host_node_type *host_node_ptr)
+{
+    int now, cnt = 0, logput_cnt = 0, non_logput_cnt = 0;
+
+    if (netinfo_ptr->netcmp_rtn == NULL)
```
markhannum: Should be getlsn_rtn

markhannum:
Approved by Akshat!

PR #682, opened 2018-02-01
Inline functions that are called a lot to gain performance

adizaimi:
Running basic.test in callgrind, shows that functions such as getdatsize(), is_raw(), getdata(), is_remote(), get called hundreds of times for each query (every sqlite3step) and as such it would be much more efficient to inline them.


akshatsikarwar:
How much difference did this make?

adizaimi:
I think optimization needs to be enabled to see the effects of this, however judging from the profiler percentages, each of these functions weighs about 0.1% of cpu -- any savings can't be that much. [While looking for something else,] I noticed these functions were called 270k times for 3400 sqlite_step(), and that they were not inlined, and thought this seems useful (at least harmless).

dorinhogea:
What was the optimization mode you tested in callgrind initially?

akshatsikarwar:
I'm fine with everything except `db/sqlglue_vdbe.h`

dorinhogea:
If we test this without optimization, it will be misleading.  Compiler should inline some calls already.

adizaimi:
Dorin, agreed. However some of these functions (is_raw, is_remote) are cross library so compiler can not possibly inline them as it needs to export the symbols.

adizaimi:
yep, you are right. pushing fix.

adizaimi:
Thank you Akshat.

PR #684, opened 2018-02-03
minor cleanup of the operations done to delete files after sc

adizaimi:
Remove redundancy is several of the steps involved to cleanup unneeded files
Don't call create_sqlmaster_records() multiple times


dorinhogea:
The slow operations are already offloaded.  Currently, the inline portion only fill up a linked list.  I doubt remove this will provide meaningful benefits.

PR #686, opened 2018-02-05
Set the sbuf timeout before SSL negotiation

markhannum:
Simple one-line fix addresses timeouts for the cinsert linearizable test.


markhannum:
Approved by Akshat

PR #688, opened 2018-02-05
check for cursor class tmptable for authentication access

adizaimi:
Authenticating a tmptable will result in a crash. Avoid this by checking if cursor is referencing a temp table.


adizaimi:
Thank you Mohit.

db/osqlsqlthr.c
```diff
@@ -1642,6 +1642,9 @@ int access_control_check_sql_read(struct BtCursor *pCur, struct sql_thread *thd)
     int rc = 0;
     int bdberr = 0;
 
+    if (pCur->cursor_class == CURSORCLASS_TEMPTABLE)
+        return 0;
+
```
nirbhayc: Does it mean that an unprivileged user can access comdb2_users and other vtabs? Could you please add a comment.

adizaimi: No, this simply avoids the crash which occurs few lines later by accessing pCur->db which is null.
This does not change access to comdb2_users and other vtabs.

PR #689, opened 2018-02-05
return error on incorrectly quoted string for user and password

adizaimi:
return error on incorrectly quoted string for user and password


mohitkhullar:
Looks good, you can merge it after fixing the coding style error.

PR #690, opened 2018-02-05
hashdb for less memcmp

adizaimi:
When db starts up, we traverse the list of files in the directory to assign a fileid to each of them.
The way this operation is implemented is innefficient by traversing a linked list over and over for each file.
This checkin optimizes that traversal by introducing a hash table and only performing the linked list traversal if fileid was found in hash. This saves about 1 Million calls to memcmp (on startup of a db with 5000 files in the data directory), however the time benefits are only in the order of tens of milliseconds.


berkdb/env/env_open.c
```diff
@@ -927,6 +931,10 @@ __dbenv_refresh(dbenv, orig_flags, rep_check)
 	 * log file handles.  Ick.
 	 */
 	LIST_INIT(&dbenv->dblist);
+    if (dbenv->fileidhash != NULL) {
+        hash_clear(dbenv->fileidhash);
```
ldennis: should we also call hash_free() ?

berkdb/db/db.c
```diff
@@ -475,8 +475,10 @@ __db_dbenv_setup(dbp, txn, fname, id, flags)
 
 	if (F_ISSET(dbp, DB_AM_HASH))
 		dbp->peer = dbp;
-	for (lldbp = LIST_FIRST(&dbenv->dblist);
-	    lldbp != NULL; lldbp = LIST_NEXT(lldbp, dblistlinks)) {
+    int foundinhash = hash_find(dbenv->fileidhash, dbp->fileid) != NULL;
```
ldennis: use tab for berkdb code

berkdb/env/env_open.c
adizaimi: hash_clear() seems to do almost same work as hash_free() so it should be enough.

dorinhogea: There are not the same; hash_clear() empty a hash (frees elements stored inside), while hash_free() does a hash_clear() followed by freeing the hash itself.

berkdb/db/db.c
```diff
@@ -475,8 +475,11 @@ __db_dbenv_setup(dbp, txn, fname, id, flags)
 
 	if (F_ISSET(dbp, DB_AM_HASH))
 		dbp->peer = dbp;
-	for (lldbp = LIST_FIRST(&dbenv->dblist);
-	    lldbp != NULL; lldbp = LIST_NEXT(lldbp, dblistlinks)) {
+
+	int foundinhash = hash_find_readonly(dbenv->fileidhash, dbp->fileid) != NULL;
+
+    for (lldbp = LIST_FIRST(&dbenv->dblist);
```
riverszhang89: tab.

berkdb/db/db.c
```diff
@@ -485,11 +488,11 @@ __db_dbenv_setup(dbp, txn, fname, id, flags)
 				break;
 			}
 		}
-	}
+    }
```
riverszhang89: tab

berkdb/db/db.c
```diff
@@ -518,6 +521,7 @@ __db_dbenv_setup(dbp, txn, fname, id, flags)
 		listc_init(&dbenv->dbs[dbp->adj_fileid], offsetof(DB, adjlnk));
 
 		LIST_INSERT_HEAD(&dbenv->dblist, dbp, dblistlinks);
+        hash_add(dbenv->fileidhash, dbp->fileid);
```
riverszhang89: tab

berkdb/env/env_open.c
```diff
@@ -927,6 +933,12 @@ __dbenv_refresh(dbenv, orig_flags, rep_check)
 	 * log file handles.  Ick.
 	 */
 	LIST_INIT(&dbenv->dblist);
+    if (dbenv->fileidhash != NULL) {
```
riverszhang89: tab

riverszhang89: Please use tabs to indent berkeleydb changes.

riverszhang89:
```
This saves about 1 Million calls to memcmp (on startup of a db with **5000** files in the data directory), however the time benefits are only in the order of **tens of milliseconds**.
```

With all due respect, this sounds a bit over-optimizing to me.

adizaimi:
Just did another comparison, starting a DB: master 3.540s vs 3.298s with this change, in a db with 3671 data/blob/index files in its directory. The difference in calls to memcmp is ~13.9M less with this change. While it is not any large gain in terms of time (I think we have lots of I/O and possibly), I wanted to eliminate the most called function which shows up in cachegrind.

riverszhang89:
> The difference in calls to memcmp is ~13.9M less with this change. 

`memcmp` will still be at the top no matter how you optimize  the setup routine. Keys are always memcmp'd - How would you plan to optimize that out?

adizaimi:
I'm not trying to eliminate memcmp, only to reduce calls to that func. With this change, we make about 7k hash_finds, thus a significant reduction of memcmp. The startup algorithm was probably not ment to deal with thousands of btrees in the db directory -- otherwise they would have implemented a less silly way to accomplish what it does.

adizaimi:
Closing

adizaimi:
moved to branch AZ_use_fileid_hash, might pick up later if I have a test that shows speedup.

PR #695, opened 2018-02-06
fix decimal index payload size for bdb_reconstruct_delete

ldennis:
fix decimal index payload size for bdb_reconstruct_delete


dorinhogea:
LGTM

ldennis:
Thanks

PR #696, opened 2018-02-06
check if query is null since it can be null on particular bad input

adizaimi:
check if query is null since it can be null on particular bad input


adizaimi:
thanks Nirbhay.

adizaimi:
ftr. the failure was during setup, failure to get pmux port 19000, not related to this checkin.

PR #697, opened 2018-02-06
Plugin: initializer plugin

nirbhayc:
* Introduced 'initializer' plugin type that can be used
  to initialize special sub-modules on server start
* Renamed some util functions to avoid collision with
  symbols in internal Bloomberg libraries
* Removed unnecessary/dead code


db/handle_buf.c
```diff
@@ -1273,12 +1197,23 @@ int handle_buf_main(struct dbenv *dbenv, struct ireq *iq, SBUF2 *sb,
...
     if (ndispatch == 0)
         nwaits++;
     return 0;
 }
 
+int handle_buf_main(struct dbenv *dbenv, struct ireq *iq, SBUF2 *sb,
+                    const uint8_t *p_buf, const uint8_t *p_buf_end, int debug,
+                    char *frommach, int frompid, char *fromtask,
+                    sorese_info_t *sorese, int qtype, void *data_hndl,
+                    int do_inline, int luxref, unsigned long long rqid)
```
akshatsikarwar: Keeping `inline`?

db/dbqueue.c
```diff
@@ -1197,7 +1194,7 @@ static void dbqueue_flush(struct dbtable *db, int consumern)
     logmsg(LOGMSG_INFO, "Beginning flush for queue '%s' consumer %d\n",
            db->tablename, consumern);
 
-    if (db->dbenv->master != machine()) {
+    if (db->dbenv->master != gbl_mynode) {
```
akshatsikarwar: Check if we need strcmp instead

db/handle_buf.c
```diff
@@ -709,22 +711,26 @@ static int reterr(struct thd *thd, struct ireq *iq, int do_inline, int rc)
...
-                    pool_relablk(p_reqs, iq);
-                }
+                pool_relablk(p_reqs, iq);
             }
         }
         UNLOCK(&lock);
     }
+    if (comdb2_ipc_sndbak) {
+#if 0
+       /* curswap is just a pointer to the buffer */
```
akshatsikarwar: Reqd?

db/dbqueue.c
nirbhayc: Old machine() simply returned gbl_mynode.
```
char *machine(void) { return gbl_mynode; }
```

PR #701, opened 2018-02-07
long long timeout for ddl transactions

dorinhogea:
Long DDL transactions (like rebuilds or alter) taking longer than 10 minutes fail with error, even though the schema changes finish successfully.
This patch provides a configurable timeout value for DDL transactions.
Ported from R6: https://bbgithub.dev.bloomberg.com/comdb2/comdb2/pull/797


dorinhogea:
thanks Adi (master still randomly fails localrep, so merging this)

adizaimi:
localrep was because of port being in use from pmux_hello test which i fixed just now.

PR #702, opened 2018-02-07
option to ignore corrupted tables during start up

dorinhogea:
Currently, any table that has corrupted metadata or missing disk files prevents a server from coming up. 
The patch adds an option for the server comes up by ignoring the broken table, allowing access to the other tables.  The missing table can be recreated and the metadata will be overwritten and fixed.
Ported from R6 https://bbgithub.dev.bloomberg.com/comdb2/comdb2/pull/811


nirbhayc:
@dorinhogea Could you also add a test for this new tunable?

dorinhogea:
@nirbhayc :testcase added

dorinhogea:
Thanks for the review

nirbhayc:
Thank you!

adizaimi:
FTR, this is the error in tests/62ebd7f950a2ba57816db7ad21e0d204065d147a/test_15970/logs/scindex15970.testcase:
21:54:17> ERROR: Verify failed.
21:54:17> Testcase failed.

PR #708, opened 2018-02-08
keep remsql communication protocol encapsulated; remove remcur

dorinhogea:
Moved fdb_comm.c that includes all communication protocol implementation inside plugin, and re-encapsulate message definition.
Remove unused remcur plugin.


dorinhogea:
Thanks for the review!

PR #713, opened 2018-02-09
Fix time partition versioning

dorinhogea:
comdb2_timepartitions->version is the same as underlying table persistent version following this check (and not the transient, schema change refreshing view counter).


dorinhogea:
simple_timepart finished ok; the regression engines malfunctions again.

PR #716, opened 2018-02-12
Use sys free for sys malloced mem

mohitkhullar:
merge from R6


adizaimi:
this looks good. Thanks.

PR #718, opened 2018-02-13
Fix net race

markhannum:
Fix bug in net: closed file-descriptors are set to -1, not to 0.


markhannum:
Reviewed with Mike!

markhannum:
Reviewed this with Mike

PR #727, opened 2018-02-15
dbname too long error to show actual name, fix tunables

adizaimi:
dbname too long error to show actual name, fix tunables


adizaimi:
thanks Nirbhay

PR #730, opened 2018-02-15
Only poll if we're not running recovery

markhannum:
We get watchdogged in rep-verify-match because of artificial latency that we've added to tease out an issue where berkley wouldn't grab page-locks in compensating transactions.  Straightforward fix: don't poll in this handler if we are in recovery mode.


markhannum:
Reviewed with Adi

Issue #726, opened 2018-02-15
macOS issue - libcdb2api.dylib missing dependency on openssl


mthorpe7:
```
$ otool -L /usr/local/opt/comdb2/lib/libcdb2api.dylib
/usr/local/opt/comdb2/lib/libcdb2api.dylib:
	/usr/local/opt/comdb2/lib/libcdb2api.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.0.0)
```

```
$ nm /usr/local/opt/comdb2/lib/libcdb2api.dylib | grep _SSL_library_init
                 U _SSL_library_init
```

I think something funky is going on here: https://github.com/bloomberg/comdb2/blob/master/cdb2api/CMakeLists.txt#L35-L37

mthorpe7:
@akshatsikarwar

akshatsikarwar:
Can you build without the funky line?

mthorpe7:
Had to tweak a few things, but I created #729

PR #731, opened 2018-02-16
Fix libcdb2api.dylib

akshatsikarwar:
Add `ssl` and `protobuf-c` dependency for `libcdb2api.dylib`
Don't need `ossp-uuid`; uuid is part of macOS and provided by `libSystem.dylib`


akshatsikarwar:
Should address issue: #726

mthorpe7:
👍

PR #735, opened 2018-02-16
DDL: Preserve user-defined tags post ALTER TABLE

nirbhayc:


akshatsikarwar:
don't check for column in tag 
remove magic nums(use xmacro)

nirbhayc:

PR #736, opened 2018-02-16
ssl for cross-database sql

dorinhogea:
If originator ssl session uses ssl, the remote sql is using ssl as well.  It relies and preconfigured server ssl keys to provide channel encryption, eliminating the need to exchange plain data in cross-database joins.
It does not serve as an access control mechanism.


db/fdb_fend.c
```diff
@@ -1192,7 +1211,7 @@ static int _failed_AddAndLockTable(sqlite3 *db, const char *dbname, int errcode,
     struct sql_thread *thd = pthread_getspecific(query_info_key);
     struct sqlclntstate *clnt = thd->sqlclntstate;
 
-    logmsg(LOGMSG_WARN, "%s: %s for db \"%s\"\n", __func__, prefix, dbname);
+    logmsg(LOGMSG_WARN, "Error \"%s\" for db \"%s\"\n", prefix, dbname);
```
adizaimi: Should this be logmsg_error?

db/fdb_fend.c
```diff
@@ -4718,24 +4789,35 @@ void _fdb_clear_clnt_node_affinities(struct sqlclntstate *clnt)
...
+    } else {
+        *flags = FDB_MSG_CURSOR_OPEN_SQL_SSL;
+#if WITH_SSL
+        if (sslio_has_ssl(clnt->sb) || fdb->ssl >= SSL_REQUIRE) {
+            *flags |= FDB_MSG_CURSOR_OPEN_FLG_SSL;
+        }
+#endif
     }
+
+    /*fprintf(stderr, "%s: return flags=%d sb=%p has_ssl=%d\n", __func__,
```
adizaimi: Can have this as logmsg_debug

adizaimi:
Looks ok to me.

riverszhang89:
Suppose 

1) the client certificate is trusted by the server; 

2) the server certificate is trusted by the remote; 

3) the client certificate is *NOT* trusted by the remote; 

4) and the server & the remote are both configured VERIFY_CA. 

Your patch will then allow the unprivileged client to access restricted remote data.

dorinhogea:
@riverszhang89:
"Your patch will then allow the unprivileged client to access restricted remote data."
Please see the description of the change: it is providing confidentiality, not access control.

db/fdb_fend.c
dorinhogea: It could be user input error.

riverszhang89:
If your patch can't even provide authentication, how do you even provide “confidentiality”? I do recommend you read up on SSL/TLS before making the change.

adizaimi:
Encrypting the channel provides confidentiality. The intention here is not to not have cleartext communication channel.

dorinhogea:
@riverszhang89: failed to see the merits of the example, it seems incorrect.   
If server B trusts client A, it means A certificate chain include a CA that is part of B certificate chain.   In other words, A and B share the same trust tree.  If C trusts B, given the fact that B has only one certificate, it means that C is part of the same of the same trust tree. Therefore point (3) is not possible ("the client certificate is NOT trusted by the remote;"), not in the current implementation based on one server certificate.
Assuming A connects directly to C, it will be able to authenticate and access data.
I will submit a suggestion for improving access control inside comdb2.

PR #737, opened 2018-02-16
Check buffer overflow around use of snprintf.

nirbhayc:


db/sigutil.c
```diff
@@ -98,26 +98,37 @@ void sprintsigact(char *buf, size_t buflen, const struct sigaction *act)
...
     pos = 0;
     for (ii = 0; ii < num_unix_signals; ii++) {
         int signo = unix_signals[ii].signum;
         if (sigismember(&act->sa_mask, signo)) {
             if (pos > 0)
                 mstr[pos++] = '|';
-            pos += snprintf(mstr + pos, sizeof(mstr) - pos, "%s",
-                            unix_signals[ii].name);
+            ret = snprintf(mstr + pos, sizeof(mstr) - pos, "%s",
+                           unix_signals[ii].name);
```
adizaimi: Can we compute ince sizeof - pos so we dont introduce bug in future?

db/sqlglue.c
```diff
@@ -1342,30 +1342,41 @@ static int using_old_style_name(char *namebuf, int len, struct schema *schema,
...
-        current += snprintf(buf + current, sizeof buf - current, "DATACOPY");
-    }
-    if (schema->flags & SCHEMA_DUP) {
-        current += snprintf(buf + current, sizeof buf - current, "DUP");
-    }
-    if (schema->flags & SCHEMA_RECNUM) {
-        current += snprintf(buf + current, sizeof buf - current, "RECNUM");
-    }
+
+    SNPRINTF(buf + current, sizeof buf - current, "%s", dbname)
```
adizaimi: Can we add parens around sizeof () obj because this is hard to read.

db/sqlglue.c
```diff
@@ -1342,30 +1342,41 @@ static int using_old_style_name(char *namebuf, int len, struct schema *schema,
     return stat1_find(namebuf, schema, db, ixnum, trans);
 }
 
+#define SNPRINTF(str, size, fmt, ...)                                          \
+    {                                                                          \
+        ret = snprintf(str, size, fmt, __VA_ARGS__);                           \
+        if (ret >= size)                                                       \
```
adizaimi: We dont do current += size here, which means that we wrote size bytes, but when we goto done to compute crc we wont do this field. Dont know if you want to include this field.

db/sigutil.c
```diff
@@ -98,28 +97,32 @@ void sprintsigact(char *buf, size_t buflen, const struct sigaction *act)
     int ii;
     char fstr[128] = "";
     int pos;
+    int ret;
```
adizaimi: Ret is no longer needed i think, in all of the changes below as well.

adizaimi:
Looks good.

PR #740, opened 2018-02-20
Comment out CTAS syntax in the parser.

nirbhayc:


dorinhogea:
are there any other unsupported SQLite syntax we want to block?

nirbhayc:
> are there any other unsupported SQLite syntax we want to block?

None, I think. I have already removed/disabled quite a few over time. (grep for COMDB2_UNSUPPORTED)

dorinhogea:
This looks good, thank you

nirbhayc:
@dorinhogea Thank you.

PR #741, opened 2018-02-20
Revert "ssl for cross-database sql"

mponomar:
Reverts bloomberg/comdb2#736

Let's revert this until we've discussed it internally.


dorinhogea:
The changes are a step forward, not backward.  It improves on the existing situation when remsql is only possible in clear text.  I don't think this should be reverted.

adizaimi:
I agree with Dorin, ftr.

akshatsikarwar:
Seems reasonable - this is similar to Mike's first cut implementation between client-server?

mponomar:
Ok, fair enough.  Let's leave the current code until we have a better model in place for passing identity between databases.

riverszhang89:
> The changes are a step forward, not backward. It improves on the existing situation when remsql is only possible in clear text. I don't think this should be reverted.

I disagree. First of all, an incomplete implementation with known security weakness isn't a step forward from not implementing it. Second of all, your patch doesn't improve anything - it doesn't change the fact that a trusted network is still required between the server and the remote to prevent MITM attacks.

If you insist to keep the code there, you should at least disable it by default or warn users about the security risk.

PR #742, opened 2018-02-21
Incremental copy mark

markhannum:
Some fixes for Mike's 'incremental copy' work - the main idea is that we're leveraging the incremental backup logic to make minimal copies of databases which have diverged from the originals.  That is, this code allows us to copy only pages that have changed, and overlay them on top of a divergent database.


markhannum:
Not quite ready for merge: I have seen corruption occur when attempting to do a partial copy of a database that's around 10Gb large.  This isn't easily reproducible.

markhannum:
Ah... reproducible..

markhannum:
Partial copy doesn't run recovery.. easy fix

PR #747, opened 2018-02-21
Add system table to track table activity.

mponomar:
Track basic operation counts on per-table basis.  Presents comdb2_table_hits system table as an interface.

### DIFF
--- a/db/comdb2.h
+++ b/db/comdb2.h
@@ -860,6 +860,8 @@ struct dbtable {
     /* lock for consumer list */
     pthread_rwlock_t consumer_lk;
     int disableskipscan : 1;
+
+    int64_t finds, nexts, inserts, deletes, updates;
 };
 
 struct log_delete_state {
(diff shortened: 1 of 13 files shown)


akshatsikarwar:
Does this track queues and/or queuedbs?

mponomar:
Nope, just plain old boring tables.

akshatsikarwar:
can haz?

akshatsikarwar:
maybe just queuedb?

mponomar:
Sure, but it should probably be a different table.  Operations are different (put/consume).

akshatsikarwar:
meh - insert/delete conveys same information

mponomar:
Fair enough.  Also going to move the tracking inside record.c.  Originally wanted it here so we wouldn't double-count on transaction replays.  Don't know how important an accurate count should be.

akshatsikarwar:
inaccurate count seems work fine for autoanalyze so far..

adizaimi:
But who tracks the trackers?

PR #753, opened 2018-02-26
Analyze support

mponomar:


adizaimi:
This looks fine. Since we dont have usage for this function yet, I assume it will be used later.

db/sqlanalyze.c
```diff
@@ -1152,6 +1152,28 @@ int analyze_set_sampling_threshold(void *context, void *thresh)
 /* get sampling threshold */
 long long analyze_get_sampling_threshold(void) { return sampling_threshold; }
 
+/* set maximum analyze compression threads */
+int analyze_set_max_comp_threads( int maxthd )
+{
+    /* must have at least 1 */
+    if( maxthd < 1 )
+    {
+        printf( "%s: invalid value for maxthd\n", __func__ );
```
riverszhang89: Should we replace print with logmsg here?

PR #754, opened 2018-02-26
properly deregister with pmux on test exit for several tests

adizaimi:
* properly deregister with pmux on test exit for the following tests: incremental backups, local rep, qcopy.
* pmux_port set and exported in runtestcase script


adizaimi:
Verified that no db names linger in pmux for this test run.

PR #756, opened 2018-03-01
specify BUILDDIR as an option to make to build in that directory

adizaimi:
* get rid of the unnecessary top level Makefile
* modify tests to call directly executables from build directory
* specify BUILDDIR as an option to build in a different directory rather than build/


riverszhang89:
You could use the following commands to build in a different directory:

```
mkdir /path/to/bin
cd /path/to/bin
cmake /path/to/src
make
```

The top-level makefile is being deprecated - We need it for now to be compatible with the build/CI tools we have internally (builddb, releasedb, roborivers, and etc), as they invoke `make`.

adizaimi:
Rivers, yes that's a good point. I guess that's excactly what the Makefile does, so i was trying to simplify it by just passing BUILDDIR to make. One thing we could improve on as far as testing piece, is to be able to do `make test` from within the build/ directory.

riverszhang89:
> One thing we could improve on as far as testing piece, is to be able to do make test from within the build/ directory.

That'd be ideal. Maybe we should do that instead?

jmleddy:
Why don't you guys have gmake just call cmake

adizaimi:
Addressed comments from offline discussions. Merging this since it is now working fine with the testsuite. If we want to change the way we store the executables, we can do so in a subsequent checkin.

PR #759, opened 2018-03-01
Remove obsolete tunables/attributes/switches.

riverszhang89:
These tunables are no longer in use.


riverszhang89:
Thank you Lingzhi!

PR #760, opened 2018-03-01
Don't truncate pthread_self() to a 32-bit interger.

riverszhang89:
We're seeing crashes in tag requests, when 2 threads whose pthread_self()'s lower 32 bits are the same are accessing the same table.

The bug is caused by our 32-bit legacy code where pthread_self() is casted to an int. Casting pthread_self() to the more portable `intptr_t` fixes it.


riverszhang89:
The change is causing new crashes in tag. Closing for now.

PR #762, opened 2018-03-02
find what is running on port on which db is supposed to run

adizaimi:
We are seeing some failures with 'FAILED TO BIND to port...' and this
checkin attempts to find out what is running on the port on which the db
is supposed to run. Previous attempt at this was targeted at single node
database; this checkin attempts to find out used ports for any node in
a cluster.
The culprit that releases port before fully exiting is qcopy test which is fixed here as well.

### DIFF
--- a/schemachange/sc_records.c
+++ b/schemachange/sc_records.c
@@ -504,6 +504,7 @@ static int convert_record(struct convert_record_data *data)
              * a lock to the last page of the stripe.
              */
 
+            sleep(1);
             // AZ: determine what locks we hold at this time
             // bdb_dump_active_locks(data->to->handle, stdout);
             data->sc_genids[data->stripe] = -1ULL;
--- a/schemachange/sc_records.c
+++ b/schemachange/sc_records.c
@@ -504,7 +504,6 @@ static int convert_record(struct convert_record_data *data)
              * a lock to the last page of the stripe.
              */
 
-            sleep(1);
             // AZ: determine what locks we hold at this time
             // bdb_dump_active_locks(data->to->handle, stdout);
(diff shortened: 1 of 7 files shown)


schemachange/sc_records.c
```diff
@@ -503,6 +503,7 @@ static int convert_record(struct convert_record_data *data)
              * a lock to the last page of the stripe.
              */
 
+            sleep(1);
```
ldennis: Remove this for this PR?

tests/tools/send_msg_port.sh
```diff
@@ -2,6 +2,9 @@
 #
 #send message to port 
 
+host=localhost
```
adizaimi: Can use this instead of nc in pmux tests.

schemachange/sc_records.c
adizaimi: Was testing lost insert -- will remove, thanks.

tests/unsetup
```diff
@@ -78,7 +79,7 @@ function cleanup_cluster {
 }
 
 echo deregistering $DBNAME from pmux:$pmux_port
-deregister_db="pgrep pmux > /dev/null && (exec 3<>/dev/tcp/localhost/${pmux_port}; echo del comdb2/replication/${DBNAME} >&3 )"
+deregister_db_port="pgrep pmux > /dev/null && (exec 3<>/dev/tcp/localhost/${pmux_port}; echo del comdb2/replication/${DBNAME} >&3 )"
```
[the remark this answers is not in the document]
adizaimi: Ah, can't write from remote host. Nevermind.

PR #763, opened 2018-03-02
Sc parallel

ldennis:
Allow multiple schema changes running in parallel


ldennis:
@markhannum reviewed this.

PR #765, opened 2018-03-04
Apple largefile fix

markhannum:
2-line change to fix apple build


akshatsikarwar:
Not sure if I like this. Perhaps we should define  `-D_FILE_OFFSET_BITS=64` in the top level makefile and will work propery across all platforms.
Ref: http://man7.org/linux/man-pages/man2/open.2.html
```
 Setting the _FILE_OFFSET_BITS feature test macro to 64 (rather than using O_LARGEFILE) is the preferred method of accessing large files on 32-bit systems
```

akshatsikarwar:
I will send you a PR to review.

PR #768, opened 2018-03-05
Prefer _FILE_OFFSET_BITS over O_LARGEFILE

akshatsikarwar:


akshatsikarwar:
@markhannum: Can you confirm this addresses largefiles on whatever platform that needed it earlier?

PR #769, opened 2018-03-07
Fix the return type of intervaldsus in comdb2api.

riverszhang89:
The bug makes the database always return `intervalds` values for `intervaldsus` columns, in comdb2api.


riverszhang89:
Thanks. I've added one to the internal test suite.

PR #771, opened 2018-03-07
Plugins: s/PLUGIN/LIBNAME in cmake scripts

nirbhayc:


adizaimi:
Looks correct.

nirbhayc:
Thanks @adizaimi !

PR #773, opened 2018-03-08
Fix cdb2api debug trace

markhannum:
A trace bug that causes a process to crash if the debug-trace flag is enabled


adizaimi:
looks good. sorry for the wasted hours investigating.

markhannum:
No worries

PR #774, opened 2018-03-09
Refactor before I can log bound params

akshatsikarwar:
Make write_response independent of client API

Plugin should not decide when to send a row or skip. This should
be done entirely in sqlinterfaces.c. If plugin is requested to send
a row, it will send the row - no smarts there.

Highlights:
* Stop double copying into fsql buffers; let plugin handle encoding
* Simplify returning rows from SP
* Unify sql row handling
* Delete dead code. Less code is running faster
* Reduce memory allocations
* Remove (almost) all references to FSQL_*

For some next steps:
* Move dbglog to fastsql plugin
* Factor out code to bind parameters; remove cnlt->req, clnt->sql_query
* Delete clnt->is_newsql. Code in sqlinterfaces.c should be independent
    of where a query originated
* Fix send_dbinforesponse: its pushing client protocol down into bdb!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # binaries -- still installing these in-source for test suite
 /build*
+/ninja*
 /deps
 /cdb2_dump
 /cdb2_printlog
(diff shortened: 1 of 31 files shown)


akshatsikarwar:
Code change only in 267b6d8

akshatsikarwar:
I can test without PR. Thx robo.

PR #783, opened 2018-03-19
block  io.open

mohitkhullar:
- bug fix

### DIFF
--- a/lua/liolib.c
+++ b/lua/liolib.c
@@ -159,14 +159,15 @@ static int io_tostring (lua_State *L) {
 
 
 static int io_open (lua_State *L) {
+  /* COMDB2 MODIFICATION */
+  return luaL_error(L, "Function disabled for comdb2.");
+#if 0
   const char *filename = luaL_checkstring(L, 1);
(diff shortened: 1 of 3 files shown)


akshatsikarwar:
Didn't we want to stub out entire `io` module?

mohitkhullar:
blocked the write for now, read mode can be used to read csv files.

lua/liolib.c
```diff
@@ -161,6 +161,9 @@ static int io_tostring (lua_State *L) {
 static int io_open (lua_State *L) {
   const char *filename = luaL_checkstring(L, 1);
   const char *mode = luaL_optstring(L, 2, "r");
+  if (strcmp(mode, "r") != 0) {
+    return luaL_error(L, "Only read mode allowed");
+  }
```
nirbhayc: IMHO, even OP user should not be allowed to read from arbitrary location. We could override this default behavior via a tunable.

nirbhayc:
Also consider providing a way to enable it back, in case this feature is being used in prod.

PR #785, opened 2018-03-19
System table wrapper.

mponomar:
Easier way to create system tables.


mponomar:
Note: import from R6,

markhannum:
I reviewed this code also in R6- okay to merge

PR #788, opened 2018-03-21
berkdb: log_archive returns wrong files.

riverszhang89:
We need the recovery LSN in the checkpoint prior to the dbreg record.
Consider the following example.

```
[1583][32542163]__txn_ckp: rec: 11 txnid 0 prevlsn [0][0]
        ckp_lsn: [1576][5255785]
[1583][32596639]__txn_ckp: rec: 11 txnid 0 prevlsn [0][0]
        ckp_lsn: [1583][32543779]
```

`[1583][32542163]` and `[1583][32596639]` are 2 most recent checkpoints.
The dbreg record may be somewhere between `[1583][32542163]` and `[1583][32543779]`.
A replicant may then truncate to `[1583][32542163]` whose ckp_lsn is `[1576][5255785]` to verify.
Hence log files 1576 - 1582 cannot be safely deleted.

### DIFF
--- a/berkdb/log/log_archive.c
+++ b/berkdb/log/log_archive.c
@@ -243,6 +243,8 @@ __log_archive(dbenv, listp, flags)
 					(void)__log_c_close(logc);
 					goto err1;
 				}
+				if ((ret = __log_c_close(logc)) != 0)
+					goto err1;
 			} else if (dbenv->attr.dbreg_errors_fatal) {
 				*listp = NULL;
 
--- a/berkdb/log/log_archive.c
+++ b/berkdb/log/log_archive.c
@@ -124,7 +124,7 @@ __log_archive(dbenv, listp, flags)
 	DBT rec;
 	DB_LOG *dblp;
 	DB_LOGC *logc = NULL;
-	DB_LSN stable_lsn;
+	DB_LSN dbreg_lsn, stable_lsn;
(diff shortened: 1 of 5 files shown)


dorinhogea:
don't you need to close the cursor here?

berkdb/log/log_archive.c
```diff
@@ -223,8 +223,27 @@ __log_archive(dbenv, listp, flags)
...
+				   The dbreg record is somewhere between [1583][32542163]
+				   and [1583][32543779]. A replicant may then truncate to
+				   [1583][32542163] whose ckp_lsn is [1576][5255785] and
+				   verify from there. Hence log files 1576 - 1582 cannot be deleted. */
+				if ((ret = __log_cursor(dbenv, &logc)) != 0)
+					goto err1;
+				if ((ret = __log_backup(dbenv, logc, &dbreg_lsn, &stable_lsn)) != 0) {
+					(void)__log_c_close(logc);
+					goto err1;
+				}
```
dorinhogea: Here

dorinhogea:
Would it be less conservative to pick the smallest of the curent_chk->chk_lsn and &prev_chk, instead of prev_chk->chk_lsn?  We need the dbreg entries, but not the logs before the prev_chk

riverszhang89:
@dorinhogea Thank you for the review :+1: !

PR #801, opened 2018-04-11
fix autoanalyze to check counters even if no recent writes

adizaimi:
Currently if a table is being analyzed, we save counter for another table, but dont analyze it. Later, when checking whether to do an autoanalyze check for writes and counters, we skip if there is no write in the last CHK_AA_TIME. This checkin makes it check the counters every CHK_AA_TIME.


bdb/attr.h
```diff
@@ -444,6 +444,7 @@ DEF_ATTR(AA_MIN_PERCENT, aa_min_percent, QUANTITY, 20,
          "Percent change above which we kick off analyze.")
 DEF_ATTR(AA_MIN_PERCENT_JITTER, aa_min_percent_jitter, QUANTITY, 300,
          "Additional jitter factor for determining percent change.")
+DEF_ATTR(AA_TESTING, aa_testing, BOOLEAN, 0, "turn AA TESTING on/off")
```
nirbhayc: I had this idea of keeping all the *debugging/testing* related tunables under "debug/test" namespace in the tunables sub-system (something like debug.aa_testing). That way, it will be easier to identify them in lrl files and elsewhere. Also, better description wouldn't hurt. :)

db/autoanalyze.c
```diff
@@ -302,12 +309,9 @@ void stat_auto_analyze(void)
 void *auto_analyze_main(void *unused)
 {
     int now = comdb2_time_epoch();
-    if (now - gbl_sc_last_writer_time >
+    static int aa_last_check_time;
```
riverszhang89: I don't think you need this extra variable.

PR #805, opened 2018-04-16
csv copy and tests

mohitkhullar:


nirbhayc:
Can we call it 'load' instead of 'csvcopy'?

tests/csvcopy.test/t00.expected
```diff
@@ -0,0 +1,11 @@
+[exec procedure sys.cmd.load('/home/mk/data/mohitdb1/agg.csv', 't1')] failed with rc -3 [sys.load(db,csv,tbl,seperator,h...]:2: File doesn't exist
```
nirbhayc: Relative path?

nirbhayc: Also, how does it play with the system's security? Theoretically, one should be able to dump data in password file into a table.. :) I'd restrict the directory from where one can load data.

mohitkhullar: this is a failure test case, it won't work.. only data files inside db's basedir will work for now

nirbhayc: Ok, I overlooked that.
```
+    } else if ((strncmp(thedb->basedir, fname, basedir_len) != 0) || (fname[basedir_len] != '/')) {
+        return luaL_error(lua, "File not in database directory");
```
Can we have a test covering that?

mohitkhullar: Added that test too  t02.req

PR #806, opened 2018-04-16
Test and fix for si logdeletion

markhannum:
We incorrectly delete logfiles that are required to recreate records for ongoing snapshot transactions.  This  pull request contains the silogdel test which reproduces the bug, as well as the fix.


markhannum:
Reviewed by Mike

PR #810, opened 2018-04-17
berkdb: log deletion should take rep_verify_match into account.

riverszhang89:
1. Find the 1st matchable commit record that precedes last sync LSN.
2. Find the checkpoint LSN of the checkpoint that precedes the commit record.
3. Find the checkpoint that precedes the checkpoint LSN.
4. Find the DBREG records of the checkpoint.

We must preserve at least these many log files.

### DIFF
--- a/berkdb/env/env_recover.c
+++ b/berkdb/env/env_recover.c
@@ -2064,4 +2064,100 @@ err:
 		return ret;
 }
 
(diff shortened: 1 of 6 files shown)


riverszhang89:
Reviewed with Mark in person. Thanks Mark!!

PR #811, opened 2018-04-17
partial index fixes

ldennis:
- skip indexes that dont exist when process index shadow table
- fix deadlock of truncating tables with partial/expression indexes


ldennis:
Thanks.

PR #812, opened 2018-04-17
Recover deadlock

ldennis:
- Verify table version does not change after recover_deadlock()
- tableversion remains unchanged for rebuild and truncate
- update db struct and bdb_state in place for alter/rebuld/truncate
- refactor truncate code to behave like alter instead of drop+add


db/sqlinterfaces.c
```diff
@@ -5296,8 +5300,8 @@ int sqlserver2sqlclient_error(int rc)
 {
     switch (rc) {
     case SQLITE_DEADLOCK:
-        return CDB2ERR_DEADLOCK;
     case SQLITE_BUSY:
+    case SQLITE_COMDB2SCHEMA:
         return CDB2ERR_DEADLOCK;
```
ldennis: make it cdb2err_schema

db/sqloffload.c
```diff
@@ -742,6 +752,8 @@ static void osql_scdone_commit_callback(struct ireq *iq)
             broadcast_sc_end(iq->sc->table, iq->sc_seed);
             if (iq->sc->db)
                 sc_del_unused_files(iq->sc->db);
+            if (iq->sc->fastinit && !iq->sc->drop_table)
```
dorinhogea: why do we run analyze after fastinit?

dorinhogea:
Looks good to me.

db/sqloffload.c
ldennis: To clear sqlite_stats

PR #814, opened 2018-04-19
Add malloc_resize() to efficiently resize a memory block.

riverszhang89:
malloc_resize() behaves as realloc(). However if it must reposition the memory block to accommodate the new size, it does not memcpy the content from the old address to the new address.


riverszhang89:
Thank you Akshat!

PR #816, opened 2018-04-21
After running a test, correctly find cores in any of the cluster machines 

adizaimi:
After running a test from the testsuite, we will check for cores. The core location and pattern 
is machine dependent and we try to parse core_pattern to look in the right directory for the 
right  core.


adizaimi:
Had to manually reset head and unmerge this. Trying again in #819

PR #817, opened 2018-04-22
 create_sqlite_master only once after SC_DONE 

adizaimi:
Create_sqlite_master only once after SC_DONE
* Avoid calling multiple times for cases such as fastinit
* create sqlite master tbls on callback, scabort as well


adizaimi:
This PR was split into two parts and submitted as https://github.com/bloomberg/comdb2/pull/851 and https://github.com/bloomberg/comdb2/pull/853. Closing.

PR #824, opened 2018-04-25
Fix oldfile delete

dorinhogea:
ported from R6 PR 820 and 834.


adizaimi:
Looks great. 
Thanks

dorinhogea:
thanks Adi

PR #825, opened 2018-04-25
ported R6 PR 833

dorinhogea:
This uses datetime precision set in the client session if not explicitly provided.

### DIFF
--- a/sqlite/dttz.c
+++ b/sqlite/dttz.c
@@ -18,6 +18,9 @@
 #include <inttypes.h>
 #include <arpa/inet.h>
 
+#include "comdb2.h"
+#include "sql.h"
+
 #include "sqliteInt.h"
 #include "vdbeInt.h"
 #include <types.h>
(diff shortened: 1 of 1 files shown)


sqlite/dttz.c
```diff
@@ -343,8 +348,16 @@ static void nowFunc(sqlite3_context *context, int argc, sqlite3_value **argv)
    const unsigned char *msus;
    assert(context->pVdbe);
 
-   if (argc == 0)
+   if (argc == 0) {
        precision = context->pVdbe->dtprec;
+       if (precision != DTTZ_PREC_MSEC
+               && precision != DTTZ_PREC_USEC) {
+           struct sql_thread *thd =pthread_getspecific(query_info_key);
```
riverszhang89: Missing a space between = and pthread_getspecific.

riverszhang89: Roborivers doesn't check sqlite code format :(

PR #830, opened 2018-04-26
ported R6 PR 855

dorinhogea:
Option to disable rollouts;  main use case is restoring an old backup: without this option, the partitions will rollout and data will be instantly deleted.

### DIFF
--- a/db/views.c
+++ b/db/views.c
@@ -298,7 +298,8 @@ int timepart_add_view(void *tran, timepart_views_t *views,
             goto done;
         }
     } else {
-        fprintf(stderr, "Time partitions rollouts are stopped; no rollouts!\n");
+        logmsg(LOGMSG_WARN,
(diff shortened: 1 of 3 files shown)


db/views.c
```diff
@@ -281,19 +281,24 @@ int timepart_add_view(void *tran, timepart_views_t *views,
...
+                             NULL, NULL, &view->source_id, err) == NULL)
+                 ? err->errval
+                 : VIEW_NOERR;
+        if (rc != VIEW_NOERR) {
+            if (tmp_str)
+                free(tmp_str);
+            goto done;
+        }
+    } else {
+        fprintf(stderr, "Time partitions rollouts are stopped; no rollouts!\n");
```
riverszhang89: logmsg

dorinhogea:
thanks for review

PR #834, opened 2018-04-27
add db create test, add checks for lrl file and dbdir

adizaimi:
* Add a test for db creation
* Add  checks for missing lrl file and for dir name to be nonempty


adizaimi:
Thanks Nirbhay.

PR #843, opened 2018-05-01
@ls/@describe shortcuts to list and describe tables.

nirbhayc:
A quick hack to list/describe tables.

```
nctdb> create table t1(i int primary key, j varchar(10)) $$
[create table t1(i int primary key, j varchar(10))] rc 0
nctdb> create table t2(i int primary key, foreign key (i) references t1(i)) $$
[create table t2(i int primary key, foreign key (i) references t1(i))] rc 0
nctdb> @ls tables
sqlite_stat1
sqlite_stat4
t1
t2
[@ls tables] rc 0
nctdb> @describe t2
Columns:
(column='i', type='int', size=5, sqltype='int', varinlinesize=NULL, defaultvalue=NULL, dbload=NULL, isnullable='N')

Keys:
(keyname='COMDB2_PK', isunique='Y', isdatacopy='N', isrecnum='N', condition=NULL)

Constraints:
(name='$CONSTRAINT_C6A17957', tablename='t2', keyname='COMDB2_PK', foreigntablename='t1', foreignkeyname='COMDB2_PK', iscascadingdelete='N', iscascadingupdate='N')

CSC2:
(csc2='schema
	{
		int i 
	}
keys
	{
		"COMDB2_PK" = i 
	}
constraints
	{
		"COMDB2_PK" -> <"t1":"COMDB2_PK"> 
	}
')
[@describe t2] rc 0
```

### DIFF
--- a/tools/cdb2sql/cdb2sql.c
+++ b/tools/cdb2sql/cdb2sql.c
@@ -548,18 +548,58 @@ static int process_escape(const char *cmdstr)
             fprintf(stderr, "table name required\n");
             return -1;
         } else {
-            int start_time_ms, run_time_ms;
-            char sql[100];
-            int saved_printmode;
+            int start_time_ms;
+            int run_time_ms;
+            int rc;
+            char sql[200];
+            FILE *out = stdout;
 
+            if (printmode & STDERR)
+                out = stderr;
+
+            fprintf(out, "Columns:\n");
+            snprintf(sql, sizeof(sql),
+                     "SELECT columnname AS column, type, size, sqltype, "
+                     "varinlinesize, defaultvalue, dbload, isnullable FROM "
+                     "comdb2_columns WHERE tablename = '%s'",
+                     tok);
+            rc = run_statement(sql, 0, NULL, &start_time_ms, &run_time_ms);
+            if (rc != 0) {
+                return rc;
+            }
+            fprintf(out, "\n");
+
+            fprintf(out, "Keys:\n");
+            snprintf(sql, sizeof(sql),
+                     "select keyname, isunique, isdatacopy, isrecnum, "
+                     "condition from comdb2_keys where tablename = '%s'",
+                     tok);
+            rc = run_statement(sql, 0, NULL, &start_time_ms, &run_time_ms);
+            if (rc != 0) {
+                return rc;
+            }
+            fprintf(out, "\n");
+
+            fprintf(out, "Constraints:\n");
+            snprintf(sql, sizeof(sql),
+                     "select * from comdb2_constraints where tablename = '%s' "
+                     "OR foreigntablename = '%s'",
+                     tok, tok);
+            rc = run_statement(sql, 0, NULL, &start_time_ms, &run_time_ms);
+            if (rc != 0) {
+                return rc;
+            }
+            fprintf(out, "\n");
+
+            fprintf(out, "CSC2:\n");
             snprintf(sql, sizeof(sql),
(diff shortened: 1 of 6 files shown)


riverszhang89:
Cool feature. Can you please write a test?

riverszhang89:
Make `@describe` show DDL instead of CSC2?

nirbhayc:
> Make @describe show DDL instead of CSC2?

IMHO, CSC2 would make more sense for the following reasons:
* That's how definitions are stored on disk
* Easily retrievable via sqlite_master
* At the moment, we don't cache/store schema definition in DDL format, converting CSC2 back to DDL could be buggy, inaccurate (and would require further maintenance)
* This gives me a quick feedback on the DDL->CSC2 correctness

nirbhayc:
> The DDL is stored in sql of sqlite_master. You don't need to convert. DDL will soon become the recommended approach to define database schema. Let's not confuse users, especially those outside bloomberg who have very little knowledge about CSC2.

sqlite_master.sql's definition is barely complete (missing FK, unique constraints, for instance).
```
nctdb> select sql from sqlite_master where tbl_name = 't2';
(sql='create table "t2"("i" int);')
(sql='create index "$COMDB2_PK_75A79413" on "t2" ("i");')

# ==== Versus =====

nctdb> select csc2 from sqlite_master where tbl_name = 't2' and type = 'table';
(csc2='schema
        {
                int i
        }
keys
        {
                "COMDB2_PK" = i
        }
constraints
        {
                "COMDB2_PK" -> <"t1":"COMDB2_PK">
        }
')
```
Will switch it once/when it is accurate, but for now, I'd prefer csc2.

akshatsikarwar:
`select csc2 from sqlite_master`

adizaimi:
I was always annoyed that i could not do 'show tables' or '.tables', and over the years I got used to type the long commands manually. But this looks useful. Are we going to document this?

nirbhayc:
> Are we going to document this?

Certainly, along with all other similar 'client-side' commands.

tools/cdb2sql/cdb2sql.c
```diff
@@ -524,6 +527,80 @@ static int process_escape(const char *cmdstr)
     } else if (strcmp(tok, "time") == 0) {
         time_mode = time_mode ? 0 : 1;
         printf("Timing mode %s\n", time_mode ? "ON" : "OFF");
+    } else if ((strcmp(tok, "ls") == 0) || (strcmp(tok, "list") == 0)) {
```
riverszhang89: strcasecmp.

tools/cdb2sql/cdb2sql.c
```diff
@@ -524,6 +527,80 @@ static int process_escape(const char *cmdstr)
...
+        tok = strtok_r(NULL, delims, &lasts);
+        if (!tok || strcmp(tok, "tables") == 0) {
```
riverszhang89: ditto.

tools/cdb2sql/cdb2sql.c
```diff
@@ -524,6 +527,80 @@ static int process_escape(const char *cmdstr)
...
+            int start_time_ms, run_time_ms;
+            const char *sql = "SELECT tablename FROM comdb2_tables";
+            int saved_printmode;
+
+            saved_printmode = printmode;
+            printmode = TABS;
```
riverszhang89: Should honor user print mode. Would be prettier under your tabular change.

nirbhayc:
Thanks!

PR #849, opened 2018-05-02
cdb2sql: tabular display mode

nirbhayc:


tools/cdb2sql/cdb2sql.cpp
```diff
@@ -338,7 +343,7 @@ static char *read_line()
             free(line);
             line = NULL;
         }
-        if ((line = readline(prompt)) != NULL && line[0] != 0)
+        if ((line = readline(prompt)) != NULL && line[1] != 0)
```
riverszhang89: Should it be 0 here?

tools/cdb2sql/cdb2sql.cpp
```diff
@@ -236,12 +241,12 @@ char *db_generator (int state, const char *sql)
             if ( sz < count + 1 ) {
                 sz = (sz == 0) ? 32 : sz * 2;
                 void * m = (char**) realloc(db_words, sz * sizeof(char *));
```
riverszhang89: Declare m as char **

PR #853, opened 2018-05-03
some cleanup of file.c

adizaimi:
Cleanup file.c a bit and more:
* use length returned from snprintf rather than running strlen uselessly
* use sizeof(char array[]) - 1 instead of strlen(array) in several cases
* strlen(name) outside of loop
* shortcut with continue while loops rather than having three indentation levels
* more tables in test to check for leaks etc.

### DIFF
--- a/bdb/file.c
+++ b/bdb/file.c
@@ -343,7 +343,7 @@ int bdb_get_new_prefix(char *buf, size_t buflen, int *bdberr)
         return -1;
     }
 
-    if (strlen(NEW_PREFIX) >= buflen) {
+    if (sizeof(NEW_PREFIX) - 1 >= buflen) {
         *bdberr = BDBERR_BUFSMALL;
         return -1;
     }
@@ -363,14 +363,14 @@ const char *bdb_unprepend_new_prefix(const char *tablename, int *bdberr)
 {
     /* if the input didn't start with new. report it by setting bdberr, note
      * that this may not be an error */
-    if (strncmp(tablename, NEW_PREFIX, strlen(NEW_PREFIX))) {
+    if (strncmp(tablename, NEW_PREFIX, sizeof(NEW_PREFIX) - 1)) {
         *bdberr = BDBERR_BADARGS;
         return tablename;
     }
 
     *bdberr = BDBERR_NOERROR;
     /* we want to remove the new. prefix from the tablename */
-    return tablename + strlen(NEW_PREFIX);
+    return tablename + sizeof(NEW_PREFIX) - 1;
 }
 
 /* this removes a new.SOMETHING. prefix from a tables name (if it exists)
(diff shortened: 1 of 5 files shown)


bdb/file.c
```diff
@@ -7662,159 +7664,156 @@ static int bdb_process_unused_files(bdb_state_type *bdb_state, tran_type *tran,
...
+                    *p_buf_end = p_buf + sizeof(invers);
+            struct bdb_file_version_num_type p_file_version_num_type;
+
+            bdb_file_version_num_get(&p_file_version_num_type, p_buf,
+                                     p_buf_end);
+
+            file_version = p_file_version_num_type.version_num;
+        }
+
+        if (file_version)
```
ldennis: I think this is wrong. Schema change is not deleting files now. Should be `if (!file_version)`

bdb/file.c
```diff
@@ -7662,159 +7664,156 @@ static int bdb_process_unused_files(bdb_state_type *bdb_state, tran_type *tran,
...
+        /* try to find the file version amongst the active data files */
+        for (i = 0; i < bdb_state->numdtafiles; ++i) {
+            if (bdb_state->bdbtype == BDBTYPE_QUEUEDB) {
+                rc = bdb_get_file_version_qdb(bdb_state, tran, &version_num,
+                                              bdberr);
+            } else {
+                rc = bdb_get_file_version_data(bdb_state, tran, i, &version_num,
+                                               bdberr);
+            }
+            if (rc == 0 && version_num == file_version) {
```
ldennis: I just figured out that this is also wrong.. We dont want to break if rc == 0..

adizaimi: That would be behavior different from original. It seems that we declare found_in_llmeta = 1 if rc==0 and if version retrieved matched.  I did not write the original but it looks correct.

bdb/file.c
adizaimi: thanks for catching. we want to continue if !file_version.

bdb/file.c
ldennis: I am saying the case where if rc == 0 but version_num != file_version, your patch will break out of the for loop but we should continue instead.

ldennis: try create a table with blob, new code will mistakenly delete all blob files.

bdb/file.c
```diff
@@ -7662,159 +7664,156 @@ static int bdb_process_unused_files(bdb_state_type *bdb_state, tran_type *tran,
...
-                for (i = 0; i < bdb_state->numdtafiles; ++i) {
-                    if (bdb_state->bdbtype == BDBTYPE_QUEUEDB) {
-                        rc = bdb_get_file_version_qdb(bdb_state, tran,
-                                                      &version_num, bdberr);
-                    } else {
-                        rc = bdb_get_file_version_data(bdb_state, tran, i,
-                                                       &version_num, bdberr);
-                    }
-                    if (rc == 0) {
-                        if (version_num == file_version) {
```
ldennis: FYI: here is the original logic

bdb/file.c
ldennis: scindex catches it occasionally because of missing files. But I think we should really have a dedicated test for file deletion.

adizaimi: You are a genius for catching this!

PR #856, opened 2018-05-03
Fix the parsing of partial index's where clause

nirbhayc:
https://github.com/bloomberg/comdb2/issues/855


nirbhayc:
Thanks @ldennis

Issue #857, opened 2018-05-03
DDL: Provide a tunable to control automatic dropping of dependent objects on DROP


nirbhayc:
Currently, dropping a column or an index automatically drops the dependent indexes and constraints.
```
nctdb> CREATE TABLE t1(i INT PRIMARY KEY) $$
[CREATE TABLE t1(i INT PRIMARY KEY)] rc 0
nctdb> CREATE TABLE t2(i INT PRIMARY KEY, j INT, FOREIGN KEY (i) REFERENCES t1(i)) $$
[CREATE TABLE t2(i INT PRIMARY KEY, j INT, FOREIGN KEY (i) REFERENCES t1(i))] rc 0
nctdb> SELECT csc2 FROM sqlite_master WHERE name = 't2';
(csc2='schema
        {
                int i
                int j null = yes
        }
keys
        {
                "COMDB2_PK" = i
        }
constraints
        {
                "COMDB2_PK" -> <"t1":"COMDB2_PK">
        }
')
[SELECT csc2 FROM sqlite_master WHERE name = 't2'] rc 0
nctdb> ALTER TABLE t2 DROP COLUMN 'i' $$
[ALTER TABLE t2 DROP COLUMN 'i'] rc 0
nctdb> SELECT csc2 FROM sqlite_master WHERE name = 't2';
(csc2='schema
        {
                int j null = yes
        }
')
[SELECT csc2 FROM sqlite_master WHERE name = 't2'] rc 0
```
The new behavior would be to only allow this if 'ddl_cascade_drop' tunable is ON.

PR #858, opened 2018-05-03
DDL: Provide a tunable to control automatic dropping of dependent objects on DROP

nirbhayc:
https://github.com/bloomberg/comdb2/issues/857

### DIFF
--- a/db/config.c
+++ b/db/config.c
@@ -358,6 +358,7 @@ static void add_legacy_default_options(struct dbenv *dbenv)
         "on accept_on_child_nets",
         "env_messages",
         "off return_long_column_names",
+        "ddl_cascade_drop 0"
         "setattr NET_SEND_GBLCONTEXT 1",
         "setattr ENABLE_SEQNUM_GENERATIONS 0",
         "setattr MASTER_LEASE 0",
--- a/db/db_tunables.c
+++ b/db/db_tunables.c
@@ -196,6 +196,8 @@ extern bool gbl_rcache;
 static char *name = NULL;
 static int ctrace_gzip;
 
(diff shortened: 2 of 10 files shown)


adizaimi:
I am a bit confused from the description in https://github.com/bloomberg/comdb2/issues/857 and the title. From the code change it seems we disallow to drop the column if it is part of an index, is that what this change is really about?

nirbhayc:
> I am a bit confused from the description in #857 and the title. From the code change it seems we disallow to drop the column if it is part of an index, is that what this change is really about?

Yes. There was a typo in the description. Thanks for pointing out.

dorinhogea:
It seems that user needs to run a bunch of DDL in one transaction now to remove a column, I am not sure about this.

mponomar:
Doesn't have to be a transaction.  I think it's better to make the changes explicit.

nirbhayc:
> It seems that user needs to run a bunch of DDL in one transaction now to remove a column, I am not sure about this.

In ALTER TABLE multiple actions can be chained in a single command.

adizaimi:
What would that ALTER look like?

nirbhayc:
> What would that ALTER look like?

Checkout the test in this PR

nirbhayc:
Thanks @mponomar !

db/config.c
```diff
@@ -358,6 +358,7 @@ static void add_legacy_default_options(struct dbenv *dbenv)
         "on accept_on_child_nets",
         "env_messages",
         "off return_long_column_names",
+        "ddl_cascade_drop 0"
```
ldennis: Missing `,`

nirbhayc:
Thanks @ldennis !

PR #861, opened 2018-05-03
Force lua to break out of sqlite_step if we fail to write a row

markhannum:
Break out of lua stored procedure if sending a row back to the client fails


adizaimi:
Looks good. Could we add a simple test program that disconnects after the query?

PR #863, opened 2018-05-04
no more kludge index syntax check

ldennis:
no more kludge index syntax check by creating a custom SQLite VM inline


adizaimi:
Looks great. Thanks.

ldennis:
Thanks.

PR #864, opened 2018-05-04
Enable CREATE UNIQUE INDEX to handle duplicate NULL values when opted-in via 'uniqnulls' index flag

mistachkin:
This change permits duplicate NULL values to be present in a UNIQUE INDEX if the 'uniqnulls' flag is present in the csc2 definition.


db/comdb2.c
```diff
@@ -3756,7 +3758,9 @@ static int init(int argc, char **argv)
 
     /* some dbs have lots of tables and spew on startup.  this just wastes
      * peoples time shunting spew */
-    /*showdbenv(thedb);*/
+    if (getenv("CDB2_SHOW_DBENV")) {
```
mponomar: I'm in favor of getting rid of that trace altogether.  I don't think it was ever helpful.

mistachkin: It was helpful to me while debugging this change.  I'm not married to it, but it's nice to have it gated behind the environment variable.

mponomar: Works for me.

db/glue.c
```diff
@@ -870,8 +870,57 @@ int cmp_context(struct ireq *iq, unsigned long long genid,
 /*        TRANSACTIONAL INDEX ROUTINES        */
 /*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 
+int ix_isnullk(void *db_table, void *key, int ixnum)
+{
+    struct dbtable *db = db_table;
+    struct schema *dbixschema;
+    int ifld;
+    if (!db || !key || ixnum < 0 || ixnum >= db->nix) {
+        /* fprintf(stderr,
```
riverszhang89: It'd be better to replace them with `logmsg(LOGMSG_DEBUG, ...)`. An example is [here](https://github.com/bloomberg/comdb2/blob/master/bdb/file.c#L2012).

tests/basic.test/runit
```diff
@@ -330,4 +338,42 @@ cdb2sql ${CDB2_OPTIONS} $dbnm default 'exec procedure sys.cmd.send("sql hist")'
 res=`cdb2sql --tabs ${CDB2_OPTIONS} $dbnm default "select 1"`
 assertres $res 1
 
+echo "create and use UNIQUE INDEX with multiple NULL values"
```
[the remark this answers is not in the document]
mistachkin: I thought about doing that; however, given the overhead involved in creating an entirely new test file and the speed at which these new tests will execute, it seemed like overkill.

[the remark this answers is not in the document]
ldennis: But it would give us a better idea of what the test is testing for.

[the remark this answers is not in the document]
adizaimi: Might I suggest we add to a different existing test, such as sql.test/, which would allow us to test this feature as a separate test (and document in the README for that test), and also keep basic simple and short.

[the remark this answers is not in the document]
mistachkin: Good idea, doing that now.

bdb/upd.c
```diff
@@ -297,19 +297,14 @@ static int bdb_prim_updkey_genid_int(bdb_state_type *bdb_state, tran_type *tran,
         return 1;
     }
 
-    if (bdb_keycontainsgenid(bdb_state, ixnum)) {
-        assert(0 == bdb_inplace_cmp_genids(bdb_state, oldgenid, genid));
-        masked_genid = get_search_genid(bdb_state, genid);
-        memcpy(keymax, key, keylen);
-        memcpy(keymax + keylen, &masked_genid, sizeof(unsigned long long));
+    bdb_maybe_use_genid_for_key(bdb_state, &dbt_key, key, ixnum, genid, isnull, &pKeyMaxBuf);
+    assert(!bdb_keycontainsgenid(bdb_state, ixnum) || bdb_inplace_cmp_genids(bdb_state, oldgenid, genid));
```
mponomar: bdb_inplace_cmp_genids(bdb_state, oldgenid, genid) == 0?

docs/src/sqlitegen/bubble-generator-data.tcl
```diff
@@ -457,6 +457,7 @@ set all_graphs {
           {line
               {opt dup}
               {opt datacopy}
+              {opt uniqnulls}
```
mponomar: We also need to re-generate the diagrams.

tests/sql.test/runit
```diff
@@ -184,4 +184,71 @@ for testcase in $files ; do
...
+res=`cdb2sql --tabs ${CDB2_OPTIONS} $a_dbn default "select x from t3 order by x"`
+assertres "${res}" $'NULL\nNULL\nNULL\n2'
+cdb2sql ${CDB2_OPTIONS} $a_dbn default "exec procedure sys.cmd.verify('t3')"
+cdb2sql ${CDB2_OPTIONS} $a_dbn default "update t3 set x = NULL"
+res=`cdb2sql --tabs ${CDB2_OPTIONS} $a_dbn default "select x from t3 order by x"`
+assertres "${res}" $'NULL\nNULL\nNULL\nNULL'
+cdb2sql ${CDB2_OPTIONS} $a_dbn default "exec procedure sys.cmd.verify('t3')"
+cdb2sql ${CDB2_OPTIONS} $a_dbn default "drop index t3_x"
+cdb2sql ${CDB2_OPTIONS} $a_dbn default "exec procedure sys.cmd.verify('t3')"
+
```
mponomar: We should have a couple of schema change tests: make sure we can

- change an existing index into uniqnulls
- change it back to a plain unique index
- schema change that doesn't change the flag still works
- changing an index that has nulls back into unique fails

mponomar:
Lingzhi pointed out one more thing.  We have a nullfkeys setting (per database, unfortunately) that makes the db not do foreign key constraint checks on null keys.  uniqnull keys that contain null should behave that way by default - skip foreign key checks if any fields are null.

docs/src/sqlitegen/bubble-generator-data.tcl
mistachkin: It looks like I'll need to find a Linux machine with both X-Windows and Tk in order to do this.

mistachkin:
I'm trying to figure out a strategy for testing the 'nullfkey' related changes to the block processor.  Perhaps these changes will be covered by some existing tests that I can modify to include a couple 'uniqnulls' indexes?

mponomar:
Hmm, there's no test for nullfkey.  It should be toggle-able live with "PUT TUNABLE nullfkey 1".  There's a constraints test - it could go there.

mistachkin:
@mponomar I would appreciate any additional feedback you may have on my most recent commits.

mponomar:
Looks good!

PR #866, opened 2018-05-04
Internal sql execution

nirbhayc:
This PR refactors existing code to execute sql commands internally.


db/isql.c
```diff
@@ -0,0 +1,222 @@
+/*
```
riverszhang89: unixODBC CLI is also called isql.

PR #867, opened 2018-05-04
prevent fk pointing to a time partition shard

dorinhogea:
Creating a time partition using a source table that has reverse foreign keys is an error.
Altering a table to point to a shard table is an error.


tests/timepart_constraints.test/runtest.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
adizaimi: This script does not need to be checked in.

tests/timepart_constraints.test/runit
```diff
@@ -0,0 +1,100 @@
...
+if (( $? != 0 )) ; then
+   echo "FAILURE to create a normal partition"
+   exit 1
+fi
+
+timepart_stats
+
+# test 3: check attempt to drop a shard
+shard=`cdb2sql -tag ${CDB2_OPTIONS} $dbname default 'select shardname from comdb2_timepartshards limit 1'`
+cdb2sql ${CDB2_OPTIONS} $dbname default "drop table ${shard}" >> $OUT
```
adizaimi: Just wondering if we could capture the returned error message and check that it also says what we want it to say.

tests/timepart_constraints.test/runtest.sh
dorinhogea: removed

dorinhogea:
the whole bpfunc error mechanism needs to be re-addressed; we will do that in another PR.

PR #868, opened 2018-05-04
copy cores from cluster nodes; fix make setup

adizaimi:
* Copy cores from cluster nodes
* Fix running 'make setup' from original test directory


adizaimi:
Closing -- the testsuite is repeatedly rerunning the same test -- need to delete this branch.

PR #872, opened 2018-05-07
removing checkpoint-based file cleanup

dorinhogea:
the cleanup was already fixed; this needs to be removed.


adizaimi:
Looks good to me.

PR #873, opened 2018-05-07
fix testsuite

markhannum:
Make sure to start pmux when runnings tests


adizaimi:
Looks good. Thank you for fixing.

PR #880, opened 2018-05-09
bdb: Fix race between election and shutdown.

riverszhang89:
Election may race with database shutdown. The patch fixes the race by making backend_close() wait for election to abort before destroying the backend resources.


riverszhang89:
Thank you!

PR #881, opened 2018-05-10
add test wich tests stripe order for inserts (round_robin)

adizaimi:
* Add a test which tests stripe ordering for inserts, round_robin vs thread affinity
* setup will use the lrl from the testdir copy


markhannum:
Looks awesome- good to merge

PR #883, opened 2018-05-13
sc_seed no timebased genid, no 2: in front of comdb2_rowid

adizaimi:
sc_seed to not have timebased genid because it screws with the genid48 generation and more:
* Use real get_genid for seed, machine and time info from hash value
* dont print '2:' in front of comdb2_rowid
* add to stripe test to print genid in hex and getting stripe from genid by modding it with 16
* better dumping to logs when setup and unsetup dbs from tests
* in setup, check for used port by checking pid as soon as possible rather than when a failing 'select 1'
* in unsetup, clean DBDIR from test runner node as well even when not part of cluster

### DIFF
--- a/bdb/bdb_api.h
+++ b/bdb/bdb_api.h
@@ -1951,7 +1951,6 @@ void lock_info_lockers(FILE *out, bdb_state_type *bdb_state);
 
 const char *bdb_find_net_host(bdb_state_type *bdb_state, const char *host);
 
-unsigned long long get_next_sc_seed(bdb_state_type *bdb_state);
 unsigned long long bdb_get_a_genid(bdb_state_type *bdb_state);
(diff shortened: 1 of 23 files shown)


ldennis:
Please also make the same change to `plugin/bb-plugins/schema_change/schema_change.c <<handle_schema_change_request>>`

ldennis:
Looks like pmux_hello failed every time, otherwise it looks good.

adizaimi:
> Looks like pmux_hello failed every time, otherwise it looks good.

Yes, I changed the output of pmux command, but did not restart it on all machines otherwise all *other* tests would fail and this would pass.

adizaimi:
Thank you Lingzhi.

PR #884, opened 2018-05-14
Schemalk fix

markhannum:
This fixes a rare and complicated schema-change crash that can occur on the replicant when a schema-change's "close" record occurs after the previous transactions LOCK_PUT_READ, but before the previous transaction's commit record.


bdb/file.c
```diff
@@ -6299,10 +6307,11 @@ static int bdb_del_file(bdb_state_type *bdb_state, DB_TXN *tid, char *filename,
     if ((rc = access(pname, F_OK)) == 0) {
         int rc;
 
+        bdb_state->dbenv->txn_begin(bdb_state->dbenv, NULL, &tid, 0);
```
ldennis: tid is already passed ihn

schemachange/sc_logic.c
```diff
@@ -1003,6 +997,12 @@ int delete_temp_table(struct ireq *iq, struct dbtable *newdb)
         return -1;
     }
 
+    rc = bdb_close_only_sc(newdb->handle, tran, &bdberr);
```
ldennis: handle retry case

bdb/file.c
```diff
@@ -6300,9 +6308,9 @@ static int bdb_del_file(bdb_state_type *bdb_state, DB_TXN *tid, char *filename,
         int rc;
 
         if ((rc = db_create(&dbp, dbenv, 0)) == 0 &&
-            (rc = dbp->open(dbp, NULL, pname, NULL, DB_BTREE, 0, 0666)) == 0) {
+            (rc = dbp->open(dbp, tid, pname, NULL, DB_BTREE, 0, 0666)) == 0) {
```
ldennis: Looks like we cant open and close using the same tid. This is causing some issues for #890 . We had a bug in file deletion before which caused sc not deleting files. Adi and I are trying to fix that via #890 and I think that triggers the bug here. The issue seems to be that on master we revoke dbreg immediately on close regardless of whether we have txn but on replicants we revoke only if we see commit record for the txn. So the order we open/close dbregs are different between master and replicants. This is a guess and we need to look more into it.

markhannum: If this is the case the easy fix is to defer closing the old-tables until after the commit for the sc has been written.

PR #890, opened 2018-05-16
fix bug from cleanup

adizaimi:
Fixing bug from a previous cleanup in https://github.com/bloomberg/comdb2/pull/853

### DIFF
--- a/bdb/file.c
+++ b/bdb/file.c
@@ -7767,14 +7767,16 @@ static int bdb_process_unused_files(bdb_state_type *bdb_state, tran_type *tran,
         }
 
         /* try to find the file version amongst the active indices */
-        for (i = 0; !found_in_llmeta && i < bdb_state->numix; ++i) {
+        for (i = 0; i < bdb_state->numix; ++i) {
             if (bdb_state->bdbtype == BDBTYPE_QUEUEDB)
                 break;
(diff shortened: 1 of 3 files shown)


bdb/file.c
```diff
@@ -7753,9 +7753,11 @@ static int bdb_process_unused_files(bdb_state_type *bdb_state, tran_type *tran,
                 rc = bdb_get_file_version_data(bdb_state, tran, i, &version_num,
                                                bdberr);
             }
-            if (rc == 0 && version_num == file_version) {
-                found_in_llmeta = 1;
-                break;
+            if (rc == 0) {
```
riverszhang89: Why?

adizaimi: Putting back to how old code was...nothing fancy here, just correcting my mistake. Thanks.

riverszhang89: Thanks!

ldennis:
This looks good to me. However, (probably) due to some issues introduced in #884, it is not passing tests yet.

adizaimi:
master branch passes. This branch not rebased to the latest master, also passes the failing tests above.
Adding checkin from #884 seems to start causing the abort to be called (rep_record.c:2968: transaction failed at xxx).

Issue #888, opened 2018-05-15
Command line option to override tunables


akshatsikarwar:
Eg: `comdb2 akdb --tunable "directio no" --tunable "dtastripe 1"`
Will add to lrl in create mode; override lrl otherwise

mponomar:
I think it's a great idea.

PR #891, opened 2018-05-16
Cmd line tunable

akshatsikarwar:
Issue: https://github.com/bloomberg/comdb2/issues/888


akshatsikarwar:
I can probably free these after done processing. No use keeping them around.

PR #899, opened 2018-05-22
have shell as bash instead of sh for two tests

adizaimi:
* have shell as bash instead of sh for two tests -- if sh is the reduced shell installed on some machines, it will fail to run these two tests.
* write verbose connection information to .conn for single node tests


akshatsikarwar:
Please see how other tests invoke `bash` and do the same. We have gnu bash dependency and `/bin/bash` leaves us no way to run those tests.

PR #900, opened 2018-05-22
invoke correct bash version for tests

adizaimi:
invoke correct bash version via /usr/bin/env for tests


adizaimi:
These tests are failing for master and are not part of the scripts that i modified here.

PR #901, opened 2018-05-23
reorder table sockops to not have deadlocks on master

adizaimi:
Reorder table sockops to not have deadlocks on master, resulting in less work done by the system (undoing aborted transaction).

When replicant node sends operations to master, they are in the order which client issued them, for instance:
```
begin
 insert into t1..
 insert into t2..
 insert into t1..
commit
```

Previously we processed them in same order on the master node. In this checkin, we now reorder these operations to look like this:
```
insert into t1..
insert into t1..
insert into t2..
```

The reordering is done by changing the key of the tmptbl [where we store the osql operations]: before the key was just a sequence number, now it is `table, stripe, genid, is_record/blob, sequence`. Since ADD operations dont have a genid preassigned, we will now insert ADD ops into a separate tmptbl using same key, and in process_this_session we will process UPD/DEL operations first (for a given tbl/stripe, then the ADD ops). This ensures that write operations will not get page locks in data tables in a manner resulting in a deadlock. [We still can deadlock for index access since that one is not ordered yet.]

As part of this checkin, the order of operations sent from replicant needs to change--we send record before blobs and partial index (because we need a genid to order the latter correctly), however the order which the operations is saved in the table is as before (blobs before record), so that toblock processing has not been modified.

Feature can be turned on or off live via tunable `reorder_socksql_no_deadlock.` When switch is on, replicant node will send in the first packet is_reorder_on to master node, and will send the osql_record before osql_blobs. Master node will save is_reorder_on flag and store that session by filling key appropriately. If is_reorder_flag is not set, the key is 0 except for seq, and thus ops will not be reordered.


db/comdb2.c
```diff
@@ -621,6 +621,7 @@ int gbl_check_wrong_db = 1;
 int gbl_broken_max_rec_sz = 0;
 int gbl_private_blkseq = 1;
 int gbl_use_blkseq = 1;
+int gbl_reorder_blkseq_no_deadlock = 1;
```
markhannum: you're reordering the socksql stream, not blkseq - maybe change this to 'gbl_reorder_socksql_no_deadlock'?

db/osqlblockproc.c
```diff
@@ -140,6 +146,34 @@ static int osql_bplog_key_cmp(void *usermem, int key1len, const void *key1,
     oplog_key_t *k2 = (oplog_key_t *)key2;
 #endif
```
adizaimi: 1) add optimization: if reordering is off, only check the seq.
2) make sure that the tunable is read only so we dont flip it on the fly.

db/osqlcomm.c
```diff
@@ -6983,7 +6999,7 @@ int osql_process_packet(struct ireq *iq, unsigned long long rqid, uuid_t uuid,
                         NULL,            /*nulls, no need as no
                                            ctag2stag is called */
                         blobs, MAXBLOBS, /*blobs*/
-                        &err->errcode, &err->ixnum, &rrn, &genid, /*new id*/
+                        &err->errcode, &err->ixnum, &rrn, &newgenid, /*new id*/
```
adizaimi: get rid of unused genid variable up.

db/osqlcomm.c
```diff
@@ -7143,7 +7162,7 @@ int osql_process_packet(struct ireq *iq, unsigned long long rqid, uuid_t uuid,
             NULL, NULL,                            /* vrec */
             NULL,                                  /*nulls, no need as no
                                                      ctag2stag is called */
-            *updCols, blobs, MAXBLOBS, &genid, dt.ins_keys, dt.del_keys,
+            *updCols, blobs, MAXBLOBS, &newgenid, dt.ins_keys, dt.del_keys,
```
adizaimi: remove genid.

db/osqlcomm.c
```diff
@@ -7614,7 +7633,7 @@ static int sorese_rcvreq(char *fromhost, void *dtap, int dtalen, int type,
        added session have sess->iq set
        to avoid racing against signal_rtoff code */
     sess = osql_sess_create_sock(sqlret, sqllenret, req.tzname, type, req.rqid,
-                                 uuid, fromhost, iq, &replaced);
+                                 uuid, fromhost, iq, &replaced, uuid_req.flags);
```
adizaimi: the legacy codepath never initalizes the uuid_req thus flags is garbage.

db/osqlblockproc.c
```diff
@@ -608,6 +642,81 @@ const char *osql_reqtype_str(int type)
...
+void setup_reorder_key(int type, osql_sess_t *sess, struct ireq *iq, char *rpl,
+                       oplog_key_t *key)
+{
+    key->tbl_idx = USHRT_MAX;
+    if (type == OSQL_USEDB) {
+        const char *get_tablename_from_rpl(const char *rpl);
+        const char *tablename = get_tablename_from_rpl(rpl);
+        assert(tablename); //table or queue name
+        if (tablename && !is_tablename_queue(tablename, strlen(tablename))) {
+            struct dbtable *tbl = get_dbtable_by_name(tablename);
```
adizaimi: Cant do this without a lock. Caching tbl that may be dropped will result in unexpected behavior. 
Cache the tablename instead and get the dbtable by name every time you need to create a new genid, guarded by appropriate lock. Tbl index can be crc32 of tablename.

db/osqlcomm.c
adizaimi: genid is the vgenid of the function, newgenid will be written with the new genid if one is created.

db/osqlblockproc.c
adizaimi: Actually i have a testcase where the tunable is flipped online. The db can handle it.

adizaimi:
Tested Sysbench insert test and the performance is very similar to that of master branch, but about 3% slower: this branch 10.2k tps, vs masterbranch 10.6k tps. The bad numbers about two weeks ago were due to having debug enabled.

db/osqlblockproc.c
```diff
@@ -64,10 +64,10 @@
 #include "osqlrepository.h"
 #include "comdb2uuid.h"
 #include "bpfunc.h"
+#include "crc32c.h"
```
ldennis: remove this?

db/comdb2.c
```diff
@@ -807,6 +822,18 @@ struct dbtable *get_dbtable_by_name(const char *p_name)
...
 }
 
+unsigned long long get_next_genid_for_table(const char *tablename)
+{
+    struct dbtable *tbl = NULL;
+    unsigned long long genid = 0;
+    Pthread_rwlock_rdlock(&thedb_lock);
+    tbl = hash_find_readonly(thedb->db_hash, &tablename);
+    if (tbl)
+        genid = bdb_get_next_genid(tbl->handle);
```
ldennis: Is there a case where tbl is valid but tbl->handle is not? Maybe thedb_lock is not enough to protect the handle from being swapped (i.e. finalize schema change).. However, I do notice that the underlying function calls only use bdb_state->parent/bdb_state->attr which should not change even during swap. So it might be fine but it worth double checking.

adizaimi:
Sysbench numbers are in:
REORDER {"tps": 221855.17} {"tps": 19901.43} {"tps": 17528.37} {"tps": 101960.46} {"tps": 1176.77} {"tps": 11168.28} {"tps": 12261.05} {"tps": 35576.56} {"tps": 24910.06}
MASTER  {"tps": 231059.23} {"tps": 20649.71} {"tps": 18666.08} {"tps": 104514.41} {"tps": 1106.84} {"tps": 11875.43} {"tps": 12913.59} {"tps": 36262.43} {"tps": 25073.20}

adizaimi:
latest bbtests:
cdb2test   Nov 29 10:50:14 2018  in-comdb2db AZ_reorder_table_no_deadlock.R20181129.2

markhannum:
Reviewed with Adi- this is very terse, fantastic code.  Adi pointed out that this is a first step toward per-table stripe affinity, which could give us some really meaningful gains.  Okay to merge- I'm looking forward to seeing the further work here in reordering our index operations.

PR #902, opened 2018-05-24
Handle case where begin gets a valid error from server

mohitkhullar:
Handle case where begin gets a valid error from server


mohitkhullar:
Added test case for begin error from server

riverszhang89:
This looks good. Please hold - I will port the change to cdb2jdbc.

riverszhang89:
Turns out that cdb2jdbc handles the case correctly. No need to port :)

PR #905, opened 2018-05-25
Fix tunables handling

nirbhayc:
Process all tunables that are not registered in the new tunables
sub-system (along with 'if', 'on' and 'setattr') before passing
the lrl line to handle_lrl_tunable(). This takes care of spurious
'non-registered tunable' warnings.


nirbhayc:
Thanks @mponomar , will wait for robo's and merge.

PR #911, opened 2018-05-30
Add Tcl bindings...

mistachkin:
This pull request adds bindings for use by the Tcl scripting language.  It is designed to provide a very thin wrapper over the low-level C API, e.g. exposing primitive commands that can be used to build more complex and/or specialized layers above it, e.g. TDBC, etc.


CMakeLists.txt
```diff
@@ -118,6 +118,7 @@ add_subdirectory(protobuf)
 add_subdirectory(schemachange)
 add_subdirectory(sockpool)
 add_subdirectory(sqlite)
+add_subdirectory(tcl)
```
mponomar: We should either document tcldev as a dependency, or make this optional.

cmake/FindTcl.cmake
```diff
@@ -0,0 +1,9 @@
+include(${CMAKE_MODULE_PATH}/pkg_helper.cmake)
+find_pkg_for_comdb2(Tcl
```
mponomar: I didn't have tcl.h on my system - it passed cmake and failed at compile time - this step should fail if not found (or tcl bindings can be optional and only build if tcl exists).

CMakeLists.txt
mistachkin: Good point, I'll need to figure out how to accomplish that with CMake.

cmake/FindTcl.cmake
mistachkin: I'll figure out how to make the necessary changes for CMake; however, I may need your help in testing them as most (other) systems will probably have the Tcl stuff installed?

CMakeLists.txt
```diff
@@ -49,6 +49,11 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL Linux OR ${CMAKE_SYSTEM_NAME} STREQUAL Darwin)
   find_package(Unwind REQUIRED)
 endif()
 
+option(WITH_TCL "Turn ON to compile with Tcl bindings" OFF)
+if(WITH_TCL)
+  find_package(Tcl REQUIRED)
+endif()
+
```
mistachkin: I think this is the "right" way to have an optional feature?

CMakeLists.txt
```diff
@@ -118,6 +123,9 @@ add_subdirectory(protobuf)
 add_subdirectory(schemachange)
 add_subdirectory(sockpool)
 add_subdirectory(sqlite)
+if(WITH_TCL)
+  add_subdirectory(tcl)
+endif()
```
mistachkin: This does seem to work.  It causes the directory to be skipped unless the feature is enabled.

cmake/FindTcl.cmake
mistachkin: This issue should be fixed now.  Tcl support is not included unless enabled.

mistachkin:
From the log files, I do not see a failure of the "yast" test.  It shows as "success"; however, there is a core dump file for the Tcl shell (tclsh).  Not sure where that came from as the logs do not seem to mention Tcl crashing.

mistachkin:
With my most recent check-in on this branch, the "yast" tests now pass within docker (checked manually).

tcl/tclcdb2.c
```diff
@@ -0,0 +1,2379 @@
...
+			   offsetof(cdb2_tm_t, tm_wday)},
+		{  "yday", 0},
+		{    "%d", offsetof(cdb2_client_datetime_t, tm) +
+			   offsetof(cdb2_tm_t, tm_yday)},
+		{ "isdst", 0},
+		{    "%d", offsetof(cdb2_client_datetime_t, tm) +
+			   offsetof(cdb2_tm_t, tm_isdst)},
+		{  "msec", 0},
+		{    "%u", offsetof(cdb2_client_datetime_t, msec)},
+		{"tzname", 0},
```
akshatsikarwar: tzname is optional

tcl/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+add_library(tclcdb2 SHARED
+  tclcdb2.c
+)
+
+set(module tclcdb2)
+set(MODULE TCLCDB2)
+configure_file(${PROJECT_SOURCE_DIR}/mem/mem.h.in mem_tcl.h @ONLY)
```
akshatsikarwar: Don't think we need this.

akshatsikarwar:
LGTM

PR #912, opened 2018-06-01
More functions inline in temp table code, _find_hash, no add_record_int

adizaimi:
More functions inline in temp table code, _find_hash, no add_record_int


adizaimi:
Thanks Lingzhi.

PR #913, opened 2018-06-01
Fix cdb2api tests

mohitkhullar:


adizaimi:
Looks good. Is this true for all platforms, including your laptop?

mohitkhullar:
yes

PR #916, opened 2018-06-04
Temp table insert test

adizaimi:
Temp table insert test.


nirbhayc:
Food for thought - I suggest using some unit testing framework to perform similar kind of work and avoid exposing such functionalities directly to the user.

PR #920, opened 2018-06-08
Compatibility fixes

mponomar:
- Don't use datetime prefix for log messages
- Fix bug where we latch an invalid timezone
- fastinit should trigger auto-analyze

### DIFF
--- a/bb/logmsg.c
+++ b/bb/logmsg.c
@@ -14,7 +14,7 @@
 
 static loglvl level = LOGMSG_WARN;
 static int do_syslog = 0;
-static int do_time = 1;
+static int do_time = 0;
 static int ended_with_newline = 1;
 
 /* from io_override.c */
--- a/bb/logmsg.c
+++ b/bb/logmsg.c
@@ -14,7 +14,7 @@
 
 static loglvl level = LOGMSG_WARN;
 static int do_syslog = 0;
-static int do_time = 0;
+static int do_time = 1;
 static int ended_with_newline = 1;
 
 /* from io_override.c */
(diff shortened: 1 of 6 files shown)


schemachange/sc_fastinit_table.c
```diff
@@ -131,10 +131,21 @@ int do_fastinit(struct ireq *iq, struct schema_change_type *s, tran_type *tran)
...
 {
     int rc = 0;
+    int is_sqlite =  strncmp((s->table), "sqlite_stat", sizeof("sqlite_stat") - 1) == 0;
+
     rc = finalize_alter_table(iq, s, tran);
+
+    if(!s->drop_table && !is_sqlite) {
+        //triger analyze of table -- when fastinit
+        char * tblname = strdup((char*) s->table); //will be freed in auto_analyze_table()
+        autoanalyze_after_fastinit(tblname);
```
ldennis: I prefer to move this analyze logic to bb-plugins/schema_change . For DDLs (especially transactional DDLs), running finalize_fastinit_table doesnt necessarily mean the transaction is going to commit. I think it is better to move analyze code to the point after the schema change finishes.

mponomar: You got it.

bb/logmsg.c
```diff
@@ -14,7 +14,7 @@
 
 static loglvl level = LOGMSG_WARN;
 static int do_syslog = 0;
-static int do_time = 1;
+static int do_time = 0;
```
ldennis: No timestamp by default? What about adding "logmsg notimestamp" to legacy default?

mponomar: You find those helpful?  The default right now depends on whether the database runs with a tty.  Can certainly keep the default and disable with legacy defaults.

ldennis: I kinda like the timestamps personally, but maybe I only want them for debugging purposes. So I guess actually it doesnt really matter

PR #921, opened 2018-06-08
Replay dup

markhannum:
Test and fixes for replay-error which would unintentionally send the same osql-packet multiple times.  Additionally, automatically retry a NON-DURABLE error the same way we retry master-swings.

### DIFF
--- a/berkdb/db/db_iface.c
+++ b/berkdb/db/db_iface.c
@@ -1029,6 +1029,7 @@ __db_get_numpages(dbp, numpages)
 	case DB_UNKNOWN:
 	default:
 		ret = __db_unknown_type(dbp->dbenv, "__db_numpages", dbp->type);
+        abort();
 		goto err;
 	}
 
--- a/berkdb/db/db_iface.c
+++ b/berkdb/db/db_iface.c
@@ -1029,7 +1029,7 @@ __db_get_numpages(dbp, numpages)
 	case DB_UNKNOWN:
 	default:
 		ret = __db_unknown_type(dbp->dbenv, "__db_numpages", dbp->type);
-        abort();
+		abort();
 		goto err;
 	}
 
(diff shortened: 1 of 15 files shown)


db/osqlsqlthr.c
```diff
@@ -215,13 +216,14 @@ int osql_insidx(struct BtCursor *pCur, struct sql_thread *thd, int is_update)
...
-        rc = osql_save_index(pCur, thd, is_update, 0);
+        rc = osql_send_insidx_logic(pCur, thd, NET_OSQL_SOCK_RPL);
         if (rc) {
             logmsg(LOGMSG_ERROR, 
                     "%s:%d %s - failed to cache socksql row rc=%d\n",
                     __FILE__, __LINE__, __func__, rc);
+            return rc;
         }
-        return osql_send_insidx_logic(pCur, thd, NET_OSQL_SOCK_RPL);
+        return osql_save_index(pCur, thd, is_update, 0);
```
adizaimi: This line is no longer needed, as is the the else

db/osqlsqlthr.c
```diff
@@ -1327,11 +1321,11 @@ static int osql_qblobs(struct BtCursor *pCur, struct sql_thread *thd,
             logmsg(LOGMSG_ERROR, 
                     "%s:%d %s - failed to cache socksql row rc=%d\n",
```
adizaimi: Message should be 'failed to send'?

markhannum: yes - tx

berkdb/db/db_iface.c
```diff
@@ -1029,6 +1029,7 @@ __db_get_numpages(dbp, numpages)
 	case DB_UNKNOWN:
 	default:
 		ret = __db_unknown_type(dbp->dbenv, "__db_numpages", dbp->type);
+        abort();
```
ldennis: tab?

ldennis:
This looks AWESOME!! Thanks Mark!!

PR #924, opened 2018-06-11
Remove libreadline6 from Dockerfiles.

riverszhang89:
libreadline-dev will pull in the correct version of libreadline.


adizaimi:
Loogs good.

riverszhang89:
Thank you for the review @adizaimi !!

PR #925, opened 2018-06-11
cdb2api send timestamp of each client query 

adizaimi:
Send client start time and retry count as optional fields of sqlquery, and print them in the event log. The purpose of having this added timestamp is to be able to find out by looking at the server logs how long it took for a query to reach the server -- since a request can be dropped by a server, the api will retry next server and so on, and that time of retrying multiple servers may be significant -- even though the query may execute quickly once accepted by a server, it can look like a slow query to the client. 

Eventlog for an sql will have extra `startlag` field measured in microseconds and `clientretries` is an integer counter: `"startlag": 113,"clientretries": 1` and an event entry would look like this:
```
{"time": 1528694347141980,"type": "sql","sql": "select comdb2_dbname()",
"cnonce": "8323329-12928-141778-1589365389","rows": 1,"host": "client1","
fingerprint": "0fc4c5d02b358b64765c65fb7139ba1a","qtime": 10,
"startlag": 113,"clientretries": 1,"context": ["cdb2sql"],"perf": {"runtime": 85}}
```

### DIFF
--- a/cdb2api/cdb2api.c
+++ b/cdb2api/cdb2api.c
@@ -5318,6 +5318,8 @@ int cdb2_open(cdb2_hndl_tp **handle, const char *dbname, const char *type,
         comdb2_cheapstack_char_array(hndl->stack, MAX_STACK);
 
     if (log_calls) {
+        struct timeval tv;
+        gettimeofday(&tv, NULL);
         fprintf(stderr, "%p> cdb2_open(dbname: \"%s\", type: \"%s\", flags: "
                         "%x) = %d => %p\n",
                 (void *)pthread_self(), dbname, type, hndl->flags, rc, *handle);
--- a/cdb2api/cdb2api.c
+++ b/cdb2api/cdb2api.c
@@ -2276,11 +2276,6 @@ static int cdb2_send_query(cdb2_hndl_tp *hndl, SBUF2 *sb, const char *dbname,
     if (sqlquery.n_set_flags)
         sqlquery.set_flags = &set_commands[n_set_commands_sent];
 
-    if (hndl && hndl->is_retry) {
-        sqlquery.has_retry = 1;
-        sqlquery.retry = hndl->is_retry;
-    }
-
     if (hndl && !(hndl->flags & CDB2_READ_INTRANS_RESULTS) && is_begin) {
         features[n_features] = CDB2_CLIENT_FEATURES__SKIP_ROWS;
         n_features++;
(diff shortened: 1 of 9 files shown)


riverszhang89:
Could you please elaborate a use case for the metric? There're a few issues as well.
1. strnlen isn't available everywhere. strlen is.
2. Use microseconds instead of milliseconds?
3. Testcase please.
4. Port to cdb2jdbc please.

adizaimi:
Thanks for looking:
1. was unintentional, I reverted it
2. client timestamp is in microsecs
3. not sure what the testcase would test presence of the field maybe?
4. will do

nirbhayc:
Why exactly do we need this? If its about logging, why not use the query execution time?

adizaimi:
Nirbhay, updated the description with the reason why. Thanks.

nirbhayc:
```
since a request can be dropped by a server, the api will retry next server and so on, and that time of retrying multiple servers may be significant
```
How significant can it be? Do we have any numbers? I was wondering if it will make more sense to have the hop/retry count instead (or alongside startdelay)? Or otherwise it might get tricky to pin point the exact explanation for the slowness. Just a thought.

mponomar:
I like having the retry count along with start time.

cdb2api/cdb2api.c
```diff
@@ -2348,13 +2348,15 @@ static int cdb2_send_query(cdb2_hndl_tp *hndl, SBUF2 *sb, const char *dbname,
     gettimeofday(&tv, NULL);
     sqlquery.timestampus = ((uint64_t)tv.tv_sec) * 1000000 + tv.tv_usec;
 
+    sqlquery.has_num_retries = 1;
+    sqlquery.num_retries = retries_done;
+
```
mohitkhullar: https://github.com/bloomberg/comdb2/blob/469098e0eea46e555a170d4db65bf08a3664ea79/cdb2api/cdb2api.c#L2281

mohitkhullar: https://github.com/bloomberg/comdb2/blob/469098e0eea46e555a170d4db65bf08a3664ea79/cdb2api/cdb2api.c#L3141

adizaimi: Mohit, i'm a bit confused, I thought there is the api retrying on the same node upon receiving rc as in is_retryable() function. Then there is the retry against a different node upon failure to get response from the server on timeout or if server has no threads available to process (we don't put in queue yet). Do we use same counter for both? 
Is retries_done not the second one?  
```
2301         if (retries_done >= hndl->num_hosts) {                                                                                                    
2302             features[n_features] = CDB2_CLIENT_FEATURES__ALLOW_QUEUING;
2303             n_features++;
2304         }
```
Thanks.

adizaimi: In my testing, if i randomly return CDB2__ERROR_CODE__CHANGENODE from server (in handle_newsql_request()) I don't see retry ever being nonzero because has_is_retry is never set!! Anyway, i see my num_retry often being nonzero, meaning when I change server it reports counter increased. Not sure which one to use yet.

adizaimi:
snap_ha_retry passes in my testing.

PR #928, opened 2018-06-12
Fix cdb2api READ_INTRANS_RESULTS mode.

### LINKED ISSUE #927: CDB2_READ_INTRANS_RESULTS is broken after #804
```java
connection.setAutoCommit(false); /* <-- this is equivalent to "BEGIN" */
executeUpdate("insert into ..."); /* <-- It hangs */
```

riverszhang89:
The pull request changes skip_response to return false, if client has explicitly asked for intrans results.
Variables (skip_feature, skip_rows and etc) are renamed to better match their functionality.

The pull request fixes #927.

### DIFF
--- a/bdb/bdb_osqltrn.c
+++ b/bdb/bdb_osqltrn.c
@@ -288,8 +288,9 @@ bdb_osql_trn_t *bdb_osql_trn_register(bdb_state_type *bdb_state,
                 DB_LSN my_lsn, dur_lsn, arg_lsn;
                 uint32_t my_gen;
 
-                if ((rc = request_durable_lsn_from_master(bdb_state, &dur_lsn.file, 
-                                &dur_lsn.offset, &durable_gen)) != 0) {
+                if ((rc = request_durable_lsn_from_master(
+                         bdb_state, &dur_lsn.file, &dur_lsn.offset,
+                         &durable_gen)) != 0) {
                     *bdberr = BDBERR_NOT_DURABLE;
                     return NULL;
                 }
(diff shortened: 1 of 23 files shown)


mohitkhullar:
looks good

db/sqlinterfaces.c
```diff
@@ -2823,7 +2823,7 @@ static int post_sqlite_processing(struct sqlthdstate *thd,
                 write_response(clnt, RESPONSE_ROW_LAST_DUMMY, 0, 0);
             }
         } else {
-            if (postponed_write) {
+            if (postponed_write && !clnt->send_one_row) {
```
akshatsikarwar: Can you please add a comment here to explain for future readers? I forget already why this is required.

riverszhang89: Will do.

akshatsikarwar: We're not skipping response and have a row to write and now this makes us skip the write?

db/comdb2.c
```diff
@@ -149,7 +149,7 @@ int gbl_recovery_lsn_file = 0;
 int gbl_recovery_lsn_offset = 0;
 int gbl_trace_prepare_errors = 0;
 int gbl_trigger_timepart = 0;
-int gbl_extended_sql_debug_trace = 0;
+int gbl_extended_sql_debug_trace = 1;
```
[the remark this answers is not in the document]
riverszhang89: Good catch thanks!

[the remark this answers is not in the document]
nirbhayc: Perhaps add this as a tunable 'debug.extended_sql_trace' to global tunables (db_tunables.h) and,
* such changes will never go unnoticed  :) (via tunables.test)
* will provide a way to tweak this setting via lrl

akshatsikarwar:
Thank you

bdb/bdb_osqltrn.c
```diff
@@ -274,63 +274,66 @@ bdb_osql_trn_t *bdb_osql_trn_register(bdb_state_type *bdb_state,
...
-        }
-        else if (gbl_rowlocks) {
-            backfill_required = 1;
-            have_lsn = 0;
-        }
-    }
+        else if (is_ha_retry) {
+            if (file || offset) {
+                backfill_required = 1;
+            } else if (gbl_rowlocks) {
```
ldennis: I think this is wrong and `gbl_rowlocks` always need backfill

PR #932, opened 2018-06-14
Writes should not fail even if live_sc_* fails

ldennis:
- writes should not fail even if live_sc_add/upd/del fails
- set db->sc_abort = 1 if live_sc_add/upd/del fails to abort sc
- before finalize (after we have table lock), backout sc if sc_abort==1


adizaimi:
This looks good. Thanks Lingzhi.

PR #933, opened 2018-06-15
Don't log non-zero retries.

mponomar:


adizaimi:
looks good to me. Thanks for fixing. Where did it show up?

mponomar:
It wasn't broken Adi - just wanted 0 to mean disabled (-1 was being returned as a LLONG_MAX).

PR #936, opened 2018-06-18
fix insert lots test race to set the gbl config file

adizaimi:
fix insert lots test race to set the gbl config file


ldennis:
Looks good

adizaimi:
Thanks Lingzhi.

PR #947, opened 2018-06-20
Maintain only a single definitions of hexdump functions

markhannum:
Maintain only a single definition of hexdump, hexdumpbuf, hexdumpdbt, and hexdumpfp in db/db_util.c


markhannum:
The failure doesn't have anything to do with these changes- I will submit a different PR for that

adizaimi:
Thanks Mark. good riddance -- wanted to embark on this since a while ago to cleanup same, but you beat me to it.

akshatsikarwar:
whoag

PR #948, opened 2018-06-21
p4_opfunc overlaps with p4_table causing crashes in writes_remsql_names

markhannum:
p4_opfunc and p4_table were both defined to -20 .. this issue seems to be present in the initial release of opensource comdb2.  It sporadically causes crashes in write_remsql_names when p4free attempts to free non-mallocd memory.


ldennis:
Yea, this looks good.. Thanks Mark! I think this is a bug I introduced when I was upgrading SQLite.. OHISORRYTHX

PR #958, opened 2018-06-26
SP's initial schema lock should work like standard SQL

akshatsikarwar:
Then switch to try_lock version. This fixes race when replicant runs:
1 ddl stmt...
2 exec procedure...

1 returns with early ack
2 fails on try_lock as replication is still applying change

Desired behavior is for 2 to block until schemachange has finished.


markhannum:
Good to merge

PR #960, opened 2018-06-26
add reqinfo to jdbc code

adizaimi:
add reqinfo to jdbc code


adizaimi:
Thanks Rivers for helping with this.

PR #968, opened 2018-07-01
separate accounting of progress for each schemachange

adizaimi:
Now that we can have more than one schemachange (for different tables) running at the same time, we need separate accounting of progress for each schemachange.


adizaimi:
thanks for looking

PR #969, opened 2018-07-02
Check the return code from cdb2_open().

riverszhang89:
Small fix to prevent the database from crashing when comdb2db is unavailable.


mohitkhullar:
looks good

riverszhang89:
Thank you for the review!

PR #973, opened 2018-07-06
PORTED: Remove infinite sleep on unhandled errors

ldennis:
PORTED: Remove infinite sleep on unhandled errors


adizaimi:
is cool

PR #975, opened 2018-07-09
Fix: Legacy defaults are not deferred options.

akshatsikarwar:
These options need to run immediately when encountered. Otherwise, there
is no way to override them.

### DIFF
--- a/db/config.c
+++ b/db/config.c
@@ -54,6 +54,7 @@ extern char **sfuncs;
 extern char **afuncs;
 static int gbl_nogbllrl; /* don't load /bb/bin/comdb2*.lrl */
 
+static int pre_read_option(char *, int);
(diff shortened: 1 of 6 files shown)


akshatsikarwar:
Should address issue #974

db/config.c
```diff
@@ -1330,10 +1320,7 @@ static int read_lrl_option(struct dbenv *dbenv, char *line,
             return -1;
 #endif
     } else if (tokcmp(tok, ltok, "legacy_defaults") == 0) {
-        /* NOOP: Already handled in pre_read_option(); check added here to keep
-         * handle_lrl_tunable() from throwing a warning as legacy_defaults has
-         * not been registered into the new tunables sub-system yet.
-         */
+        read_legacy_defaults(dbenv, options);
```
nirbhayc: "legacy_defaults" being a bit special, I'd add a line of comment explaining why this needs to be handled here (again).

PR #977, opened 2018-07-10
db: Don't override the error code and error message from record.c.

riverszhang89:
These error messages from record.c are more detailed and should be returned to clients.


riverszhang89:
@ldennis reviewed the pull request on bbgh. Thanks!

PR #979, opened 2018-07-10
PORTED: add net clipper from R6

adizaimi:
add net clipper from R6


adizaimi:
We now pass disabled_subnets in robomark.

markhannum:
okay to merge

PR #980, opened 2018-07-10
add test that makes sure we dont allow bad index names

adizaimi:
add test that makes sure we dont allow index names that are blank, that have spaces, or that are more than 53 characters in length.


tests/scindex.test/runit
```diff
@@ -744,6 +744,64 @@ schema
...
+cdb2sql ${CDB2_OPTIONS} $dbnm default - <<"EOF"
+alter table t1 {
+schema
+{
+   int  id
+   blob b1 null=yes
+}
+keys
+{
+   dup "KEY_12345678901234567890123456789012345678901234567890"  = id   # 54 characters -- too long
```
markhannum: It's failing on this because it can't parse the # character - comments in csc2 are //

markhannum: Adi, I make some fixes in the test_index_and_table_names branch

adizaimi: Cool. we might want to add indexname len to this sys tbl:
SELECT * FROM comdb2_limits

PR #987, opened 2018-07-12
dont use bset/btst for osql flags (backwards compatible)

ldennis:
dont use bset/btst for osql flags (backwards compatible)

This also fixes verify_error test on sun and ibm


ldennis:
whoah commit hash has 'c1' and breaks robo

PR #993, opened 2018-07-16
add conv_flags to comdb2_appsock_arg for internal fastdump

ldennis:
- added conv_flags to comdb2_appsock_arg for internal fastdump
- also added APPSOCK_PLUGIN_DESC macro to reduce typing


ldennis:
Thanks @nirbhayc

PR #994, opened 2018-07-17
Make out-of-range key search work for bytearray.

riverszhang89:
To reproduce, run the following queries:

```SQL
create table bytes (a byte(1))
create index bytes_a on bytes(a)
insert into bytes values(x'11')
select * from bytes where a >x'ffff'
```


akshatsikarwar:
Missing `sqls`?

riverszhang89:
Mike reviewed it in bbgh. Thanks!

PR #1001, opened 2018-07-19
Move send_intrans_response into newsql

akshatsikarwar:


akshatsikarwar:
I didn't like how existing code read
```
if (skip_intrans_response)
    // do something
if (send_intrans_results)
    // do something
```

PR #1012, opened 2018-07-20
Fix oldfile list wrap-around bug

ldennis:
- init `list_hd` and `list_tl` to 0
- if the oldfile list wraps around (`list_hd < list_tl`), `oldfile_list_contains` should check from `list_tl` to the end of the list and then from 0 to `list_hd`.
```
i.e.
    list_hd          list_tl
       |                 |
[+++++++-----------------+++++++]
```

### DIFF
--- a/bdb/file.c
+++ b/bdb/file.c
@@ -7304,16 +7304,22 @@ struct unused_file {
 
 #define OF_LIST_MAX 16384
 static struct unused_file of_list[OF_LIST_MAX];
-static int list_hd, list_tl;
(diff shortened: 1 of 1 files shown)


bdb/file.c
```diff
@@ -7304,16 +7304,22 @@ struct unused_file {
 
 #define OF_LIST_MAX 16384
 static struct unused_file of_list[OF_LIST_MAX];
-static int list_hd, list_tl;
+static int list_hd = 0, list_tl = 0;
```
dorinhogea: Static variables.

PR #1015, opened 2018-07-23
fix check for uuid osql for net_osql_master_check

ldennis:
- made osql_nettype_is_uuid inline
- refactor checks for uuid to call osql_nettype_is_uuid


db/osqlcomm.c
```diff
@@ -5450,8 +5450,7 @@ static void net_osql_master_check(void *hndl, void *uptr, char *fromhost,
         return;
     }
 
-    if (usertype > NET_OSQL_UUID_REQUEST_MIN ||
```
ldennis: The bug is here: `||` was wrong and should be `&&`

ldennis:
Thanks @mohitkhullar !!!

PR #1016, opened 2018-07-23
Fix memory leak.

riverszhang89:
bdb_temp_table_find uses DB_DBT_MALLOC now. Hence no need to allocate our key buffer.


riverszhang89:
Mike reviewed it on bbgh. Thanks!

PR #1025, opened 2018-07-27
DDL: On ALTER TABLE, handle field type & length for client tags properly

nirbhayc:
For mis-configured client tags resort to GIGO.


nirbhayc:
Thanks Akshat!

PR #1026, opened 2018-07-28
comdb2_limits: More additions to the limits system table

nirbhayc:
Also removed some duplicate constants.


nirbhayc:
Thank you.

Issue #796, opened 2018-04-03
SP leaking temp tables


ldennis:
Store procedures that create local SQL temporary tables are leaking temp table objects. 

```
cdb2sql ldengdb - <<'EOF'
create procedure foo version 'spfoo' {
local function main()
 local tab = db:table('tmp',{{"id", 'integer'}, {"j", 'integer'}})
 tab:insert({id="1",j="1"})
 local result_set, rc = db:exec("select * from tmp")
 db:emit(result_set)
 return rc
end}$$
put default procedure foo 'spfoo'
exec procedure foo()
EOF
```

Every execution of `exec procedure foo()` leaks temp tables.

ldennis:
Looks like the bug is: https://github.com/bloomberg/comdb2/blame/master/db/sqlglue.c#L7078

Maybe we should do:
```
diff --git a/db/sqlglue.c b/db/sqlglue.c
index 113e4ade3..7e15f8e3b 100644
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -7060,6 +7060,7 @@ sqlite3BtreeCursor_temptable(Btree *pBt,      /* The btree */
                              struct sql_thread *thd)
 {
     int bdberr = 0;
+    struct sqlclntstate *clnt = thd->clnt;
     cur->cursor_class = CURSORCLASS_TEMPTABLE;

     if (iTable < 0 || iTable > pBt->num_temp_tables) {
@@ -7074,8 +7075,10 @@ sqlite3BtreeCursor_temptable(Btree *pBt,      /* The btree */
         return SQLITE_INTERNAL;
     }

-    int num_tables = 0;
-    sqlite3BtreeCreateTable(pBt, &num_tables, BTREE_INTKEY);
+    if (clnt->is_expert) {
+        int num_tables = 0;
+        sqlite3BtreeCreateTable(pBt, &num_tables, BTREE_INTKEY);
+    }

     struct temptable *src = &pBt->temp_tables[iTable];
     cur->tmptable->tbl = src->tbl;
```

ldennis:
@mohitkhullar can you double check?

PR #1029, opened 2018-08-01
Fix temp table leak with SPs

mohitkhullar:
Issue https://github.com/bloomberg/comdb2/issues/796


mohitkhullar:
reviewed in person with lingzhi

ldennis:
Thanks :+1:

PR #1032, opened 2018-08-02
Feature: Decouple installation and data locations

EricHripko:
# Overview
comdb2 build scripts currently force user to have data files next to the binaries. This is not necessarily  a feasible idea for some platforms, for example  [brew](https://brew.sh). Reason being that this package manager installs binaries in `/usr/local/Cellar/$PACKAGE/$VERSION` whilst data remains in `/usr/local/var` (the former location is wiped when software is reinstalled or updated). This behaviour is created because `CMAKE_INSTALL_PREFIX` is passed in as `COMDB2_ROOT` to the compiler. New behaviour decouples these two and thereby provides more flexibility for above-mentioned cases. The code reverts to old behaviour if data location is unspecified, so existing builds are unaffected.

### Type of the change
Feature

### Current behaviour
Unable to keep comdb2 data files in a path different from the install location.

### Expected behaviour
Should be able to keep comdb2 data files in a path different from the install location.

### New behaviour
Allows to keep comdb2 data files in a path different from the install location by passing `COMDB2_ROOT` variable to CMake.


EricHripko:
Judging from
```
/home7/cdb2test/build_root/cdb2test/132053/builddb/cbuild: line 214: pushd: /home7/cdb2test/build_root/R20180802.1/comdb2: No such file or directory
fatal: not a git repository (or any parent up to mount point /)
```
it seems like Bloomberg's build system failed to clone the repository properly.
This in turn failed the builds on all the platforms with the following error:
```
CMake Error: The source directory "/home7/cdb2test/build_root/R20180802.1/comdb2" does not exist.
```

I don't believe that this is an issue with my PR - but rather an issue with CI/CD environment.

riverszhang89:
@EricHripko Thank you for the pull request! Yes - this is not an issue with your PR.

EricHripko:
Additionally, I have downloaded __log.tar.gz__ and it contains (what appears to be) an RSA private/public key pair. Is this expected?

riverszhang89:
It's fine. These are test-generated key pair for our SSL tests :smiley:

EricHripko:
That's a relief - thank you for the clarification :)

riverszhang89:
You're welcome!

EricHripko:
Is there anything I could do to get this merged?

akshatsikarwar:
Thank you!

EricHripko:
Thank you for the merge 👍

PR #1034, opened 2018-08-02
Transactional SC with Constraints

ldennis:
Allow transactional schema changes on multiple tables with foreign key constraints

### DIFF
--- a/db/sql.h
+++ b/db/sql.h
@@ -418,6 +418,15 @@ int clr_high_availability(struct sqlclntstate *);
 uint64_t get_client_starttime(struct sqlclntstate *);
 int get_client_retries(struct sqlclntstate *);
 
+struct clnt_ddl_context {
+    /* Name of the table */
+    char *name;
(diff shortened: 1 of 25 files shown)


db/sql.h
```diff
@@ -418,6 +418,15 @@ int clr_high_availability(struct sqlclntstate *);
 uint64_t get_client_starttime(struct sqlclntstate *);
 int get_client_retries(struct sqlclntstate *);
 
+struct clnt_ddl_context {
+    /* Name of the table */
+    char *name;
```
nirbhayc: Do you plan to support multiple ALTERs to the same table within a trans?

schemachange/sc_drop_table.c
```diff
@@ -69,6 +70,12 @@ int finalize_drop_table(struct ireq *iq, struct schema_change_type *s,
     int rc = 0;
     int bdberr = 0;
 
+    if (db->n_rev_constraints > 0) {
```
ldennis: check ourself

Issue #963, opened 2018-06-29
Cannot insert string into the table via C API (incompatible values from SQL string)


librehat:
Run `CDB2_LOG_CALLS=1 ./test_insert `, then it prints errors:

```
cdb2_socket_pool_get(comdb2/testdb/local/newsql/random_room,1): fd=-1
cdb2_socket_pool_get(comdb2/testdb/local/newsql/random_room,1): fd=-1
td 1224217216 parse_dbresponse:2981
td 1224217216 parse_dbresponse:3011, 0) host=localhost.localdomain(19000)*
0x7f0648f81280> cdb2_open(dbname: "testdb", type: "local", flags: 14) = 0 => 0x1746ad0
0x7f0648f81280> cdb2_bind_param(0x1746ad0, "val", CDB2_CSTRING, 0x400fb1, 6) = 0
cdb2_socket_pool_get(comdb2/testdb/local/newsql/random_room,1): fd=-1
td 0x7f0648f81280 cdb2_send_query line 2380
0x7f0648f81280> cdb2_run_statement(0x1746ad0, "INSERT INTO t0 (a) VALUES (@val);") = 113
0x7f0648f81280> cdb2_errstr(0x1746ad0) = "incompatible values from SQL string of length 6 to bcstring field 'a' for table 't0'"
Failed to execute statement "INSERT INTO t0 (a) VALUES (@val);": incompatible values from SQL string of length 6 to bcstring field 'a' for table 't0'
```

Environment: Latest `comdb2` master branch code, built with `gcc 6.4.0`, 64-bit Linux

To reproduce, create a table `t0`: 
```
CREATE TABLE t0 {
 schema {
    cstring a[20]
  }
}
```

This is the C program code

```c
#include <cdb2api.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdio.h>
#include <string.h>
#include <strings.h>
#include <time.h>

void print_datetime(const cdb2_client_datetime_t* src)
{
    printf("%d-%d-%dT%d:%d:%d (%s)",
           src->tm.tm_year + 1900, src->tm.tm_mon + 1, src->tm.tm_mday,
           src->tm.tm_hour, src->tm.tm_min, src->tm.tm_sec,
           src->tzname);
}

void print_cdb2_value(const void *src, int type)
{
    switch (type) {
    case CDB2_INTEGER:
        printf("%ld", *(int64_t*)src);
        break;
    case CDB2_REAL:
        printf("%e", *(double*)src);
[...]

PR #1036, opened 2018-08-03
 Check for the terminating '\0' in bound string to adjust the supplied length

nirbhayc:
This is needed to preserve the backwards compatibility with R6, which worked
even when the bound string's length included the terminating '\0'.


nirbhayc:
https://github.com/bloomberg/comdb2/issues/963

mohitkhullar:
the code change looks good, however the test case isn't added to regression testing

nirbhayc:
@mohitkhullar the included test runs via cdb2bind.test. Thanks for the review and help pinning down the issue!

PR #1037, opened 2018-08-03
DDL: Allow whitespaces between type and its size

nirbhayc:
For instance, the following command used to fail:
CREATE TABLE t1(c VARCHAR (10))$$


nirbhayc:
Thanks @ldennis !

PR #1040, opened 2018-08-05
Feature: Allow running cdb2sockpool in foreground

EricHripko:
# Overview
`cdb2sockpool` demonises itself by default and has no option to run in foreground. A better approach (according to *nix best practices) would be to offer both foreground and background options to the user. This is something that many daemons and other COMDB2 tools do already (e.g., `pmux` via `-f` flag). New behaviour follows this example to implement a similar flag for `cdb2sockpool` and offers user more control. If flag is unspecified, startup procedure follows the existing behaviour, so current users are unaffected. Both startup modes were tested and successfully accepted socket donations and fulfilled socket requests.

### Type of the change
Feature

### Current behaviour
Unable to run `cdb2sockpool` in foreground.

### New behaviour
Allows to run `cdb2sockpool` in foreground by passing in `-f` flag. This makes the daemon easier to integrate with a wider range of supervisors/init systems. Additionally, it makes troubleshooting easier in the development environment.

### DIFF
--- a/tools/cdb2sockpool/cdb2sockpool.c
+++ b/tools/cdb2sockpool/cdb2sockpool.c
@@ -93,6 +93,7 @@ struct client {
 
 static char unix_bind_path[128] = SOCKPOOL_SOCKET_NAME;
 static char msgtrap_fifo[128] = "/tmp/msgtrap.sockpool";
+static int foreground_mode = 0;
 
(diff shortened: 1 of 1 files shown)


EricHripko:
Thank you for merging my request 👍

PR #1041, opened 2018-08-05
Bug: Fix the typo in client API config initialisation

EricHripko:
# Overview
The [documentation](https://bloomberg.github.io/comdb2/clients.html#dnssuffix) indicates that `dnssuffix` is the configuration key for specifying the location of COMDB2 databases via DNS. However, the client code is currently looking for `dnssufix` key instead (seems to be a minor typo). To preserve the current behaviour, the client code was modified to support both correct and wrong spellings.

### Type of the change
Bug

### Current behaviour
Unable to provide DNS suffix via the documented configuration key.

### New behaviour
Allows providing DNS suffix with both documented and misspelled versions.


EricHripko:
Thank you for merging my request 👍

PR #1052, opened 2018-08-13
add option to have changes with tid

mohitkhullar:
add option to have consumer changes with transaction id


mohitkhullar:
reviewed in bbgithub

PR #1057, opened 2018-08-15
Return lua error rather than crash if user attempt to emit nothing

markhannum:
Discovered this when I accidentally wrote a buggy lua sp- if you attempt to emit a select from a table with no rows the database aborts on an assert.  This branch removes the assert and returns an error message.

### DIFF
--- a/lua/sp.c
+++ b/lua/sp.c
@@ -4790,7 +4790,9 @@ static int l_send_back_row(Lua lua, sqlite3_stmt *stmt, int nargs)
 {
     int rc = 0;
     SP sp = getsp(lua);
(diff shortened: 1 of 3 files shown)


akshatsikarwar:
Perfer to use `luaL_argcheck` so we get std error message.

akshatsikarwar:
`luaL_argcheck` will have additional effect of terminating sp execution. `luabb_error` does not stop the run, so unless programmer checks return code from emit (exactly zero do because its tedious) it will result in silent failure.

markhannum:
Cool- thanks guys

PR #1069, opened 2018-08-20
schemachange: Don't double close an SBUF object.

riverszhang89:
When the master downgrades while performing a schema change, it sends an error to the schema change client and closes its SBUF. It then closes it again as part of the regular cleanup in free_schema_change_type().


riverszhang89:
Thank you for the review.

PR #1070, opened 2018-08-20
adjust tblvers and gbl open gen on bulkimport

adizaimi:
adjust tblvers and gbl open gen on bulkimport

/cbuild-compile-only


db/tag.c
```diff
@@ -7083,8 +7084,6 @@ int reload_after_bulkimport(struct dbtable *db, tran_type *tran)
     }
     db->tableversion = table_version_select(db, NULL);
     update_dbstore(db);
-    create_sqlmaster_records(tran);
-    create_sqlite_master();
```
ldennis: bb-plugin `bulkimport_switch_files` may still depend on this function calling create sql master (unless you also moved these out there).

PR #1072, opened 2018-08-21
Make the cdb2api functions, e.g. cdb2_column_size, etc, more robust in the face of responses that have no inner value.

mistachkin:
These changes make several of the public cdb2api functions more robust when the inner value for the first and/or last response is null.  For example, when using the cdb2_column_size() function after iterating over the result set, it should return -1, not cause a segfault.


markhannum:
Okay to merge

PR #1074, opened 2018-08-22
Add node uptime func

adizaimi:
add comdb2_node_uptime() and comdb2_node_starttime() sql functions


riverszhang89:
`ps` can tell us these.

nirbhayc:
Won't work with `nowatch`.

adizaimi:
While ps can show you that and more info, you'd need to login onto the box to get that and, besides the inconvenience of doing that, you might not even have shell access onto a box.

adizaimi:
It does indeed work even with nowatch because nowatch does not disable the thread from running, but it skips over doing work if gbl_nowatch is set.

riverszhang89:
Fair. But please avoid using node. We even changed comdb2_node() to comdb2_host().

adizaimi:
ok, no need for node in the name. thanks.

sqlite/ext/misc/completion.c
```diff
@@ -338,7 +338,7 @@ static int completionNext(sqlite3_vtab_cursor *cur){
         break;
       }
       case COMPLETION_FUNCTIONS: {
-        static char *cfuncs[] = {"comdb2_version()", "table_version()", "partition_info()", "comdb2_host()", "comdb2_port()", "comdb2_dbname()", "comdb2_prevquerycost()", "sys.cmd.send()"};
+        static char *cfuncs[] = {"comdb2_version()", "table_version()", "partition_info()", "comdb2_host()", "comdb2_port()", "comdb2_dbname()", "comdb2_prevquerycost()", "comdb2_uptime()", "comdb2_starttime()", "sys.cmd.send()"};
```
nirbhayc: Please break this into multiple lines.

nirbhayc:
@adizaimi: Please also update the docs.

PR #1081, opened 2018-08-23
Docs: Fix set commands in certification-based auth #skip-build

nirbhayc:
.


nirbhayc:
thanks @riverszhang89 !

PR #1082, opened 2018-08-23
 dont run tests that require ssl enabled if SKIPSSL is set to 1

adizaimi:
* dont run tests that require ssl enabled, if SKIPSSL is set to 1
* add a higher level of cleanup by CLEANUPDBDIR=2 when test is successful
* dont run tests that rely on tclcdb2 if it is not enabled

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -14,7 +14,7 @@ include $(TESTSROOTDIR)/Makefile.common
 DISABLED_TESTS:=$(shell awk '/^DISABLED/{f=1;next};/<END>/{f=0}f' TODO | cut -d' ' -f1)
 
 ifeq ($(SKIPSSL), 1)
-SKIPS:=$(shell grep ssl *.test/lrl*  | grep 'ssl_client_mode VERIFY' | cut -f1 -d'/'| sort -u)
+SKIPS:=$(shell grep ssl *.test/lrl*  | grep 'ssl_client_mode VERIFY' | cut -f1 -d'/' | sort -u)
 DISABLED_TESTS:="${DISABLED_TESTS} ${SKIPS}"
 endif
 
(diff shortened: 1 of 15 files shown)


riverszhang89:
I'm not sure if we should make a change for just that. I disable SSL hence these SSL tests fail - this makes perfect sense to me.

riverszhang89:
Okay. I'm sold.

tests/Makefile
```diff
@@ -11,7 +11,20 @@ include $(TESTSROOTDIR)/Makefile.common
 
 
 # List of disabled tests
-export DISABLED_TESTS:=$(shell awk '/^DISABLED/{f=1;next};/<END>/{f=0}f' TODO | cut -d' ' -f1)
+DISABLED_TESTS:=$(shell awk '/^DISABLED/{f=1;next};/<END>/{f=0}f' TODO | cut -d' ' -f1)
+
+ifeq ($(SKIPSSL), 1)
+SKIPS:=$(shell grep ssl *.test/lrl*  | grep 'ssl_client_mode VERIFY' | cut -f1 -d'/' | sort -u)
```
nirbhayc: Just grep for 'ssl_client_mode'? presence of this tunable implies an ssl test.

adizaimi: many tests have 'ssl_client_mode OPTIONAL' so that wont work.

nirbhayc: What's OPTIONAL mode? I don't see it in our docs: https://bloomberg.github.io/comdb2/ssl.html. 

Anyway, what about tests with other modes like REQUIRE, VERIFY_HOSTNAME? They won't get skipped with this logic then.

adizaimi: IIRC, optional is the mode that if ssl cert exists, it will be used, otherwise we will fallback to non ssl; i might missrememer.
We don't currently have any tests with REQUIRE, and verify_hostname starts with verify so it is correctly skipped.

PR #1084, opened 2018-08-24
refactor a bit watchog thread, remove superflous function

adizaimi:
* Dont create threads every second in watchdog thread, rather
  at a more reasonable frequency, once in 10 seconds
* Use calloc instead of malloc to actually write to allocated area
  since linux may return non null even when out of memory
* Remove unneeded function comdb2_uptime because its value can be
  gotten simply via 'select cast (now() - comdb2_starttime() as INT)'
* Cleanup a bit better after tests in cluster nodes, enable more cleanup

### DIFF
--- a/db/watchdog.c
+++ b/db/watchdog.c
@@ -265,18 +265,18 @@ static void *watchdog_thread(void *arg)
 
                     /* try to create a thread */
                     rc = pthread_create(&dummy_tid, &gbl_pthread_joinable_attr,
-                            dummy_thread, thedb);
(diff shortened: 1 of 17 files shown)


nirbhayc:
> Remove unneeded function comdb2_uptime because its value can be
> gotten simply via 'select cast (now() - comdb2_starttime() as INT)'

You shouldn't have removed this. I'd prefer ```comdb2_uptime()``` anytime over ```cast (now() - comdb2_starttime() as INT)```.

Issue #1080, opened 2018-08-22
upsert.test assertion failure


nirbhayc:
```
(gdb) bt
#0  0x00007fcf59b951d7 in raise () from /lib64/libc.so.6
#1  0x00007fcf59b968c8 in abort () from /lib64/libc.so.6
#2  0x00007fcf59b8e146 in __assert_fail_base () from /lib64/libc.so.6
#3  0x00007fcf59b8e1f2 in __assert_fail () from /lib64/libc.so.6
#4  0x00000000006d547d in bdb_prim_updkey_genid_int (bdb_state=0x7fcf4d3bd830, tran=0x77fb028, key=0x7fcf4af78cb0, keylen=5, ixnum=0, oldgenid=182518930210816, genid=186916976721920, dta=0x0, dtalen=0, isnull=0,
    bdberr=0x7fcf4af78988) at /home/nchoubey/mbig5773/comdb2-upsert-assert-fix/bdb/upd.c:302
#5  0x00000000006d5703 in bdb_prim_updkey_genid (bdb_state=0x7fcf4d3bd830, tran=0x77fb028, key=0x7fcf4af78cb0, keylen=5, ixnum=0, oldgenid=186916976721920, genid=182518930210816, dta=0x0, dtalen=0, isnull=0, bdberr=0x7fcf4af78988)
    at /home/nchoubey/mbig5773/comdb2-upsert-assert-fix/bdb/upd.c:382
#6  0x00000000004cc1aa in ix_upd_key (iq=0x7fcf4b036048, trans=0x77fb028, key=0x7fcf4af78cb0, keylen=5, ixnum=0, oldgenid=182518930210816, genid=186916976721920, dta=0x0, dtalen=0, isnull=0)
    at /home/nchoubey/mbig5773/comdb2-upsert-assert-fix/db/glue.c:1003
#7  0x000000000052ec71 in upd_record (iq=0x7fcf4b036048, trans=0x77fb028, primkey=0x0, rrn=2, vgenid=182518930210816, p_buf_tag_name=0x7fcf4af7f740 ".ONDISK", p_buf_tag_name_end=0x7fcf4af7f748 "\200\070\212\a",
[...]

PR #1092, opened 2018-08-30
Fix an incorrect assert

nirbhayc:
.


nirbhayc:
https://github.com/bloomberg/comdb2/issues/1080

nirbhayc:
@riverszhang89 looking into mem_tracker test failure.

PR #1093, opened 2018-09-01
Fix comdb2 to correctly use hardcoded port in the lrl file

adizaimi:
When we provide a port number in the lrl file via
    `port localhost 12345`
and when a pmux process is running, comdb2 will contact pmux
and try to register with configured port, but if such port is out of range
then pmux will fail to register and comdb2 will try to register again
with a new port (in heartbeat_check_thread()) and that will result in
comdb2 calling abort. This fix includes:

* pmux should include the end value of the port range (was off by one)
* pmux should allow to register even when port requested is outside its port range
* pmux will honor latest use port msg and deregister previous user
* when port hardcoded in lrl, comdb2 should send `use port` to pmux (rather than `reg port` which can result in a totally new port issued -- ex. when pmux restarts with empty cache)
* add more test cases to pmux_hello.test


tools/pmux/pmux.cpp
```diff
@@ -660,11 +661,15 @@ static int run_cmd(struct pollfd &fd, std::vector<struct pollfd> &fds, char *in,
         for (auto it : active_services) {
             conn_printf(c, "%s\n", it.c_str());
         }
-    } else if (strcmp(cmd, "exit") == 0) {
-        if (c.writable) {
-            return 1;
-        } else {
-            disallowed_write(c, cmd);
+    } else if (strcmp(cmd, "active") == 0) {
```
riverszhang89: Already have "active" command (a few lines above).

tools/pmux/pmux.cpp
```diff
@@ -660,11 +661,15 @@ static int run_cmd(struct pollfd &fd, std::vector<struct pollfd> &fds, char *in,
...
+        std::lock_guard<std::mutex> l(active_services_mutex);
+        conn_printf(c, "%d\n", active_services.size());
+        for (auto it : active_services) {
+            conn_printf(c, "%s\n", it.c_str());
+        }
+    } else if (strcmp(cmd, "range") == 0) {
```
riverszhang89: Update help message to include the new subcommand

util/portmuxusr.c
```diff
@@ -58,7 +58,6 @@ int gbl_pmux_route_enabled = 1;
 #define PORTMUX_SET_NO_LINGER() 1
 #define PORTMUX_USE_ECHO_GET() 1
 #define PORTMUX_USE_REFACTORED_GET() 1
-#define PORTMUX_VALIDATE_NAMES() 1
```
riverszhang89: Keep the macro. Just change it to 0.

util/portmuxusr.c
```diff
@@ -160,18 +159,17 @@ static int remaining_timeoutms(int startms, int timeoutms)
...
 
 static int portmux_cmd(const char *cmd, const char *app, const char *service,
-                       const char *instance, char *post)
+                       const char *instance, const char *post)
 {
     char name[strlen(app) + strlen(service) + strlen(instance) + 3 +
               MAX_DBNAME_LENGTH];
     char res[32];
     SBUF2 *ss;
-    int rc, fd, port;
```
riverszhang89: I really don't see any problems with the C90 style. Let's not change it.

tools/pmux/pmux.cpp
adizaimi: Yep. Thanks.

tools/pmux/pmux.cpp
adizaimi: Copy paste error. Thanks.

util/portmuxusr.c
adizaimi: Dont have a clue why i removed this. Putting back.

util/portmuxusr.c
adizaimi: Tried to make this behave similarly in many of the functions below. There is no gain here, I agree.

adizaimi:
Addressed all the comments.

PR #1096, opened 2018-09-06
Fail with verify error if generation has changed

markhannum:
This test devised to fix a socksql_master_swings test error that we see usually after a week or more of continuous testing.  Lingzhi analyzed and explained the issue: it occurs an osql stream refers to genids which have been rolled back by a new master.  In the best case, the  genids won't be found against the new master, and the transaction will return with a verify error.  In the worst case, the new master could have generated the same set of genids which might refer to different records.

The fix for this is straightforward- record the cluster's generation at the beginning of a transaction on the replicant.  If the generation has changed by the time that transaction is executed on the master, automatically fail with a verify error.


akshatsikarwar:
Perhaps `osql_send_startgen` should be default on and disabled by legacy_defaults so R6 can run. After R7 transition is complete, it can be removed from legacy_defaults.

akshatsikarwar:
Or enable this iff genid48 is enabled?

PR #1102, opened 2018-09-07
comdb2makecluster: tool to create a cluster and start it.

adizaimi:
Add comdb2makecluster which is a tool to create a db, copy it to cluster nodes
and start it.
Tool will create DB in current directory or use --dir to specify DB dir.

Example run:

```
bash> comdb2makecluster mydb node1 node2
mydb: creating
mydb: copying to cluster
mydb: starting
mydb: waiting until ready
mydb: Cluster is running with 2 nodes. You can query it with:
   cdb2sql --cdb2cfg mydb/comdb2.cfg mydb 'select 1'
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -140,6 +140,7 @@ endif()
 add_subdirectory(tools)
 add_subdirectory(plugins)
 add_subdirectory(db)
+add_subdirectory(contrib/comdb2makecluster)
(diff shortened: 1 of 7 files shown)


adizaimi:
This is same as https://github.com/bloomberg/comdb2/pull/41 which was closed because it was unable to run tests.

PR #1104, opened 2018-09-08
add a test for correctness of event logging of sql; set sql after done

adizaimi:
add a test for correctness of event logging of sql; set sql after done


adizaimi:
Found issue with cached statements which use last querie's fingerprint.

PR #1107, opened 2018-09-10
nodepick fix for BB env

dorinhogea:
fix R7 remsql to run on BB dev

### DIFF
--- a/db/fdb_boots.c
+++ b/db/fdb_boots.c
@@ -356,14 +356,16 @@ static int _discover_remote_db_nodes(const char *dbname, const char *class,
     char *node;
     cdb2_hndl_tp *db;
     const char *comdb2dbname = "comdb2db";
+    const char *comdb2dbclass = "prod";
 
     /* NOTE: test is dev */
     if (strncasecmp(class, "test", 4) == 0) {
         class = "dev";
         comdb2dbname = "comdb3db";
+        comdb2dbclass = "dev";
     }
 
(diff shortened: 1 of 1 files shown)


db/fdb_boots.c
```diff
@@ -355,13 +355,15 @@ static int _discover_remote_db_nodes(const char *dbname, const char *class,
...
+    const char *comdb2dbname = "comdb2db";
 
     /* NOTE: test is dev */
     if (strncasecmp(class, "test", 4) == 0) {
         class = "dev";
+        comdb2dbname = "comdb3db";
     }
 
-    rc = cdb2_open(&db, "comdb2db", "default", 0);
+    rc = cdb2_open(&db, comdb2dbname, "default", 0);
```
mohitkhullar: This won't work on beta and alpha.
instead of default, you can put class with its default as prod, as comdb3db exists only on dev.

dorinhogea: Why is "default" not working on alpha and beta CDB2 machines?

dorinhogea: I would have preferred to have name mapping in cdb2api, this way server could say comdb2db and talk to whatever db is behind (comdb3db in BB dev case).

dorinhogea:
This would work, until we decide to better encapsulate stage details inside a BB plugin.

mohitkhullar:
this should work

dorinhogea:
Thanks for review!

PR #1111, opened 2018-09-11
comdb2_systables: system table to list all system tables

nirbhayc:
* Also added '@ls systables' to cdb2sql as shortcut


sqlite/ext/comdb2/systables.c
```diff
@@ -0,0 +1,175 @@
+/*
```
riverszhang89: I prefer ezsystable (`create_system_table()`) interface because it makes creating a new systable a lot simpler:

https://github.com/bloomberg/comdb2/blob/master/sqlite/ext/comdb2/cluster.c

riverszhang89: Can you iterate the hashtable and store the elements in the array in `create_system_table()`'s init callback (the 3rd argument)?

riverszhang89: For instance, `bdb_fill_cluster_info()` converts netinfo to an array which eztable expects.

riverszhang89: Okay then.

nirbhayc:
Thanks for the review @riverszhang89 !

PR #1112, opened 2018-09-12
fix some compiler warnings pt1

adizaimi:
fix some compiler warnings pt1


mem/mem.c
```diff
@@ -384,6 +377,7 @@ int comdb2ma_init(size_t init_sz, size_t max_cap)
              * 3. If a thread is not allowed to create a mspace, it will
              *    be uniformly assigned one (to minimize contention).
              */
+            char *env_mspace_max;
```
riverszhang89: What compiler warning do you see here?

mem/mem.c
```diff
@@ -2090,6 +2082,13 @@ static void pfx_ctrace(const char *format, ...)
 
 #if !defined(USE_SYS_ALLOC) && !defined(COMDB2MA_OMIT_BMEM)
 /* COMDB2 BLOCKING MEMORY ALLOCATOR { */
+
+/* the pthread key is shared amongst all blocking allocators */
```
riverszhang89: I specifically placed these lines in "static code" section. Is this causing a compiler warning?

mem/mem.c
adizaimi: it's an unused variable warning -- only used if PER_THREAD_MALLOC is defined.

riverszhang89: Gotcha.

mem/mem.c
adizaimi: Yes, i put them back where you had them, and surrounded them with appropriate ifdef.
Thanks

PR #1113, opened 2018-09-12
Don't clobber net_send rcode

ldennis:
.


adizaimi:
I assume masterbranch always fails the new test, right?  

Just a suggestion: we can do two checkins, one with the testcase (so we clearly see masterbranch failing) and another one with the fix. (Not sure if worth doing so now with this PR though).

ldennis:
I did it manually already

PR #1118, opened 2018-09-14
Logical-redo-based Live SC

ldennis:
Unlike existing schema change scheme which propagates any ongoing writes (applied to the old btrees) inline to the new btrees synchronously, the redo-based schema change scheme uses logical logs to capture table changes and asynchronously applies them to new btrees.


ldennis:
Mark approved it.
`cdb2test   Dec 19 11:57:50 2018  in-comdb2db logical_livesc.R20181219.2`

PR #1119, opened 2018-09-14
Fix rep all req

markhannum:
Fixing a bug introduced by my work on the fast-catchup project.  The main fix is to remove the code which allocates an extraordinarily large cursor-buffer.  Akshat brought this to my attention (thank you).

### DIFF
--- a/db/db_tunables.c
+++ b/db/db_tunables.c
@@ -79,6 +79,7 @@ extern int gbl_max_lua_instructions;
 extern int gbl_max_sqlcache;
 extern int __gbl_max_mpalloc_sleeptime;
 extern int gbl_mem_nice;
+extern int gbl_verbose_net;
 extern int gbl_netbufsz;
 extern int gbl_net_lmt_upd_incoherent_nodes;
 extern int gbl_net_max_mem;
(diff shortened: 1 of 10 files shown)


db/db_tunables.c
```diff
@@ -79,6 +79,7 @@ extern int gbl_max_lua_instructions;
 extern int gbl_max_sqlcache;
 extern int __gbl_max_mpalloc_sleeptime;
 extern int gbl_mem_nice;
+extern int gbl_verbose_net;
 extern int gbl_netbufsz;
```
mistachkin: Is the "verbose_net" tunable actually checked somewhere?  I don't see where it is used in this pull request.

akshatsikarwar:
I think initial fix (`replace __log_cursor_complete`)  was the correct one -- it was missing the net patch. I've been testing net + this and system is working very well.

akshatsikarwar:
And in fact changing only `REP_LOG_REQ` was sufficient to fix issue for me. I'll submit new PR to get those changes into master and also separate PR for `--admin`. Can come back to this one if changes aren't sufficient to resolve issue.

PR #1122, opened 2018-09-15
debugprint number of rows written on master node as part of a transaction

adizaimi:
Print under debug flag, on commit, the number of rows written on master node as part of a transaction. Use nops which counts the number of writes done via a osql stream.


adizaimi:
Check if counter is reset on restarting after a verify error.

PR #1123, opened 2018-09-17
db: Fix spurious poke timeout.

riverszhang89:
`osql_chkboard_wait_commitrc` may incorrectly time out an OSQL session when the last poked time is extremely close to the next second.


markhannum:
Yes - looks great - okay to merge

riverszhang89:

PR #1129, opened 2018-09-19
open_dbs: retry on deadlock

ldennis:
.


adizaimi:
Looks good. Thanks Lingzhi.

PR #1134, opened 2018-09-21
sc_logical_tran can be null

markhannum:
This fixes a crash in toblock that can occur for a tranddl transaction that fails to commit due to a master swing.

### DIFF
--- a/db/toblock.c
+++ b/db/toblock.c
@@ -5604,7 +5604,8 @@ add_blkseq:
                 if (iq->tranddl) {
                     if (backed_out) {
                         assert(trans == NULL);
-                        bdb_ltran_put_schema_lock(iq->sc_logical_tran);
+                        if (iq->sc_logical_tran)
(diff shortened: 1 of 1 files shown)


db/toblock.c
```diff
@@ -5604,7 +5604,8 @@ static int toblock_main_int(struct javasp_trans_state *javasp_trans_handle,
                 if (iq->tranddl) {
                     if (backed_out) {
                         assert(trans == NULL);
-                        bdb_ltran_put_schema_lock(iq->sc_logical_tran);
+                        if (iq->sc_logical_tran)
+                            bdb_ltran_put_schema_lock(iq->sc_logical_tran);
```
ldennis: Under what circumstances is sc_logical_tran NULL? If master swings, I think we should `goto cleanup` instead of adding blkseq.. Also later at line 5627, we are trying to commit sc_logical_tran and if sc_logical_tran is NULL we have a bigger problem.. I dont think we should reach here at all for master swings

PR #1138, opened 2018-09-24
bdb: Check for null in temptable creation.

riverszhang89:
Fail the request gracefully on OOM.


riverszhang89:
Thank you @ldennis

PR #1139, opened 2018-09-25
ddl schemachange: move sending tableversion in OSQL_SCHEMACHANGE

adizaimi:
instead of sending table version via usedb, this checkin sends the tableversion for a schemachange
in the OSQL_SCHEMACHANGE osql packet.


ldennis:
Lets make sure we dont break cbuild. Looks good otherwise

PR #1140, opened 2018-09-25
Fix getservbyname MacOS linking issue

pkryger:
MacOS does not have `getservbyname_r` function. The `getservbyname` is
reentrant by default on MacOS. From `man getservbyname`:

> These functions use a thread-specific data storage; if the data is needed for future use, it should be copied before any sub-sequent calls overwrite it.

However due to `_LINUX_SOURCE` being set while
compiling on MacOS it fails to link with:

```
Undefined symbols for architecture x86_64:
  "_getservbyname_r", referenced from:
      _comdb2_getservbyname in libutil.a(bb_oscompat.c.o)
ld: symbol(s) not found for architecture x86_64
```

This commit fixes this by using `getservbyname` in a presence of `_DARWIN_C_SOURCE`.

bug fix
The `comdb2`. does not link on MacOS. The `comdb2` should link on MacOS.
```
mkdir bld && cd bld && cmake .. & make -j
```

### DIFF
--- a/util/bb_oscompat.c
+++ b/util/bb_oscompat.c
@@ -66,8 +66,7 @@ void comdb2_getservbyname(const char *name, const char *proto, short *port)
 {
     struct servent result_buf, *result = NULL;
     char buf[1024];
-#   if defined(_DARWIN_C_SOURCE) // Should be first, as _LINUX_SOURCE is
-                                 // also defined on Mac.
+#   if defined(__APPLE__) // Should be first, as _LINUX_SOURCE is also defined.
     result = getservbyname(name, proto);
 #   elif defined(_LINUX_SOURCE)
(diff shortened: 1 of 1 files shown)


util/bb_oscompat.c
```diff
@@ -66,7 +66,10 @@ void comdb2_getservbyname(const char *name, const char *proto, short *port)
 {
     struct servent result_buf, *result = NULL;
     char buf[1024];
-#   if defined(_LINUX_SOURCE)
+#   if defined(_DARWIN_C_SOURCE) // Should be first, as _LINUX_SOURCE is
```
riverszhang89: Thanks for the pull request! _DARWIN_C_SOURCE is optional on MacOS so it can't be used to reliably detect the OS. Can you please change it to __APPLE__?

PR #1143, opened 2018-09-26
Pass tableversion in sc structure, remove from iq

adizaimi:
Pass tableversion in sc structure, remove from iq.


adizaimi:
This is in #1139 so closing this since will be checked in later in that PR.

PR #1146, opened 2018-09-27
Add an entry to verify_error test, debug for stepper.c, debug level for tests, missing osql type.

adizaimi:
* Add an entry to verify_error test
* Debug env var to tests/tools/stepper.c to print stmts being run
* Have tests run in debug log level for easier debugging on error
* Add printing missing OSQL type


nirbhayc:
A side note: I’d push logically unrelated changes into separate PRs. It gives us a clean history and makes reviewing easier.

adizaimi:
> Why change default log level to DEBUG?
Its only for tests -- not for a normal db not created by test setup. This way when debugging an isssue or test failure we can easily identify issue rather than struggle and have to remember to turn this on manually. On successful tests we cleanup directories so it does not matter how much is generated by test suite.

adizaimi:
I will enter separate prs for distinct pieces in future. Thanks.

PR #1147, opened 2018-09-28
berkdb: Check ALL hash buckets if not enough space for a new page in the buffer pool.

riverszhang89:
The current logic always skips 1 hash bucket.
If the bucket is the only one containing writable pages, we will mistakenly abort.
This happens occasionally for temptable operations because a temptable has a fairly small buffer pool.


riverszhang89:
@markhannum and @mponomar have reviewed it. Thanks!

PR #1149, opened 2018-09-29
SOSQL: dont lock table on every write in add record

adizaimi:
For SQL writes on master we lock table when we get usedb, which locks table. We don't need to lock table on add_record, upd_record, and del_record if we come from osql.

### DIFF
--- a/db/record.c
+++ b/db/record.c
@@ -173,7 +173,6 @@ int add_record(struct ireq *iq, void *trans, const uint8_t *p_buf_tag_name,
                int rec_flags)
 {
     char tag[MAXTAGLEN + 1];
-    int is_od_tag;
     int rc = 0;
     int retrc = 0;
     int expected_dat_len;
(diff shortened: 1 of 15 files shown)


db/record.c
```diff
@@ -265,25 +264,30 @@ int add_record(struct ireq *iq, void *trans, const uint8_t *p_buf_tag_name,
...
-                reqprintf(iq, "LOCK TABLE READ DEADLOCK");
-            retrc = RC_INTERNAL_RETRY;
-            ERR;
-        } else if (rc) {
-            if (iq->debug)
-                reqprintf(iq, "LOCK TABLE READ ERROR: %d", rc);
-            *opfailcode = OP_FAILED_INTERNAL;
-            retrc = ERR_INTERNAL;
-            ERR;
+        if(!iq->have_blkseq) {
```
ldennis: Why is have_blkseq relevant? How does `if (!iq->is_sorese)` sound?

adizaimi: Hmm, that might work too, i didn't look. It looked like have_blkseq was set for sql and looked like what we wanted; sorese might do the same, i'll check.

db/record.c
```diff
@@ -265,25 +264,30 @@ int add_record(struct ireq *iq, void *trans, const uint8_t *p_buf_tag_name,
...
-                reqprintf(iq, "LOCK TABLE READ DEADLOCK");
-            retrc = RC_INTERNAL_RETRY;
-            ERR;
-        } else if (rc) {
-            if (iq->debug)
-                reqprintf(iq, "LOCK TABLE READ ERROR: %d", rc);
-            *opfailcode = OP_FAILED_INTERNAL;
-            retrc = ERR_INTERNAL;
-            ERR;
+        if(!iq->is_sorese) {
```
ldennis: On second thought, this may not be safe for cascading delete/update? Can you double check?

ldennis: add_record is fine i guess.. I meant for upd_record and del_record

adizaimi: Will add flag to make sure we call these functions from a place we know we have gotten a read table lock.

adizaimi:
So the localrep path for add_record does have iq->is_sorese set, but does not prelock the tables. I think there is no need to assert at this point, and that the flag that I pass is the right way to avoid locking table multiple times.

adizaimi:
Thank you Lingzhi.

PR #1151, opened 2018-09-30
fix more warnings

adizaimi:
fix more warnings

### DIFF
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -1429,24 +1429,20 @@ char *sql_field_default_trans(struct field *f, int is_out)
     int default_type;
     void *this_default;
     unsigned int this_default_len;
-    struct field_conv_opts_tz outopts = {0};
-    unsigned long long uival = 0;
-    long long ival = 0;
-    double dval = 0;
     char *cval = NULL;
     unsigned char *bval = NULL;
     int rc = 0;
     int null;
     int outsz;
     char *dstr = NULL;
-    int i;
 
     default_type = is_out ? f->out_default_type : f->in_default_type;
     this_default = is_out ? f->out_default : f->in_default;
     this_default_len = is_out ? f->out_default_len : f->in_default_len;
 
     switch (default_type) {
-    case SERVER_UINT:
+    case SERVER_UINT: {
+        unsigned long long uival = 0;
         rc = SERVER_UINT_to_CLIENT_UINT(
             this_default, this_default_len, NULL, NULL, &uival,
             sizeof(unsigned long long), &null, &outsz, NULL, NULL);
@@ -1454,8 +1450,9 @@ char *sql_field_default_trans(struct field *f, int is_out)
         if (rc == 0)
             dstr = sqlite3_mprintf("%llu", uival);
         break;
-
-    case SERVER_BINT:
+    }
+    case SERVER_BINT: {
+        long long ival = 0;
         rc = SERVER_BINT_to_CLIENT_INT(this_default, this_default_len, NULL,
                                        NULL, &ival, sizeof(long long), &null,
                                        &outsz, NULL, NULL);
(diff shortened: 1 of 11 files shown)


db/sqlglue.c
```diff
@@ -2233,8 +2217,8 @@ static void genid_hash_add(BtCursor *cur, int rrn, unsigned long long genid)
 
     /* printf("genid_hash_add(cur %d rrn %d genid %08llx rc %d)\n",
      * cur->cursorid, rrn, genid, rc); */
-    rc = bdb_temp_hash_insert(cur->bt->genid_hash, &key, sizeof(struct key),
-                              &genid, sizeof(genid));
```
nirbhayc: Does bdb_temp_hash_insert() log about failures? If not, may be we should check the return code and log an appropriate message?

db/sqlglue.c
```diff
@@ -2672,15 +2643,14 @@ static int cursor_move_compressed(BtCursor *pCur, int *pRes, int how)
...
      * and if peer_dropped_connection is true, we'll get NO SQL ENGINE and
      * a wasted thread apparently.
-     rc = cursor_move_preprop(pCur, pRes, how, &done, 0);
-     if (done)
-     return rc;
+    int done = 0;
+    int rc = SQLITE_OK;
+    rc = cursor_move_preprop(pCur, pRes, how, &done, 0);
+    if (done)
+    return rc;
```
nirbhayc: Check indentation

db/sqlglue.c
```diff
@@ -2672,15 +2643,14 @@ static int cursor_move_compressed(BtCursor *pCur, int *pRes, int how)
...
-    int rc = SQLITE_OK;
-
     /* skip preprop. if we're called from sqlite3_open_serial
      * and if peer_dropped_connection is true, we'll get NO SQL ENGINE and
      * a wasted thread apparently.
-     rc = cursor_move_preprop(pCur, pRes, how, &done, 0);
-     if (done)
-     return rc;
+    int done = 0;
+    int rc = SQLITE_OK;
```
nirbhayc: No point in initializing rc, gets overwritten below.

db/sqlglue.c
```diff
@@ -5415,7 +5361,7 @@ int sqlite3BtreeMovetoUnpacked(BtCursor *pCur, /* The cursor to be moved */
     if (pCur->bt->is_temporary) {
         if (pIdxKey) {
             if (bdb_is_hashtable(pCur->tmptable->tbl)) {
-                Mem mem = {0};
+                Mem mem = {{0}};
```
nirbhayc: What was the warning? Did you check if this really works (or only suppresses the warning)?

db/sqlglue.c
```diff
@@ -6410,7 +6352,7 @@ int get_data(BtCursor *pCur, struct schema *sc, uint8_t *in, int fnum, Mem *m,
 {
     int null;
     i64 ival;
-    double dval;
+    double dval = 0;
```
nirbhayc: IMO, we should move this to the specific switch case.

db/types.c
```diff
@@ -9935,9 +9916,7 @@ static void decimal64_ondisk_to_double(server_decimal64_t *pdec64,
                                        decDouble *dn)
 {
     char decimals[DECDOUBLE_PACKED_COEF];
-    int exponent;
-    int sign;
-    int rc = 0;
+    int exponent = 0;
```
nirbhayc: Looks like exponent always gets a value, why would compiler complain about this one?

db/sqlglue.c
```diff
@@ -1440,9 +1430,9 @@ char *sql_field_default_trans(struct field *f, int is_out)
     void *this_default;
     unsigned int this_default_len;
     struct field_conv_opts_tz outopts = {0};
-    unsigned long long uival;
-    long long ival;
-    double dval;
+    unsigned long long uival = 0;
+    long long ival = 0;
+    double dval = 0;
```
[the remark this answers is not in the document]
adizaimi: ok. doing so.

db/types.c
adizaimi: I dont think in this case compiler is that smart:
/bb/comdb2a/adi/master_branch/db/types.c:9578:15: warning: ‘exponent’ may be used uninitialized in this function [-Wmaybe-uninitialized]
         logmsg(LOGMSG_USER, "Dec64 unmake_order IN (exponent %d %x):\n", *exponent,

adizaimi:
Thank you Nirbhay.

PR #1157, opened 2018-10-03
put more trace for db create test

adizaimi:
dbcreate.test sometimes fails in check 5, as if the db dir exists (it's not supposed to). Putting more trace to investigate failure. Also cleanup some warnings.

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -2413,22 +2413,25 @@ struct dbenv *newdbenv(char *dbname, char *lrlname)
     if (!dbenv->basedir) {
         logmsg(LOGMSG_FATAL, "DB directory is not set in lrl\n");
         return NULL;
-    } else if (gbl_create_mode) {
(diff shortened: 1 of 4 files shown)


db/comdb2.c
```diff
@@ -2413,22 +2413,25 @@ struct dbenv *newdbenv(char *dbname, char *lrlname)
...
-        if (!S_ISDIR(sb.st_mode)) {
-            logmsg(LOGMSG_FATAL, "DB directory '%s' does not exist\n",
-                   dbenv->basedir);
-            return NULL;
         }
+    } 
+    
+    /* make sure the database directory exists! */
+    struct stat sb;
+    stat(dbenv->basedir, &sb);
```
ldennis: I think I saw this failed before in my branch. I think maybe we need to check rc here for stat. On errors, sb doesnt seem to be initialized.

Issue #1059, opened 2018-08-15
Replicant restart should reset master's slow replication moving average


akshatsikarwar:

akshatsikarwar:
@mponomar

mponomar:
Fixed in #1159

PR #1159, opened 2018-10-03
Reset slow replicant on disconnect

mponomar:
When we mark a node coherent, or a new code joins the cluster, reset information about it's historical ack times.  In the case of a node becoming coherent, it's already proved that it it's caught up.  In the case of a newly connected node, it won't ack old LSN.


bdb/rep.c
```diff
@@ -2665,72 +2696,79 @@ static void bdb_slow_replicant_check(bdb_state_type *bdb_state,
...
+        if (proctime[nodeix(host)] > worst_time) {
+            worst_time = proctime[nodeix(host)];
             worst_node = host;
+        }
     }
+    double second_worst_time = 0;
     for (int i = 0; i < numnodes; i++) {
         host = hosts[i];
         state = bdb_state->coherent_state[nodeix(host)];
         if (state != STATE_COHERENT && state != STATE_INCOHERENT_WAIT)
```
markhannum: Exclude incoherent_wait here

akshatsikarwar:
Looks good otherwise - I cannot reproduce issue #1059  with this build anymore.

PR #1162, opened 2018-10-04
 Don't use default in typestr if we have cdb2cfg override

mohitkhullar:
Don't use default in typestr if we have cdb2cfg override


cdb2api/cdb2api.c
```diff
@@ -927,8 +930,10 @@ void cdb2_set_comdb2db_config(const char *cfg_file)
                 cfg_file);
     memset(CDB2DBCONFIG_NOBBENV, 0, sizeof(CDB2DBCONFIG_NOBBENV) /* 512 */);
     if (cfg_file != NULL) {
+        cdb2cfg_override = 1;
         strncpy(CDB2DBCONFIG_NOBBENV, cfg_file, 511);
     } else {
+        cdb2cfg_override = CDB2CFG_OVERRIDE_DEFAULT;
```
mistachkin: It seems like this line should be moved into the reset_the_configuration() function?

cdb2api/cdb2api.c
```diff
@@ -947,10 +952,12 @@ void cdb2_set_comdb2db_info(const char *cfg_info)
         CDB2DBCONFIG_BUF = NULL;
     }
     if (cfg_info == NULL) {
+        cdb2cfg_override = CDB2CFG_OVERRIDE_DEFAULT;
```
mistachkin: Same as above, line should probably move into the reset_the_configuration() function.

mohitkhullar:
Thanks Joe

riverszhang89:
Discussed with Mohit. No need to port it to cdb2jdbc.

PR #1165, opened 2018-10-05
More metrics

mponomar:
* rework min/max in averager
* standing queue time
* dirty page evictions
* replication deadlocks
* net queue size
* net drops
* temptable spills
* total locks


markhannum:
okay to merge

PR #1169, opened 2018-10-06
Phys rep checkin rebase truncate

markhannum:
Submit for robo-rivers (better description in a bit)


schemachange/sc_stripes.c
```diff
@@ -43,29 +43,57 @@ int do_alter_stripes_int(struct schema_change_type *s)
...
 
     /* RENAME BLOB FILES */
     if (newblobstripe && !gbl_blobstripe) {
-        tran_type *tran = NULL;
-        struct ireq iq = {0};
-
-        init_fake_ireq(thedb, &iq);
-        iq.usedb = thedb->dbs[0];
 
         rc = trans_start(&iq, NULL, &tran);
```
markhannum: Can we use the same physical transaction that we just got?

schemachange/sc_callbacks.c
```diff
@@ -80,6 +81,53 @@ static int reload_rename_table(bdb_state_type *bdb_state, const char *name,
...
+{
+    void *tran = NULL;
+    int rc;
+    int bdberr = 0;
+    int stripes, blobstripe;
+    uint32_t lid = 0;
+    extern uint32_t gbl_rep_lockid;
+
+    stop_threads(thedb);
+    if (close_all_dbs() != 0)
```
markhannum: Do we have to stop_threads here?

tests/morestripe.test/runit
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env bash
```
markhannum: Add this test: writes lots of blobs when blobstripe is disabled, blobstripe it, write more, make sure all blobs are correct

tests/truncatesc.test/runit
```diff
@@ -0,0 +1,1695 @@
+#!/usr/bin/env bash
```
markhannum: Test for read access to non-existent tables that re-appear after the truncation

markhannum: Because we have static dbs[0], test the NO TABLES CASE

PR #1172, opened 2018-10-09
Parallel union all queries

dorinhogea:
This feature allows union all statements to run in parallel.  This supports syntax like:
select [col]* from table1 [where] [union all select [col]* from table2 [where]]* [order by <expr>] [limit <expr> [offset <expr>]]
The statements that cannot run in parallel are executed sequentially as before.
This change is meant as a standalone deliverable that introduces support for future sharding and more complex query rewrites.


CMakeLists.txt
```diff
@@ -76,7 +76,7 @@ add_definitions(
   -D_FILE_OFFSET_BITS=64
 )
 
-option(COMDB2_PER_THREAD_MALLOC "Turn OFF to run under Valgrind" ON)
+option(COMDB2_PER_THREAD_MALLOC "Turn OFF to run under Valgrind" OFF)
```
riverszhang89: Make it on?

riverszhang89:
/skipbuild because the pull request hangs the testsuite.

db/dohast.c
```diff
@@ -0,0 +1,568 @@
...
+}
+
+static char *describeExprList(Vdbe *v, const ExprList *lst)
+{
+    char *ret;
+    char *tmp;
+    char *newterm;
+    int i;
+
+    ret = sqlite3ExprDescribe(v, lst->a[0].pExpr);
```
ldennis: Handle `sortOrder` or return `NULL` to disable parallel exec for `order by ... desc` if this is a `TODO`

bdb/attr.h
```diff
@@ -645,6 +645,9 @@ DEF_ATTR(DEBUG_TIMEPART_SQLITE, dbg_timepart_SQLITE, BOOLEAN, 0, NULL)
 DEF_ATTR(DELAY_FILE_OPEN, delay_file_open, MSECS, 0, NULL)
 DEF_ATTR(DELAY_WRITES_IN_RECORD_C, delay_writes_in_record_c, MSECS, 0, NULL)
 
+DEF_ATTR(DOHSQL_VERBOSE, dohsql_verbose, BOOLEAN, 0, NULL)
+DEF_ATTR(DOHAST_VERBOSE, dohast_verbose, BOOLEAN, 0, NULL)
+DEF_ATTR(DOHSQL_DISABLE, dohsql_disable, BOOLEAN, 0, NULL)
```
mponomar: Can you please add a description for these?

db/dohsql.h
```diff
@@ -0,0 +1,79 @@
...
+struct dohsql_node {
+    enum ast_type type;
+    char *sql;
+    int ncols;
+    struct dohsql_node **nodes;
+    int nnodes;
+    int order_size;
+    int *order_dir;
+};
+typedef struct dohsql_node dohsql_node_t;
```
mponomar: What does doh stand for?

db/dohast.c
```diff
@@ -0,0 +1,657 @@
...
+        return 0;
+
+    if (ast->stack[0].op != AST_TYPE_SELECT)
+        return 0;
+    if (!ast->stack[0].obj)
+        return 0;
+
+    node = (dohsql_node_t *)ast->stack[0].obj;
+
+    if (strstr(node->sql, "sqlite_"))
```
mponomar: strcasestr (or really the parser should mark it as ineligible).

db/dohast.c
```diff
@@ -0,0 +1,657 @@
...
+        return "()";
+    default:
+        return "INVALID";
+    }
+}
+
+void ast_print(ast_t *ast)
+{
+    int i;
+    logmsg(LOGMSG_ERROR, "AST: [%d]\n", ast->nused);
```
mponomar: LOGMSG_USER?

db/dohast.c
dorinhogea: This will be reviewed on a further iteration; for now, to avoid internally generated queries that nest with the user queries, I am simply skipping the parallel optimization for these statements .

dorinhogea: switched to case insensitive

db/dohsql.h
dorinhogea: d stands for distributed.

bdb/attr.h
dorinhogea: hm, there is no way to provide description for bdb/attr.h parameters.  I have switched them to tunables.

sqlite/vdbeaux.c
```diff
@@ -5292,6 +5292,47 @@ Mem* sqlite3GetCachedResultRow(sqlite3_stmt *pStmt, int *nColumns)
 
   return NULL;
 }
+
```
mistachkin: When using the internal memory allocation routines (e.g. sqlite3Malloc, et al), the database mutex should be held; otherwise, an assert() will be fired (because the usage is not thread safe.

dorinhogea: why do we need a mutex for single-threaded execution?

mistachkin: When the per-thread malloc is disabled, the SQLite memory allocator is used and its mutex's must be respected.

dorinhogea: this is done during sqlite normal processing, so the mutex is acquired if enabled.

PR #1174, opened 2018-10-10
Turnaround test

mponomar:
Test a cluster turn.   Bring up cluster, turn all nodes twice.  Make sure that all nodes are queryable after turn.


riverszhang89:
Working on integrating the test into aws-robo.

PR #1175, opened 2018-10-11
fix small leak and a small refactor

adizaimi:
fix small leak and a small refactor


adizaimi:
Thank you Lingzhi. (Will do format now.)

PR #1182, opened 2018-10-12
osql_comm_is_done: avoid getting type from hdr unnecessarily

adizaimi:
We needlessly extract type and more from buffer when osql_comm_is_done() is called. This patch reduces the amount of work by passing type into the function and only getting snapinfo DONE_SNAP message.

### DIFF
--- a/db/osqlblockproc.c
+++ b/db/osqlblockproc.c
@@ -636,9 +636,6 @@ int osql_bplog_saveop(osql_sess_t *sess, char *rpl, int rplen,
     int bdberr;
     int debug = 0;
 
-    int ltype = 0;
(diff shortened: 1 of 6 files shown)


db/osqlblockproc.c
```diff
@@ -637,17 +636,18 @@ int osql_bplog_saveop(osql_sess_t *sess, char *rpl, int rplen,
     int bdberr;
     int debug = 0;
 
-    int type = 0;
-    buf_get(&type, sizeof(type), rpl, rpl + rplen);
+    int ltype = 0;
+    buf_get(&ltype, sizeof(ltype), rpl, rpl + rplen);
+    assert(ltype == type);
```
adizaimi: this is for debugging. will remove.

PR #1183, opened 2018-10-12
PORTED: fix leak of improper use of c_get DBT flags

ldennis:
ported from R6: fix leak of improper use of c_get DBT flags

### DIFF
--- a/bdb/ll.c
+++ b/bdb/ll.c
@@ -258,10 +258,15 @@ int ll_dta_del(bdb_state_type *bdb_state, tran_type *tran, int rrn,
     unsigned long long search_genid;
     int crc;
     int is_blob = 0;
+    int is_rowlocks = 0;
 
     if (dta_out) {
         bzero(dta_out, sizeof(DBT));
         dta_out->flags = DB_DBT_MALLOC;
+    }
+
+    if (tran->logical_tran) {
+        is_rowlocks = 1;
     }
 
     /* Always delete a masked genid for blobs. */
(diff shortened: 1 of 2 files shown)


bdb/ll.c
```diff
@@ -258,12 +258,17 @@ int ll_dta_del(bdb_state_type *bdb_state, tran_type *tran, int rrn,
...
     int crc;
     int is_blob = 0;
+    int is_rowlocks = 0;
 
     if (dta_out) {
         bzero(dta_out, sizeof(DBT));
         dta_out->flags = DB_DBT_MALLOC;
     }
 
+    if (tran->logical_tran) {
```
riverszhang89: Do you think we need to check `dtafile == 0` here?

ldennis: Sure, that could be an optimization to not verify updateid for blobs since we already have a rowlock on this row.

bdb/ll.c
```diff
@@ -258,12 +258,17 @@ int ll_dta_del(bdb_state_type *bdb_state, tran_type *tran, int rrn,
...
+    int is_rowlocks = 0;
 
     if (dta_out) {
         bzero(dta_out, sizeof(DBT));
         dta_out->flags = DB_DBT_MALLOC;
     }
 
+    if (tran->logical_tran && dtafile == 0) {
+        is_rowlocks = 1;
+    }
```
riverszhang89: Let's combine the declaration above and the if block here into

```
int is_rowlocks = (tran->logical_tran && dtafile == 0);
```

bdb/ll.c
```diff
@@ -909,17 +918,17 @@ static int ll_dta_upd_int(bdb_state_type *bdb_state, int rrn,
     int oldsz = -1;
     int newstripe = 0;
 
+    /* Verify updateid for ondisk data in rowlows mode.
```
riverszhang89: typo (rowlows)

PR #1190, opened 2018-10-16
Fix bb tests: get type from p_osql_rpl.type

adizaimi:
Fix bb tests: get type from p_osql_rpl.type


adizaimi:
comdb2test -S fix_type_from_osql_rpl.R20181016.8 | grep -v success
reports only three remsql tests not passing.

PR #1195, opened 2018-10-19
dont cleanup aggressively on successful test if run locally

adizaimi:
When running locally, it is best to keep the logs to make investigating issues or writing tests easier. In roborivers we now pass CLEANUPDBDIR=2 to make so it will still cleanup aggressively.


adizaimi:
When running tests locally, it is best to keep the logs to make investigating issues or writing tests easier. In roborivers we now pass CLEANUPDBDIR=2 to make so it will still cleanup aggressively.

PR #1196, opened 2018-10-19
osql: remove pending and complete lists because single session

adizaimi:
We no longer have block sql support (which had a session sent from replicant for every sql executed). Now all sql is sent under one session from replicant to master and most of the underlying infrastructure can be simplified to only work with this single session. 
This checkin cleansup by removing:
* unneeded lists pending and complete and replaces with iscomplete boolean. 
* conditinal var and mutex from tran object because no longer needed
* waiting and signaling between sessions (we only have one session)


adizaimi:
cdb2test   Nov 24 02:30:30 2018  in-comdb2db AZ_osql_remove_lists.R20181124.2

PR #1202, opened 2018-10-23
ported R6 PR912, no stats cleanup for view

dorinhogea:
- if views cannot be recreated, don't run.
- don't generate delete row operations from sqlite_stats when dropping a view, there are none there, and we end up generating a redundant transaction (sqlite patch 2f5be3a2)


db/sqlinterfaces.c
```diff
@@ -3512,6 +3512,8 @@ static int prepare_engine(struct sqlthdstate *thd, struct sqlclntstate *clnt,
                     logmsg(LOGMSG_ERROR,
```
mponomar:
```suggestion
                    logmsg(LOGMSG_FATAL, 
```

dorinhogea:
thank you

PR #1204, opened 2018-10-24
R6port few more views fixes

dorinhogea:
Ported adjusted fixes from R6 for proper order of the big locks: bdb, schema, views.


dorinhogea:
thanks

PR #1210, opened 2018-10-24
Add blob operations in restart_socksql test

riverszhang89:
Updating a blob consists of a couple of OSQL opcodes - OSQL_USEDB, OSQL_UPDCOLS, OSQL_QBLOB, OSQL_UPDREC and finally OSQL_DONE. If any of them fails, we should retry the operation from the beginning. The original OSQL retry logic did not handle OSQL_QBLOB failure correctly. Therefore, a restarted OSQL session would miss OSQL_QBLOB, resulting in the "no blobs flags with blob buffers" error.

The bug was fixed in 46f9eb3856d3b925825e41dd5988233441fbfda1. However we don't have a testcase for it. The commit adds blob updates to restart_socksql test.


db/osqlsqlthr.c
```diff
@@ -144,15 +144,15 @@ static inline int osql_should_restart(struct sqlclntstate *clnt, int rc)
...
         if (rc) {                                                              \
             logmsg(LOGMSG_ERROR,                                               \
                    "%s: failed to restart socksql session rc=%d\n", __func__,  \
                    rc);                                                        \
+        } else {                                                               \
+            restarted = 1;                                                     \
         }                                                                      \
-    }                                                                          \
-    if (rc) {                                                                  \
+    } else if (rc) {                                                           \
```
ldennis: if `osql_sock_restart` fails, do we want to coalesce rcode to `SQLITE_INTERNAL`?

ldennis:
Looks good otherwise.

db/osqlsqlthr.c
riverszhang89: Good point. Will undo this change.

riverszhang89:
@ldennis Appreciate the review!

PR #1212, opened 2018-10-25
Wrap pthread functions to abort on failure

adizaimi:
wrap pthread functions to abort on failure


adizaimi:
sc_inserts_deletes was a verify error which i am fixing in https://github.com/bloomberg/comdb2/pull/1216

PR #1214, opened 2018-10-25
Release locks

markhannum:
Release locks from heartbeat thread


akshatsikarwar:
This looks like a much larger patch than I was expecting. For e.g. do we still need to pass `thd` through `sbuf`?

markhannum:
It's used for access to the clnt structure from sql_writer .. heartbeat thread needs access to this to determine if the sql thread has set clnt->emitting_flag.

db/comdb2.h
```diff
@@ -3325,8 +3336,9 @@ int read_spfile(char *file);
 struct bdb_cursor_ifn;
 int recover_deadlock(bdb_state_type *, struct sql_thread *,
                      struct bdb_cursor_ifn *, int sleepms);
```
markhannum: From review: make this recover_deadlock_flags_int, check the rcode in recover_deadlock_flags, and do all cleanup (put_curtran, close_cursors, assert lockref == 0) if rcode is non-zero.

PR #1218, opened 2018-10-26
fix sp snapshot test

ldennis:
- fixed trigger/consumer crash under read committed and higher (only allow trigger/consumer in default transaction mode)
- used @adizaimi generated test framework for snapshot sp test.


adizaimi:
looks good. you can steal some verification from

PR #1220, opened 2018-10-29
initialize mutex

adizaimi:
initialize mutex

### DIFF
--- a/util/locks_wrap.h
+++ b/util/locks_wrap.h
@@ -29,7 +29,7 @@
         int rc;                                                                \
         LKDBG_TRACE(TRY, FUNC, OBJ);                                           \
         if ((rc = FUNC(__VA_ARGS__)) != 0) {                                   \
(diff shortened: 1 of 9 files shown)


akshatsikarwar:
Should we also catch `pthread_create` calls?

util/locks_wrap.h
```diff
@@ -43,7 +31,7 @@
         if ((rc = FUNC(__VA_ARGS__)) != 0) {                                   \
             logmsg(LOGMSG_USER, "%s:%d " #FUNC "(%p) rc:%d thd:%p\n",          \
```
akshatsikarwar: LOGMSG_FATAL

adizaimi:
you mean to toggle it off? not sure why you'd ever not want to abort(). It is a missbehavior and you should not continue -- imagine working with an unititialized lock.
Anyway, we can add that later if we agree that's what we want.

PR #1232, opened 2018-11-02
Client API: read config only once.

riverszhang89:
Currently cdb2api and cdb2jdbc read the config files twice when database discovery is needed.


riverszhang89:
/slient because the pull request fixes both cdb2api and cdb2jdbc.

PR #1233, opened 2018-11-04
schema_lk: call unlock was missing from ROWLOCKS codepath

adizaimi:
* call unlock_schema_lk() from missing place causing hang on OSQL_DONE for vutf8_rowlocks test
* call accessors instead of actual manipulating schema_lk
* move schema_lk to separate file and allow to modify only via accessors


db/osqlcomm.c
```diff
@@ -6743,6 +6743,10 @@ int osql_process_packet(struct ireq *iq, unsigned long long rqid, uuid_t uuid,
                 iq->usedb = iq->sc->db;
             rc = finalize_schema_change(iq, iq->sc_tran);
             iq->usedb = NULL;
+            if (iq->sc_locked) {
+                unlock_schema_lk();
+                iq->sc_locked = 1;
```
ldennis: This is wrong. The schema change is being committed and the schemalk will be released after it commits/aborts. See the comment above. There might be bugs there, but I dont think releasing lock here is correct.

ldennis:
The bug is [here](https://github.com/bloomberg/comdb2/blob/master/db/toblock.c#L5599). DDL commit/abort handling for rowlock is missing. I will work on a fix.

db/osqlcomm.c
adizaimi: Yep, this was definitely wrong. Thanks

PR #1234, opened 2018-11-06
fix sc under rowlocks

ldennis:
- make rowlocks same as pagelock mode when handling schema change
- create a single parent physical trans for a logical tran
- create two children from the parent, one for dml, one for ddl
- commit all in good case
- in bad case, abort ddl first, reload table (parent holds table locks for us), and abort the rest.


adizaimi:
This looks good. 
I wanted to add in https://github.com/bloomberg/comdb2/pull/1219 ability to have transactions of basic.test [rowlocks.testopts] in snapshot isolation via setting 'sql_tranlevel_default snapshot'. I am a bit confused as to what init_with_rowlocks makes the db do, but does it set default tran to snapshot?

ldennis:
In rowlocks mode, it has to be snapshot (i.e. default). Otherwise, you would have dirty read.

PR #1237, opened 2018-11-06
Track service time per client.

riverszhang89:
Add `svc_time` column in comdb2_clientstats table to keep track of client service time.


riverszhang89:
The latest commit fixes missing client info from an SSL connection in cdb2api. /silent because cdb2jdbc handles this correctly.

riverszhang89:
@ldennis Appreciate the review!

PR #1239, opened 2018-11-07
SSL hostname match: honor RFC 6125 6.4.3

riverszhang89:
The original logic misinterprets it.


riverszhang89:
The pull request also passes all internal testcases.

Issue #1187, opened 2018-10-15
uniqnulls doesnt work for non-inplace updates


ldennis:
```
ldengdb> create table t options ipu off {schema{int i null=yes} keys{uniqnulls "IDX" = i}} $$
[create table t options ipu off {schema{int i null=yes} keys{uniqnulls "IDX" = i}}] rc 0
ldengdb> insert into t values (NULL)
(rows inserted=1)
[insert into t values (NULL)] rc 0
ldengdb> update t set i = NULL
[update t set i = NULL] failed with rc 304  unable to update record rc = 177
```

I think the bug is [here](https://github.com/bloomberg/comdb2/blob/master/db/record.c#L1553):

For uniqnulls and null keys, if genid changed due to an update, we should fall into the del+add case instead of calling ix_upd_key()

mistachkin:
Fix is now merged to master.

PR #1252, opened 2018-11-13
Disable in-place update when the 'uniqnulls' option is in use and applies to the record being updated.

mistachkin:
Add checking in 'upd_record' to prevent an update from being done in-place when one or more NULL values are used and the 'uniqnulls' option is enabled.  This should fix the issue described in #1187.


markhannum:
Okay to merge

PR #1253, opened 2018-11-13
cdb2api: Keep a string duplicate of the last query.

riverszhang89:
Currently we latch the query in case of an HASQL retry by remembering the address of the query pointer.
This is not safe if the content in the address changes. For instance, the following code will result in
an invalid read of freed memory on the HASQL retry.

```c
/* pseudo code */
char *sql = malloc(...);
sprintf(sql, "SELECT * FROM TABLE");
cdb2_run_statement(sql);
free(sql);
while (cdb2_next_record() == CDB2_OK); /* server crashes here. */
```

### DIFF
--- a/cdb2api/cdb2api.c
+++ b/cdb2api/cdb2api.c
@@ -3716,8 +3716,10 @@ retry_queries:
             is_begin ? 0 : run_last, __LINE__);
     } else {
         /* Latch the SQL only if we're in a SNAPSHOT HASQL txn. */
-        if (hndl->snapshot_file != 0)
+        if (hndl->snapshot_file != 0) {
+            free(hndl->sql);
             hndl->sql = strdup(sql);
+        }
 
         hndl->query_no += run_last;
         rc = cdb2_send_query(hndl, hndl->sb, hndl->dbname, (char *)sql, 0, 0,
--- a/cdb2api/cdb2api.c
+++ b/cdb2api/cdb2api.c
@@ -2884,14 +2884,10 @@ int cdb2_close(cdb2_hndl_tp *hndl)
         hndl->commands = NULL;
     }
 
-    if (hndl->query)
-        free(hndl->query);
-
-    if (hndl->query_hint)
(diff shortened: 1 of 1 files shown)


riverszhang89:
/silent because strings are managed by GC in Java and they are immutable.

cdb2api/cdb2api.c
```diff
@@ -3708,7 +3704,6 @@ static int cdb2_run_statement_typed_int(cdb2_hndl_tp *hndl, const char *sql,
 
     clear_responses(hndl);
 
-    hndl->sql = (char *)sql;
```
mistachkin: So, prior to this change, we always kept track of the SQL (even when not in HA mode)?  It seems like that would be a  useful thing to preserve?  In that case, we could just always strdup() it?

riverszhang89: We always kept track of the last SQL even when not in HA mode, although only HA needs it to replay the query in case of a `cdb2_next_record()` failure. I agree that this would be useful to keep around, however calling `strdup()` on every single query sounds a bit too much overhead to me. By only strdup()'ing the SQL in HA mode, we could cut down a few CPU cycles for a non-HA query. A micro-optimization :)

mistachkin: I don't think one call to strdup() per query (?) is going to be that costly.  It seems to me that *not* calling strdup() on an arbitrary string passed by the caller and then possibly trying to free() it could be problematic.  e.g. What if the caller passes in a string constant?  Conditionally calling strdup() could be OK *if* we keep track of whether the "sql" string needs to be freed using an additional flag of some kind.

riverszhang89: With this change, `hndl->sql` is either `NULL` (implied by `calloc()`-ing the handle), or a `strdup()`'d string. They both can be freed safely (although `free()`-ing NULL is a no op).

riverszhang89:
The pull request has passed the internal testsuite as well.

PR #1256, opened 2018-11-15
check for null db and handle before calling cleanups

adizaimi:


adizaimi:
Lingzhi fixed in 1258.

PR #1257, opened 2018-11-15
Fix master info in dbinfo response

nirbhayc:
.

### DIFF
--- a/bdb/info.c
+++ b/bdb/info.c
@@ -689,7 +689,7 @@ void fill_dbinfo(void *p_response, bdb_state_type *bdb_state)
             nodeinfos[i]->room =
                 bdb_state->callback->getroom_rtn(bdb_state, nodes[j].host);
             nodeinfos[i]->name = strdup(nodes[j].host);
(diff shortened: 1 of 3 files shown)


bdb/info.c
```diff
@@ -689,7 +689,7 @@ void fill_dbinfo(void *p_response, bdb_state_type *bdb_state)
             nodeinfos[i]->room =
                 bdb_state->callback->getroom_rtn(bdb_state, nodes[j].host);
             nodeinfos[i]->name = strdup(nodes[j].host);
-            if (strcmp(bdb_state->repinfo->master_host, nodes[i].host) == 0) {
+            if (strcmp(bdb_state->repinfo->master_host, nodes[j].host) == 0) {
                 master->number = 0; /* will not be used by client */
```
mohitkhullar: can we have master->number = j
This will make it work with bbgithub cdb2api too without any change on its end.

mohitkhullar: nodeinfos[i]->number = j 
too @ line 685

mohitkhullar:
Looks good for R7, just 2 changes to make it work with older version of cdb2api

PR #1259, opened 2018-11-16
refactor osql sock restart function

adizaimi:
refactor osql sock restart function


db/osqlsqlthr.c
```diff
@@ -700,123 +699,113 @@ int osql_sock_restart(struct sqlclntstate *clnt, int maxretries,
...
     if (!thd) {
         logmsg(LOGMSG_ERROR, "%s:%d Bug, not sql thread !\n", __func__, __LINE__);
         cheap_stack_trace();
     }
 
-again:
+    do {
+        /* if we're shaking really badly, back off */
+        if (retries > 1)
+            usleep(retries * 10000); //sleep for a multiple of 10ms
```
ldennis: We dont sleep for the first retry? `if (retries)` instead?

adizaimi: original code was doing the same...was trying to keep the logic. A bit silly actually, will probably do what you say.

PR #1262, opened 2018-11-20
better table name check and skip

dorinhogea:
avoid blindly searching for strings, instead check the table name and skip internal tables for parallel scheme (sqlite_ & comdb2_).

### DIFF
--- a/db/dohast.c
+++ b/db/dohast.c
@@ -134,7 +134,7 @@ char *sqlite_struct_to_string(Vdbe *v, Select *p, Expr *extraRows,
         return NULL; /* no group by */
     if (p->pSrc->nSrc > 1)
         return NULL; /* no joins */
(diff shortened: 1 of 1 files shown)


db/dohast.c
```diff
@@ -427,6 +428,24 @@ static dohsql_node_t *gen_union(Vdbe *v, Select *p, int span)
     return node;
 }
 
+static int skip_tables(Select *p)
```
nirbhayc: [Optional] s/skip_tables/skip_system_tables ?

dorinhogea:
thank you for review

db/dohast.c
dorinhogea: This is a callback to skip arbitrary tables, not only internal ones.

PR #1264, opened 2018-11-20
Fix race between memp_alloc_flags and trickle_do_work

markhannum:
Fix race between memp_alloc_flags and trickle_do_work which can double-flip a bufferpool buffer.  Add tunable which I will use to attempt to reproduce this.


markhannum:
No .. im wrong .. alloc makes sure ref is 0 ..

PR #1272, opened 2018-11-26
bdb: More accurate online page compaction request size limit.

riverszhang89:
The maximum size should be maximum key size which is sizeof genid + max record size.


riverszhang89:
The pull request passes all internal tests as well.

PR #1278, opened 2018-11-27
Add new keywords to fallback list

nirbhayc:
.


nirbhayc:
> These changes should not be necessary once the new SQLite version is merged?

Right

PR #1280, opened 2018-11-28
sqlite/parse.y: Add 'KEY' back to fallback ID

nirbhayc:
This is needed to preserve backwards compatibility with R6. This change, however, introduced a conflict in the parser which broke the DDL syntax introduced in R7 - where one could use 'KEY' to create a dup key - causing the following command to fail with syntax error.

`CREATE TABLE t(i INT KEY);`

The syntax error was fixed by disallowing parser to accept multi-word types, as it caused parser to think of 'KEY' (in the above command) as part of data type. Internal use of multi-word 'interval' types has also been updated.


nirbhayc:
Fix based on @akshatsikarwar's idea: https://github.com/bloomberg/comdb2/pull/1283

PR #1286, opened 2018-11-29
Fix stored procedure service time.

riverszhang89:
Service timers aren't reset for a stored procedure.


riverszhang89:
The pull request has passed robomark:

```
cdb2test   Nov 29 22:37:03 2018  in-comdb2db sp_svc_time.R20181129.8
```

PR #1287, opened 2018-11-30
Include git sha in executable for identification purposes

adizaimi:
Include git sha in executable for identification purposes:
```
> strings db/comdb2 | grep GIT_VERSION_SHA
GIT_VERSION_SHA=49aeacd6
```


adizaimi:
If there are local changes, it will show: 
`GIT_VERSION_SHA=d4b5da5c with local changes`

PR #1293, opened 2018-12-03
Fix cinsert_linearizable

markhannum:
Simple cdb2api fix: don't send snapinfo from a non-existant stackframe.  I've added a server-side sanity check which optionally cores on an insane request.


akshatsikarwar:
I like https://github.com/bloomberg/comdb2/pull/1294 better.

PR #1296, opened 2018-12-04
cleanup more warnings

adizaimi:
* cleanup more warnings
* exclude noncomdb2 directories from some warning types


mem/mem.c
```diff
@@ -1129,8 +1129,10 @@ char *comdb2_strndup_static(int indx, const char *s, size_t n)
 
 struct mallinfo comdb2_mallinfo_static(int indx)
 {
+#ifdef INSECURE
     struct mallinfo empty = {0};
     STATIC_RANGE_CHECK(indx, empty);
```
riverszhang89: `STATIC_RANGE_CHECK(indx, (struct mallinfo){0});` would read neater than this.

akshatsikarwar: Should that be Unsecure?

adizaimi: That seems to work, thanks Rivers.

akshatsikarwar:
If only gcc then test for gcc (instead of linux)

akshatsikarwar:
`if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")`

akshatsikarwar:
`if ("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")`

mem/mem.c
akshatsikarwar: INSECURE means something else, no?

adizaimi: STATIC_RANGE_CHECK is wrapped in an ifdef INSECURE   block. I was using same, but not needed if I use Rivers' trick. Thanks.

PR #1297, opened 2018-12-05
Update documentation to pass 'local' to cdb2sql

AidanDelaney:
The overview documentation invokes `cdb2sql testdb` without passing
the required parameter `dbtype`.  The 'local' `dbtype` is a valid
way to access a comdb2 instance running locally.

    - This is a *documentation* fix

    - Currently when running 7.0.R2 I can't work through the overview documentation.  This is because the `cdb2sql` binary requires a `dbtype` parameter to be passed to it.

    - After applying this PR the documentation is modified to suggest users execute `cdb2sql testdb 'local' 'select (1+1) as "math"'`.  By passing the `dbtype` to `cdb2sql` the query successfully runs.

### DIFF
--- a/docs/pages/overview/example_db.md
+++ b/docs/pages/overview/example_db.md
@@ -18,7 +18,7 @@ $ comdb2 --create testdb  &
 $ comdb2 testdb &
 ...
 # Run queries
-$ cdb2sql testdb 'select 1+1 as "math"'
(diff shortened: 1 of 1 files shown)


mponomar:
'local' is the default in the current master - it's assumed if no type is specified.  Documentation tracks that.  There's older versions where that wasn't the case.

PR #1305, opened 2018-12-07
Authorizer callback for SQLite for use by the stored procedures subsystem

mistachkin:
These changes forbid any DDL statements from being executed via Lua stored procedures.


mistachkin:
These changes are not quite complete.  The code in 'comdb2build.c' needs to be modified to call into the SQLite authorizer subsystem.

akshatsikarwar:
I prefer to see this done a bit more surgically.
Enable authorizer only during `lua_prepare_sql_int`. That is the only site where user provided queries are run. Undo all changes in `exec_thread_int/exec_procedure_int`. This will also allow us to deny `SQLITE_CREATE_TEMP_TABLE` but only when done through user specified query.

akshatsikarwar:
We want to use stmt caching if its enabled whether authorizer is active or not.

akshatsikarwar:
Here is an alternate implementation which does not change the world:
https://github.com/akshatsikarwar/comdb2/commit/6d34550a315ebe5bff5cc5e84c45418ff9705cdd

PR #1306, opened 2018-12-08
Fix issues with the integration of query parallelization (e.g. UNION)...

mistachkin:
This change fixes crashes in the 'sc_timepart' and 'union_parallel' tests.


mistachkin:
Both test suites are now passing.

PR #1310, opened 2018-12-12
send r6 compatible poke

ldennis:
.

### DIFF
--- a/db/osqlcomm.c
+++ b/db/osqlcomm.c
@@ -48,6 +48,7 @@
 #include "views.h"
 #include "str0.h"
 #include "sc_struct.h"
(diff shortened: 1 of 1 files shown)


db/osqlcomm.c
```diff
@@ -3444,8 +3445,8 @@ int osql_comm_send_poke(char *tohost, unsigned long long rqid, uuid_t uuid,
 
         poke.tstamp = comdb2_time_epoch();
 
-        poke.from = 0;
-        poke.to = 0;
+        poke.from = nodenum(gbl_mynode);
```
akshatsikarwar: Use `gbl_mynodeid`

PR #1316, opened 2018-12-14
Register database tunables in a more memory friendly way.

riverszhang89:
`gbl_tunables->array` is only used by the tunables system table to loop through all the tunables, which can be replaced by invoking hash_next() on `gbl_tunables->hash`.
This reduces the memory footprint of the tunables system by ~800KiB.


riverszhang89:
The pull request is passing robomark as well:

```
cdb2test   Dec 14 12:52:15 2018  in-comdb2db malloc2.R20181214.5
```

@nirbhayc and @mponomar Thank you for the review.

PR #1323, opened 2018-12-15
Another dumb bug breaking robomark

markhannum:
Don't try to grab a mutex before its initialized


markhannum:
The sc_tableversion failure failed in verify with 'table does not exist'.  This is a real error, but not related to this pr.

PR #1324, opened 2018-12-17
#ifdef osql session tracking code

riverszhang89:
Reduce the size of the data segment by ~66MiB.


riverszhang89:
It is passing robomark. Mark is looking at localrep failure on roborivers. I'm merging the pull request so I can get R7 rolling this afternoon.

PR #1325, opened 2018-12-17
set cnonce to NULL if not available to avoid crashes

adizaimi:
On master side when printing deadlocks, set cnonce to NULL if not available (ex. SC thread) to avoid crashes.


adizaimi:
Failures are for verify in both above cases.

adizaimi:
Thank you Joe.

PR #1328, opened 2018-12-18
explain distribution

dorinhogea:
cdb2sql db1 local 'explain distribution select a from t1 union all select a from t2'
(Plan='Threads 2')
(Plan='SeLeCT a FRoM "t1"')
(Plan='SeLeCT a FRoM "t2"')


nirbhayc:
also formatting

dorinhogea:
thank you

PR #1330, opened 2018-12-18
Fix checkpoint handling on backwards roll

markhannum:
Fixed a bug where the backwards roll phase of recovery blindly replaces the checkpoint with the previous checkpoint.  This breaks if recovery needs to start near the beginning of the first log we have, and the previous checkpoint lsn points to a log which we don't have.  Simple solution: only update if the log record exists.

### DIFF
--- a/berkdb/txn/txn_rec.c
+++ b/berkdb/txn/txn_rec.c
@@ -634,10 +634,28 @@ __txn_ckp_recover(dbenv, dbtp, lsnp, op, info)
 		return (ret);
 
 	if (op == DB_TXN_BACKWARD_ROLL) {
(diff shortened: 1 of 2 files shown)


berkdb/txn/txn_rec.c
```diff
@@ -634,10 +634,28 @@ __txn_ckp_recover(dbenv, dbtp, lsnp, op, info)
 		return (ret);
 
 	if (op == DB_TXN_BACKWARD_ROLL) {
+        DB_LOGC *logc;
```
riverszhang89: tabz.

riverszhang89:
👍 Thank you for fixing it Mark

PR #1334, opened 2018-12-20
DDL: Create/select right indices for FKs during ALTER

nirbhayc:
This commits fixes 2 issues:
* Logic to find the right key for an FK did not work properly
* Keys previously deleted in the ALTER command were not ignored


sqlite/src/comdb2build.c
```diff
@@ -3015,6 +3018,7 @@ static char *prepare_csc2(Parse *pParse, struct comdb2_ddl_context *ctx)
 
             if (key_found == 1) {
                 constraint->child = key;
+                break;
```
ldennis: Looks like this "find key" logic is used a few times in this code. It would be nice to have this logic in one function. Given a list of indexes and a column list to be matched, the function could return a pointer to the key object or NULL if not found.

nirbhayc:
Thanks @ldennis !

PR #1346, opened 2018-12-27
Fix reference count handling for temp table cursors.

mistachkin:
These changes address issues that can arise when a temp table is shared by multiple threads.  When a cursor is closed, the VDBE may call sqlite3BtreeClose, which will reduce the reference counts for *all* temp tables attached to that btree.  This may cause one or more temp tables in use by other threads to be prematurely freed.


db/sqlglue.c
```diff
@@ -7342,7 +7410,9 @@ sqlite3BtreeCursor_temptable(Btree *pBt,      /* The btree */
     }
 
     cur->cursor_class = CURSORCLASS_TEMPTABLE;
-    cur->tmptable->tbl = src->tbl;
+    assert( src->tbl );
+    assert( src->nRef>0 );
+    cur->tmptable->tbl = src->tbl; src->nRef++;
```
adizaimi: line break?

adizaimi:
Looks good to me. Just apply formatting as last step. 
Thanks Joe.

db/sqlglue.c
mistachkin: Fixed.

PR #1347, opened 2018-12-28
pthread_mutex_timedlock expects an absolute time.

riverszhang89:
The bug causes high CPU use for remote SQL queries.


riverszhang89:
```
cdb2test   Dec 28 11:31:21 2018  in-comdb2db abstime.R20181228.2
```

PR #1348, opened 2018-12-31
Pg free recovery

markhannum:
Disable online recovery by default

### DIFF
--- a/bdb/llmeta.c
+++ b/bdb/llmeta.c
@@ -2705,7 +2705,7 @@ int bdb_get_pagesize_allindex(tran_type *tran, int *pagesize, int *bdberr)
                             pagesize, bdberr);
 }
 
-int bdb_add_dummy_llmeta_flags(uint32_t flags)
+int bdb_add_dummy_llmeta_wait(int wait_for_seqnum)
 {
     tran_type *tran;
     int rc;
@@ -2749,7 +2749,7 @@ retry:
     rc = bdb_tran_commit_with_seqnum_size(llmeta_bdb_state, tran, &ss, NULL,
                                           &bdberr);
 
-    if (rc == 0 && flags) {
+    if (rc == 0 && wait_for_seqnum) {
         int timeoutms;
         rc = bdb_wait_for_seqnum_from_all_adaptive_newcoh(llmeta_bdb_state, &ss,
                                                           0, &timeoutms);
@@ -2784,7 +2784,7 @@ fail:
 
 int bdb_add_dummy_llmeta(void)
 {
-    return bdb_add_dummy_llmeta_flags(1);
+    return bdb_add_dummy_llmeta_wait(1);
 }
 
 /* store a new csc2 schema in the llmeta table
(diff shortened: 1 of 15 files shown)


markhannum:
This is ready for review

berkdb/rep/rep_record.c
```diff
@@ -6613,8 +6613,14 @@ __rep_dorecovery(dbenv, lsnp, trunclsnp, online)
...
 	/* Tell replicants to truncate */
-	if (dbenv->rep_truncate_callback)
-		dbenv->rep_truncate_callback(dbenv, trunclsnp, i_am_master);
+	if (dbenv->rep_truncate_callback) {
+        uint32_t rep_truncate_flags = 0;
+        if (i_am_master)
+            rep_truncate_flags |= DB_REP_TRUNCATE_MASTER;
+        if (online)
+            rep_truncate_flags |= DB_REP_TRUNCATE_ONLINE;
+		dbenv->rep_truncate_callback(dbenv, trunclsnp, rep_truncate_flags);
```
adizaimi: format is off?

bdb/rep.c
```diff
@@ -2192,10 +2192,11 @@ uint32_t bdb_get_rep_gen(bdb_state_type *bdb_state)
     return mygen;
 }
 
-void send_newmaster(bdb_state_type *bdb_state)
+void send_newmaster(bdb_state_type *bdb_state, int online)
 {
     bdb_state->dbenv->rep_start(bdb_state->dbenv, NULL, 0, DB_REP_MASTER);
-    bdb_add_dummy_llmeta();
+    /* Online recovery will wait-for-seqnum */
+    bdb_add_dummy_llmeta_flags(online);
```
adizaimi: The parameter is an int, not really flags, right?

bdb/llmeta.c
```diff
@@ -2749,7 +2749,7 @@ int bdb_add_dummy_llmeta(void)
     rc = bdb_tran_commit_with_seqnum_size(llmeta_bdb_state, tran, &ss, NULL,
```
adizaimi: what about if rc != 0 here?

markhannum:
The pg_free_recovery test intends to tease out a parallel replication bug: single threaded replication (original Berkeley) does not need to aquire pagelocks for compensating transactions.  Parallel rep requires them: if there's no pagelock, nothing prevents a subsequent transaction from using a page before it has been returned to the freelist.  pg_free_recovery tests this by introducing latency in the pg_free_recovery handler.

Online recovery allows for the possibility of additional latency while truncating the logfile.  This is because rather then evicting all threads from Berkeley, the recovery-thread contends with other transactions in acquiring pagelocks.  This additional latency, plus the artificial latency introduced by the pg_free_recovery tunable is sometimes sufficient to cause a replicant to be watchdog'd ("DATABASE MAKES NO PROGRESS").

The fix is to disable online recovery in the pg_free_recovery test.  Additionally, this pr cleans up the logic for explicit logfile truncations by disabling wait-for-seqnum if the master is holding the bdb writelock (the not-online case).  This only means that if online-recovery is not enabled, the master won't attempt to wait for replicants to finish truncation before returning control to the client.

PR #1349, opened 2018-12-31
berkdb: Grow fileid track array exponentially.

riverszhang89:
Realloc by 1 element at a time is not optimal. It increases the likelihood of malloc-copy-free and fragmentation.

(DRQS 135224293)

### DIFF
--- a/berkdb/dbreg/dbreg_rec.c
+++ b/berkdb/dbreg/dbreg_rec.c
@@ -100,15 +100,15 @@ __dbreg_register_recover(dbenv, dbtp, lsnp, op, info)
 		int i;
 		ft = &dbenv->fileid_track;
 		if (argp->fileid >= ft->numids) {
+			int dblsize = argp->fileid * 2 + 1;
 			if ((ret = __os_realloc(dbenv,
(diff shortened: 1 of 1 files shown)


berkdb/dbreg/dbreg_rec.c
```diff
@@ -100,13 +100,15 @@ __dbreg_register_recover(dbenv, dbtp, lsnp, op, info)
 		int i;
 		ft = &dbenv->fileid_track;
 		if (argp->fileid >= ft->numids) {
-			__os_realloc(dbenv, (argp->fileid+1) *
```
mistachkin: It would make this change easier to read if you added a local variable with the new allocation size right inside this "if" statement, e.g.: "int newNumids = argp->fileid * 2 + 1;" and then make use of that instead of re-calculating it later.

riverszhang89:
```
cdb2test   Dec 31 14:27:08 2018  in-comdb2db realloc_resize.R20181231.5
```

riverszhang89:
@mistachkin Thanks for the review!

PR #1355, opened 2019-01-02
handle deadlock for ix_find_by_key_tran

ldennis:
.


db/indices.c
```diff
@@ -98,6 +98,12 @@ static int check_index(struct ireq *iq, void *trans, int ixnum,
     } else if (rc == RC_INTERNAL_RETRY) {
         *retrc = RC_INTERNAL_RETRY;
         return 1;
+    } else if (rc != IX_FNDMORE || rc != IX_NOTFND || rc != IX_PASTEOF ||
+               rc != IX_EMPTY) {
```
nirbhayc: I think you meant && (and not ||) ?

PR #1356, opened 2019-01-03
Release locks

markhannum:
Release locks on slow reader if there are any waiters (i.e., the replication thread which is attempting to replicate a schema change).  Add a 'truncate_test' to the cldeadlock testcase to verify that this logic works correctly.


markhannum:
Simple change - okay to merge

PR #1357, opened 2019-01-03
newsql: Handle "SET REMOTE" failure correctly.

riverszhang89:
The SQL statement should fail if the "SET REMOTE" command is invalid.


adizaimi:
May I suggest we reuse an existing test such as sql.test (or maybe simple_remsql.test) for this instead of creating a new one?

riverszhang89:
@adizaimi Thanks for the suggestion.
@mohitkhullar  Thank you for the review.

PR #1364, opened 2019-01-08
Disjoint pgallocs

markhannum:
Don't grab the lock for the meta-page inline with a transaction: rather, spawn a different transaction (a detached child) which allocates a page, commits, and returns the page to the parent transaction.  This is a very low level change- I've temporarily added disjoint_pgallocs.testops file to all of our tests to verify that they work correctly with this feature enabled.


riverszhang89:
Test `killcluster` was hung .. had to manually kill the database..

PR #1365, opened 2019-01-08
Put the abort on missing osql session logic on a tunable

markhannum:
cldeadlock manages to trigger the osql-missing-session abort regularly.  The abort is incorrect as osql sessions can go missing legitimately on a master swing.


markhannum:
Test errors are unrelated (will address them in a different PR)

PR #1367, opened 2019-01-09
Add tunable to enable use of the PRAGMA SQL command.

mistachkin:
These changes add a tunable that may be used to enable use of the PRAGMA SQL command.


db/sqlinterfaces.c
```diff
@@ -3177,14 +3174,12 @@ static int handle_non_sqlite_requests(struct sqlthdstate *thd,
         handle_stored_proc(thd, clnt);
         *outrc = 0;
         return 1;
-#if defined(SQLITE_DEBUG) || defined(COMDB2_ALLOW_SQLITE_PRAGMA)
     } else if (is_pragma) {
         /* currently, all PRAGMA requests, when allowed, are handled
         ** by SQLite */
         logmsg(LOGMSG_WARN, "%s:%d %s ALLOWING PRAGMA [%s]\n", __FILE__,
```
mponomar: We probably don't want this trace anymore.

mistachkin: Maybe we can retain it and change the level to 'INFO' instead?  Seems like it could be useful.

PR #1379, opened 2019-01-14
made pglogs use one temp table per log file

ldennis:
optimization to use less temp tables


markhannum:
Reviewed - okay to merge

PR #1381, opened 2019-01-15
Eventlog: Add ability for db to delete old event log files

adizaimi:
Add ability for db to delete all except the last N even log files:
* Currently set default eventlog_nkeep to 10. 
* Set to 0 to disable deletion
* Fix setting eventlog_rollat  correcly. 
* Initial size for eventlog_rollat  set to a more reasonable 100MB.

### DIFF
--- a/db/eventlog.c
+++ b/db/eventlog.c
@@ -108,8 +108,8 @@ static void eventlog_roll_cleanup()
     // WARNING: MAKE SURE NO SPACE BETWEEN THE TWO CHARACTERS '%s*'
     // IN THE CALL TO ls IN NEXT LINE
     snprintf(cmd, sizeof(cmd) - 1, 
-             "ls -1tr %s* | head -n -%d | xargs -d '\n' rm -f --",
-             fname, eventlog_nkeep);
+             "ls -1t %s* | grep '%s' | grep '.events.' | sed '1,%dd' | xargs rm -f",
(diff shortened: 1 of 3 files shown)


db/eventlog.c
```diff
@@ -84,8 +84,28 @@ static inline void free_gbl_eventlog_fname()
...
 
+static void eventlog_roll_cleanup()
+{
+    if (eventlog_nkeep == 0)
+        return;
+
+    char cmd[512] = {0};
+    char *fname = comdb2_location("logs", "%s.events", thedb->envname);
+    snprintf(cmd, sizeof(cmd) - 1, 
+             "ls -1tr %s* | head -n -%d | xargs -d '\n' rm -f --",
```
mponomar: Maybe being unreasonable paranoid here, but let's check that fname is not empty?  Afraid of an unbounded delete.

adizaimi: Sure. I added check that the fname ends with '.events' for more accuracy.

adizaimi:
Add ability for db to delete all except the last N even log files:
* Currently set default eventlog_nkeep to 10. Can be set to 0 to disable deletion.
* Fix setting eventlog_rollat correcly. 
* Initial size for eventlog_rollat set to a more reasonable 100MB.
* Add to eventlog.test a case to verify that we roll the logs and keep the last N correctly

PR #1382, opened 2019-01-15
More constraints fixes

ldennis:
- always assign `sc_to` under mutex
- verify schema change constraints once per genid *after* processing all indexes from the `ct_add_table` temp table
- refactor is_genid_right_of_stripe_pointer to reduce duplicate code
- only verify sc constraints if genid is to the left of sc pointers

### DIFF
--- a/schemachange/sc_callbacks.c
+++ b/schemachange/sc_callbacks.c
@@ -167,9 +167,9 @@ int is_genid_right_of_stripe_pointer(bdb_state_type *bdb_state,
 {
     int stripe = get_dtafile_from_genid(genid);
     if (stripe < 0 || stripe >= gbl_dtastripe) {
-        logmsg(LOGMSG_ERROR, "%s: genid 0x%llx stripe %d out of range!\n",
+        logmsg(LOGMSG_FATAL, "%s: genid 0x%llx stripe %d out of range!\n",
                __func__, genid, stripe);
-        return -1;
+        abort();
     }
     if (!sc_genids[stripe]) {
         /* A genid of zero is invalid.  So, if the schema change cursor is at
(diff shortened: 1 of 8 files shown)


schemachange/sc_callbacks.c
```diff
@@ -163,9 +163,33 @@ static int reload_rowlocks(bdb_state_type *bdb_state, scdone_t type)
...
                                      unsigned long long genid,
-                                     unsigned long long stripe_ptr)
+                                     unsigned long long *sc_genids)
 {
-    return bdb_inplace_cmp_genids(bdb_state, genid, stripe_ptr) > 0;
+    int stripe = get_dtafile_from_genid(genid);
+    if (stripe < 0 || stripe >= gbl_dtastripe) {
+        logmsg(LOGMSG_ERROR, "%s: genid 0x%llx stripe %d out of range!\n",
+               __func__, genid, stripe);
+        return -1;
```
adizaimi: FTR, we might want to abort here -- at least abort schemachange maybe?

schemachange/sc_callbacks.c
```diff
@@ -163,9 +163,33 @@ static int reload_rowlocks(bdb_state_type *bdb_state, scdone_t type)
...
+    if (!sc_genids[stripe]) {
+        /* A genid of zero is invalid.  So, if the schema change cursor is at
+         * genid zero it means pretty conclusively that it hasn't done anything
+         * yet so we cannot possibly be behind the cursor. */
+        return 1;
+    }
+    return bdb_inplace_cmp_genids(bdb_state, genid, sc_genids[stripe]) > 0;
+}
+
+unsigned long long get_genid_stripe_pointer(unsigned long long genid,
```
adizaimi: Can you make this inline?

schemachange/sc_records.c
```diff
@@ -2713,7 +2713,7 @@ static int live_sc_redo_update(struct convert_record_data *data, DB_LOGC *logc,
 
     if (!data->s->sc_convert_done[rec->dtastripe] &&
         is_genid_right_of_stripe_pointer(data->to->handle, genid,
```
adizaimi: Should we have two functions, one that takes the stripe ptr value just like before, and the new one that figures out the stripe from genid?

ldennis: I thought about that.. but the stripe info from genid should be the same as the one passed in. else we have other issues

ldennis:
cdb2test   Jan 16 15:26:03 2019  in-comdb2db more_constraints.R20190116.4

PR #1387, opened 2019-01-17
Make it easier to debug pthread mutex/rwlock issues...

mistachkin:
These changes are designed for use in a debug build.  They should make it easier to debug issues related to pthread mutexes and rwlocks held by threads.


adizaimi:
This looks great so far. I will change the check to ommit pthread_wrap so you wont get the noise printed above. 
In future we can write a deadlock detector on the locks when db warchdogs for instance.

mistachkin:
Cool, thanks for the review.  I'm pretty happy with how this eventually turned out.  Deadlock detection is something of an open question.  Doing fully generalized N-way deadlock detection is an open computing problem that involves some rather complex algorithms.

tests/docker/Dockerfile.install
```diff
@@ -45,7 +45,7 @@ RUN cd /comdb2 &&  \
     rm -rf build && \
     mkdir build && \
     cd build && \
-    cmake -DWITH_TCL=1 .. && \
+    cmake -DCMAKE_BUILD_TYPE=Debug -DWITH_TCL=1 .. && \
     make -j4 && \
```
mistachkin: TODO: Back this out before merging.

util/dbg_locks.c
```diff
@@ -0,0 +1,562 @@
...
+  const char *func;
+  int line;
+};
+
+typedef struct dbg_lock_pthread_outer_pair_t outer_pair_t;
+typedef struct dbg_lock_pthread_inner_key_t inner_key_t;
+typedef struct dbg_lock_pthread_inner_pair_t inner_pair_t;
+#endif /* defined(DBG_PTHREAD_LOCKS) */
+
+int gbl_debug_pthread_locks = 0;
```
adizaimi: maybe move this to the top so no break in the ifdef

cdb2api/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ set_source_files_properties(
 include_directories(
   ${PROJECT_SOURCE_DIR}/bbinc
   ${PROJECT_BINARY_DIR}/protobuf
+  ${PROJECT_SOURCE_DIR}/util
```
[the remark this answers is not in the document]
mistachkin: Apparently, this is due to sbuf2.c (which is shared) needing to include the header(s) for pthread stuff.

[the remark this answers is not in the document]
mistachkin: It needs this include directory due to the shared source file "sbuf2.c".

tests/random_osql_replay.test/runit
```diff
@@ -17,10 +17,10 @@ cdb2sql ${CDB2_OPTIONS} ${DBNAME} default "delete from t1 where a = 1"
 # Enable random restarts
 if [[ -n $CLUSTER ]]; then
     for x in $CLUSTER ; do
-        cdb2sql ${CDB2_OPTIONS} ${DBNAME} --host $x "put tunable osql_random_restart 'on'"
+        cdb2sql ${CDB2_OPTIONS} ${DBNAME} --host $x "put tunable 'debug.osql_random_restart' 'on'"
     done
```
mistachkin, line 21: TODO: Is this wrong on master?

PR #1388, opened 2019-01-18
fix key return for create view: <cached ddl row different> spew

dorinhogea:
fix new sqlite merge; time partitions started to spew after it (I still think sqlite3BtreeIntegerKey needs further review, since comdb2 doesn't have rowid-s for sqlite_master).


mistachkin:
Is there some way we can create rowIds for sqlite_master, even if they have to be simulated?

PR #1390, opened 2019-01-20
Fix empty blob handling for logical-op-redo schema change.

riverszhang89:
Make sure blob_status_t.blobptrs are not NULL even if the blob size is 0, because `blob_status_to_blob_buffer()` uses `blobptrs` to determine the existence of blobs.

The pull request also fixes a memory overrun in `bdb_pack()` caused by a negative length arugment being passed to LZ4/CRLE compression routine.

We do have a few sc* tests testing empty blobs. Logical SC passes those tests because of the memory overun described above, since blobs are LZ4 compressed by default. Specifically, packing an empty blob (where `odh->length` is 0) results in passing (-1) as the max output length to `LZ4_compress_default()`, which then mistakenly compresses the 0-length blob to size 1, overrunning 1 byte after the 7-byte (ODH header size [7] + blob size [0]) allocated output buffer. When reconstructing the empty blob from the logical operation, its `odh->length` will be 1, hence will be marked existent.

### DIFF
--- a/bdb/odh.c
+++ b/bdb/odh.c
@@ -331,7 +331,7 @@ int bdb_pack(bdb_state_type *bdb_state, const struct odh *odh, void *to,
             to = mallocmem;
         }
 
-        alg = flags & ODH_FLAG_COMPR_MASK;
+        alg = (odh->length) ? flags & ODH_FLAG_COMPR_MASK : BDB_COMPRESS_NONE;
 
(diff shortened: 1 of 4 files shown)


riverszhang89:
Yep - I'm waiting for robomark to finish.

riverszhang89:
@ldennis Thank you for the review!

PR #1391, opened 2019-01-22
Fix get_next_addrem_buffer() abort for data items smaller than genid size.

riverszhang89:
We pass 0 as the key length to `bdb_reconstruct_delete()`. As a result, `get_next_addrem_buffer()` will abort if the data length is smaller than 8 bytes (key is always a genid in a data or blob file).

Obviously, this does not work with an ODH'd 0-length (ODH overhead is 7 bytes) data item in a blob file (e.g. an empty blob), or a single-int-column table which does not support ODH.


riverszhang89:
@ldennis Thank you for the review!

riverszhang89:
```
cdb2test   Jan 21 23:50:20 2019  in-comdb2db fix_noodh_snapshot.R20190121.4
```

PR #1394, opened 2019-01-22
Fix verify for the blob2 type (i.e. inline blob).

riverszhang89:
(DRQS 130232162)


riverszhang89:
@mponomar Thank you for the review!

PR #1397, opened 2019-01-24
Update fingerprinting code to use the SQLite API

mistachkin:


nirbhayc:
If it's meant to test the prepare phase, I'd suggest adding it as a knob in clnt (SET) so that it won't affect other connections.

db/db_fingerprint.c
```diff
@@ -0,0 +1,122 @@
...
+            t->count++;
+            t->cost += cost;
+            t->time += time;
+            t->rows += nrows;
+            assert( memcmp(t->fingerprint,fingerprint,FINGERPRINTSZ)==0 );
+            assert( t->zNormSql!=zNormSql );
+            assert( t->nNormSql==nNormSql );
+            assert( strncmp(t->zNormSql,zNormSql,t->nNormSql)==0 );
+            free(zNormSql);
+        }
```
riverszhang89: You need to call `reqlog_set_fingerprint(logger, fingerprint);` at the end. Otherwise reqlog will not have the fingerprint.

mponomar:
Err, sorry.  I was commenting on the last commit that changes prepare behavior, not the entire PR.

mistachkin:
Things are working much better now.  Would appreciate some review if/when you have time.

mistachkin:
This pull request is now fully passing both test suites.

PR #1398, opened 2019-01-24
PORTED: dont allow adding field without dbstore or nullable

ldennis:
ported from R6


akshatsikarwar:
Should also have test which adds field in the middle of schema (original bug).

akshatsikarwar:
duh

ldennis:
in-comdb2db sc_nullable.R20190124.3

PR #1399, opened 2019-01-24
Send preprocessed blobs to the master.

riverszhang89:
Instead of sending a blob to the master as is, pack and compress the blob on a replicant, and send the preprocessed blob to the master.
This approach significantly reduces the network IO between replicants and the master, as well as the memory footprint and CPU use incurred by compression on the master.

The chart below demonstrates how it works.

```
+-----------+                          +-----------+
| Replicant |                          |   Master  |
+-----------+                          +-----------+
     |                                       |
odhfy_blob_buffer()                          |
     |                                       |
     +--- Add ODH                            |
     +--- Compress                           |
     +--- Mask OSQL_BLOB_ODH_BIT             |
     |    in blob_buffer_t.odhind            |
     |                                       |
osql_send_qblob() -------------------> osql_process_packet()
                                             |
                                             ~
                                             |
                                       bdb_pack(..., odhready=(odhind&OSQL_BLOB_ODH_BIT))
                                             |
                                       is odhready true?
                                            / \
                                        Y  /   \ N
                                          /     \
                                 Use as is       +--- Add ODH
                                                 +--- Compress
```

If a schema changing is modifying the compression algorithm of the blob field, or creating a new expressional index on the blob column, the proprocessed blob must be unpacked on the master.


ldennis:
Looking at record.c, does this work for sp? And we may need more changes internally for oplog based localrep to work? By the way, Mark's physical rep should just work. :+1:

riverszhang89:
It should work inside a SP - `dbtable:insert` internally forms an INSERT statement and passes the statement down to SQLite.

Blobs are "unodhfied" (remove ODH and decompress) first and then are saved into oplog (see [here](https://github.com/bloomberg/comdb2/pull/1399/files#diff-20b8ebf66c0ff339c2b20563d1f0d548R116)) . So it should also work with cdb2rcv.

ldennis:
What about trigger? Do we unodhfied before adding to queue? I was looking for that logic.

riverszhang89:
Gotchat. I guess not...

schemachange/schemachange.c
```diff
@@ -631,6 +633,22 @@ int live_sc_post_add_int(struct ireq *iq, void *trans, unsigned long long genid,
...
+
+        /* If we're removing the ODH, or changing the compression algorithm,
+           unpack the blobs. */
+        if ((oldodh && !newodh) || oldcompr != newcompr) {
+            rc = unodhfy_blob_buffer(iq->usedb, blobs + i, i);
+            if (rc != 0)
+                return rc;
+        }
+    }
+
```
ldennis: Consider unpacking blobs only after checking the stripe pointer so we dont do unnecessary work? Also consider making this logic into a function (so update can reuse)?

db/osqlshadtbl.c
```diff
@@ -1399,13 +1436,15 @@ int osql_save_qblobs(struct BtCursor *pCur, struct sql_thread *thd,
 
         if (blobs[i].exists == 1) {
 
+            (void)odhfy_blob_buffer(pCur->db, blobs + i, i);
```
ldennis: I think this is fine in general. Saving compressed blobs in temp table favors osql resubmission but not transactions (read committed or higher) who read their own inserts. I am not saying this is a bottleneck but it might worth benchmarking.

schemachange/schemachange.c
riverszhang89: Will do.

db/osqlshadtbl.c
riverszhang89: Gotcha. Will do a benchmark and post results in the pull request.

riverszhang89:
## Benchmarks

Write test: Inserting a blob into a LZ4-compressed blob column
Read test: Selecting a pre-processed blob from the shadow table

### Results

| Dataset | Original Size | Compressed Size | Compression Ratio | Write Test Runtime Percent of Change Comparing to Master | Read Test Runtime Percent of Change Comparing to Master
|----------|----------|------|---------|----------------------------------------------------------|--|
| [NYC Bus Delay Open Data](https://data.cityofnewyork.us/Transportation/Bus-Breakdown-and-Delays/ez4e-fazm)|66MiB|18MiB|26.63%| -32% | -77% |
| Comdb2 github repo | 102MiB | 57MiB | 55.95% | -22% | -52% |
| /dev/urandom | 67MiB | 67MiB | 100% | -3%|-5%|

ldennis:
So reading compressed blobs out from shadow is even faster? whoah ..

riverszhang89:
Yeah - Looks like the overhead of saving lengthy records is even greater than the overhead of compression. I will do another benchmark on small blobs.

ldennis:
That makes sense actually.. Temp table needs to do disk io for that size as I think temp table cache is definitely smaller than that ..

riverszhang89:
### Benchmarks Ⅱ

| Dataset | Original Size | Compressed Size | Compression Ratio | Write Test Runtime Change (%) | Read Test Runtime Change (%)
|----------|----------|------|---------|----------------------------------------------------------|--|
| [OATH Trials Division Case Status](https://data.cityofnewyork.us/City-Government/OATH-Trials-Division-Case-Status/y3hw-z6bm)|176KiB|56KiB|29.56%| -77% | +20% |
| /dev/urandom |180KiB | 180KiB | 100% | -8%|+29%|

Reading pre-processed small blobs from the shadow table is slower. However the runtime is neglectable - it is sub-millisecond.

db/osqlshadtbl.c
```diff
@@ -602,20 +608,51 @@ int osql_fetch_shadblobs_by_genid(BtCursor *pCur, int *blobnum,
...
-        blobs->bloblens[0] = bdb_temp_table_datasize(tbl->blb_cur);
+        tmptblblb = bdb_temp_table_data(tbl->blb_cur);
+        tmptblblblen = bdb_temp_table_datasize(tbl->blb_cur);
+
+        if (tmptblkey->odh) {
+            rc = bdb_unpack_heap(pCur->db->handle, tmptblblb, tmptblblblen,
+                                 (void **)blobs->blobptrs, blobs->bloblens,
+                                 &freeptr);
+            if (rc != 0) {
+                free(freeptr);
```
ldennis: Do we double free here? Looks like on errors, `bdb_unpack_heap` also frees `freeptr`

riverszhang89: Yes we do - thanks for catching this!

bdb/odh.c
```diff
@@ -1377,3 +1364,45 @@ inline void bdb_cleanup_fld_hints(bdb_state_type *bdb_state)
...
+int bdb_unpack_heap(bdb_state_type *bdb_state, void *in, size_t inlen,
+                    void **out, size_t *outlen, void **freeptr)
+{
+    struct odh odh;
+    int rc;
+    char odhd = bdb_state->ondisk_header;
+
+    /* Force ODH in case that a schema change
+       is removing the ODH from the table. */
+    bdb_state->ondisk_header = 1;
```
ldennis: Reviewed the potential race issue here in person with @riverszhang89 . Otherwise, looks good in general.

riverszhang89:
Thank you for the review!

adizaimi:
> ### Benchmarks Ⅱ
> Dataset	Original Size	Compressed Size	Compression Ratio	Write Test Runtime Change (%)	Read Test Runtime Change (%)
> [OATH Trials Division Case Status](https://data.cityofnewyork.us/City-Government/OATH-Trials-Division-Case-Status/y3hw-z6bm)	176KiB	56KiB	29.56%	-77%	+20%
> /dev/urandom	180KiB	180KiB	100%	-8%	+29%
> Reading pre-processed small blobs from the shadow table is slower. However the runtime is neglectable - it is sub-millisecond.

What if we could compress only if above a certain size? 32kb or whatever is the cut-over point from your research.

riverszhang89:
Writes are still much faster for small blobs after pre-processing. In addition, constantly reading from a shadow table isn't a common use case. In real life, shadow tables are merged with data from the real btrees. The overhead of reading pre-processed blobs is just a small portion.

PR #1401, opened 2019-01-25
Childhandle

markhannum:
Add the ability to use redundant connections to the cluster via the 'set concurrent' directive.  This allows for instantaneous changeover should an HASQL connection be dropped.


markhannum:
The truncatesc test failure is not related to this code- it's a physical replicant bug.  I will address it under a separate PR.

PR #1402, opened 2019-01-25
used in llmeta load, create sql thread key early

dorinhogea:
trivial fix, currently we call pthread_getspecific before pthread_key_create.


adizaimi:
looks good.
I wonder why we dont get lots of crashes.

PR #1406, opened 2019-01-25
Stat clnt fix

ldennis:
- make send stat clnt output backwards compatible
- ported Mark's argv0 memory leak fix from R6
- patch bbplugin to track stat clnt for fastsql

/plugin-branch stat_clnt


ldennis:
in-comdb2db stat_clnt.R20190125.11

PR #1408, opened 2019-01-26
Allow option to choose pkg type

akshatsikarwar:
/silent


akshatsikarwar:
Sometimes I want to build tarball on a debian based distro for example. This allows me to choose pkg type during configure or later using ccmake or cmake.

PR #1410, opened 2019-01-28
autoanalyze request mode

embiv22:
Added new feature to auto-analyze called "AA_REQUEST_MODE"

When active, the database will print a message to stdout requesting analyze be performed on the appropriate table. 

This is useful for when you want to hand-off running analyze to another task that can schedule running analyze at a time of its choosing.


riverszhang89:
Please /runtests for Ed.

riverszhang89:
Thank you Ed!

PR #1411, opened 2019-01-28
Optimize password authentication.

riverszhang89:
Authenticate the user/password for a new connection instead of every statement. If a user's password has changed, force re-authentication on all existing connections.

In addition, make the number of PBKDF2 iterations configurable.


bdb/llmeta.c
```diff
@@ -8978,10 +8998,11 @@ int bdb_user_password_set(tran_type *tran, char *user, char *passwd)
         return rc;
     key.passwd.file_type = htonl(LLMETA_USER_PASSWORD_HASH);
     passwd_hash data;
-    data.ver = 1;
+    data.ver = 2;
+    data.niterations = gbl_pbkdf2_iterations;
```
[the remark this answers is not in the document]
riverszhang89: Gotcha. Might need to htonl data.ver as well?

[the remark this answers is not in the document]
akshatsikarwar: Not really. Its 1 byte.

bdb/llmeta.c
```diff
@@ -8920,6 +8921,21 @@ static int llmeta_get_user_passwd(char *user, llmetakey_t type, void ***out)
 }
 #define ITERATIONS_V0 1000
 #define ITERATIONS_V1 16 * 1024
+#define ITERATIONS_MIN ITERATIONS_V0
```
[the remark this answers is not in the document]
riverszhang89: The default is 16K. And it turns out to be a bit expensive. We're gonna need a smaller number in production.

db/sqlinterfaces.c
```diff
@@ -3668,8 +3668,14 @@ static int check_sql_access(struct sqlthdstate *thd, struct sqlclntstate *clnt)
     if (gbl_check_access_controls) {
         check_access_controls(thedb);
         gbl_check_access_controls = 0;
+        /* Force all clients to  reauthenticate if passwords have changed. */
+        clnt->authenticated = 0;
```
nirbhayc: How will this force all clients to re-authenticate?

riverszhang89: There's an if branch right below this line, checking if clnt->authenticated is true. If not, perform password check.

nirbhayc: Right, but that only re-auth's this client.

riverszhang89: Once gbl_check_access_controls is set to true, all clients will invalidate the flag. gbl_check_access_controls  is set to true when 1) authentication has been enabled; 2) a user's password has changed.

riverszhang89: Yes you're right - only the 1st client who gets in here will re-authenticate.

PR #1395, opened 2019-01-23
Docs: System tables

nirbhayc:
/skipbuild

PR #1412, opened 2019-01-29
more system tables and refactored cron schedulers

dorinhogea:
New system tables to see the schedulers and what events are queued in a server:

>>> cdb2sql db1 local "select * from comdb2_cron_schedulers"
(name='timepart_cron', type='WALLTIME', running=0, nevents=1, description='Time partition scheduler')
(name='Global Job Scheduler', type='WALLTIME', running=0, nevents=1, description='Default cron scheduler')

>>> cdb2sql db1 local "select name, type, cast(epoch as datetime) 'when', sourceid from comdb2_cron_events"
(name='AddShard', type='timepart_cron', when="2019-01-29T100730.000 America/New_York", sourceid='a952c15a-e44c-43e9-a9ab-fef7952aac62')
(name='Module memory stats update', type='Global Job Scheduler', when="2019-01-29T101029.000 America/New_York", sourceid='00000000-0000-0000-0000-000000000000')

Also:
- switched old timepartition system tables to eztables (and events now reuse the generic cron_events systable)
- renamed views_cron as cron, since it has nothing to do with views or time partitions; updated the api to accept a plugin interface that will allow more than one type of scheduler (a different scheduler is shown in this PR)
- skeleton framework for logical clock "time" partitions (i.e. partitions manually rolled by manually ticking a logical clock); advantages:
--- a logical scheduler can offer the same data retention mechanism as existing time partitions using a simple timer, but are not coerced to roll by a physical clock
--- customers can create any number of schedulers and are in full control of when the rollouts happen, allowing for heterogeneous data retention policies
--- manual clocks will allow to better unit testing by the partition rollouts and recovery;
NOTE: will have a follow-up PR to release the logical partitions as a feature


nirbhayc:
Could you also document the new system tables (https://bloomberg.github.io/comdb2/system_tables.html)? I am documenting the rest https://github.com/bloomberg/comdb2/pull/1395.

dorinhogea:
Sure, I will update the docs; will do once this is merged in and available.

db/cron.c
```diff
@@ -0,0 +1,614 @@
...
+    listc_init(&crons.scheds, offsetof(struct cron_sched, lnk));
+    Pthread_rwlock_init(&crons.rwlock, NULL);
+}
+
+/**
+ * Add a new event to a scheduler.
+ * Create a scheduler if none exists yet
+ *
+ */
+cron_sched_t *cron_add_event(cron_sched_t *sched, const char *name, int epoch,
```
mponomar: Want to split this into 2 functions?  One to create a scheduler, one to add an event to an existing scheduler?

db/cron.c
```diff
@@ -0,0 +1,614 @@
...
+}
+
+static int _queue_event(cron_sched_t *sched, int epoch, FCRON func, void *arg1,
+                        void *arg2, void *arg3, uuid_t *source_id,
+                        struct errstat *err)
+{
+    cron_event_t *event;
+
+    if (!sched) {
+        errstat_set_strf(err, "Folloup called instead of regular event add");
```
mponomar: Error message is a bit confusing.

db/cron.c
```diff
@@ -0,0 +1,614 @@
...
+    int found = 0;
+
+    cron_lock(sched);
+
+    LISTC_FOR_EACH_SAFE(&sched->events, event, tmp, lnk)
+    {
+        if (comdb2uuidcmp(event->source_id, source_id) == 0) {
+
+            /* we can process this */
+            _set_event(event, epoch, func, arg1, arg2, arg3);
```
mponomar: Do we need to free arg1, arg2, arg3, or is that the caller's job?

db/cron.c
```diff
@@ -0,0 +1,614 @@
...
+    systable_cron_scheds_t *arr = NULL, *temparr;
+    cron_sched_t *sched;
+    int narr = 0;
+    int nsize = 0;
+    int rc;
+
+    Pthread_rwlock_rdlock(&crons.rwlock);
+    LISTC_FOR_EACH(&crons.scheds, sched, lnk)
+    {
+        if (narr >= nsize) {
```
mponomar: No need to realloc, can get size from crons.scheds.count

db/cron.c
```diff
@@ -0,0 +1,614 @@
...
+    int narr = *nrecords;
+    cron_event_t *event;
+    int nsize = *pnsize;
+    int rc = 0;
+    uuidstr_t us;
+
+    cron_lock(sched);
+    LISTC_FOR_EACH(&sched->events, event, lnk)
+    {
+        if (narr >= nsize) {
```
mponomar: Same as above, sched->events.count

db/views_systable.c
```diff
@@ -0,0 +1,147 @@
...
+    timepart_view_t *view;
+    systable_timepartition_t *arr = NULL, *temparr;
+    int narr = 0;
+    int nsize = 0;
+    int rc = 0;
+    uuidstr_t us;
+
+    Pthread_rwlock_rdlock(&views_lk);
+    for (narr = 0; narr < views->nviews; narr++) {
+        if (narr >= nsize) {
```
mponomar: Allocate once?  You know the number of views.

db/cron.c
dorinhogea: We don't block schedulers here while we collect events, so we don't know ahead of time how long the list it will be.

PR #1416, opened 2019-01-30
Wrlock netinfo ptr when cleaning up

adizaimi:
Wrlock netinfo ptr when cleaning up.
print_net_memstat() to get netinfo_ptr lock.
Increase timeout for verify.


net/net.c
```diff
@@ -6318,18 +6325,15 @@ int net_set_max_bytes(netinfo_type *netinfo_ptr, uint64_t x)
 int net_sanctioned_list_ok(netinfo_type *netinfo_ptr)
```
akshatsikarwar: This looks wrong. You need to break on the first bad node found. Otherwise, answer is the state of the last node in sanc list.

akshatsikarwar: Oh wait, you made it harder to read when you rolled that into the for-loop ;)

adizaimi:
```suggestion
int net_sanctioned_list_ok(netinfo_type *netinfo_ptr)
```

adizaimi: Sorry, will move the ok before the ptr check.

PR #1417, opened 2019-01-30
Fix a memory leak in remsql.

riverszhang89:
The remote database leaks [a tiny bit of memory](https://github.com/bloomberg/comdb2/blob/master/plugins/remsql/fdb_comm.c#L1002) every time it opens a remote cursor.

(DRQS 120637632)


dorinhogea:
Looks good to me, thank you.

riverszhang89:
Thank you for the review!

PR #1418, opened 2019-01-30
Refactor constraint.c and other constrait related files

adizaimi:
Refactored a bit constraint.c functions to make them easier to understand
Moved structures local into constraint.c
Added wrappers to checking flags in record.c for adding record from schema and for checking has constraint


adizaimi:
Thank you Lingzhi.

PR #1421, opened 2019-02-01
PORTED: IS_ZERO_LSN should check (file == 0 && offset == 0)

ldennis:
(Ported from R6)


ldennis:
Thanks @akshatsikarwar

PR #1422, opened 2019-02-01
osql: Skip saving OSQL ops for single statements.

riverszhang89:
Do not save OSQL operations to shadow tables for a single statement unless it runs under read committed or above.

### DIFF
--- a/db/osqlsqlthr.c
+++ b/db/osqlsqlthr.c
@@ -268,6 +268,8 @@ int osql_delrec(struct BtCursor *pCur, struct sql_thread *thd)
                    __LINE__, __func__, rc);
             return rc;
         }
+        if (clnt->ctrl_sqlengine == SQLENG_NORMAL_PROCESS)
+            return 0;
     }
 
     if (gbl_expressions_indexes && pCur->db->ix_expr) {
@@ -389,6 +391,9 @@ int osql_insrec(struct BtCursor *pCur, struct sql_thread *thd, char *pData,
                    __LINE__, __func__, rc);
             return rc;
         }
+
+        if (clnt->ctrl_sqlengine == SQLENG_NORMAL_PROCESS)
+            return 0;
     }
 
     if (gbl_expressions_indexes && pCur->db->ix_expr) {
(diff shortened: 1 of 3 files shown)


riverszhang89:
oltp_insert runs 15% faster with this simple change.

ldennis:
How does this handle master swings?

riverszhang89:
You're right - it does not. We still need to send commit separately from insrec.

adizaimi:
We have a test, restart_socksql.test, which you should try -- test randomly fails to deliver to master so we retry. It would seem that if you dont save, you cant retry.
Try running socksql_master_swings.test a large number of times, i think it will also fail.

ldennis:
This is still useful though. As master swings do not happen that often, in the case when it does happen, maybe we can somehow treat this as a "verify retry" and so we can replay the sql?

riverszhang89:
I thought of that. The problem is that we would need to change sqlinterfaces layer more aggressively to permit SELECTV's to retry. I guess I could try that.

ldennis:
At least for single statements, we dont have the SELECTV problems.

riverszhang89:
Oh ... I wasn't aware of that. Then it is totally do-able.

db/osqlsqlthr.c
```diff
@@ -800,6 +809,12 @@ int osql_sock_restart(struct sqlclntstate *clnt, int maxretries,
         cheap_stack_trace();
     }
 
+    if (clnt->ctrl_sqlengine == SQLENG_NORMAL_PROCESS) {
+        osql->single_stmt_retry = 1;
+        osql->xerr.errval = (ERR_BLOCK_FAILED + ERR_VERIFY);
+        return 0;
```
ldennis: We can probably get rid of `single_stmt_retry` if we return nonzero rcode (e.g. `SQLITE_ABORT`) here. With this approach, we may also need to change `sock_restart_retryable_rcode` to not retry in this case. I think sqlinterfaces.c should be able to handle these correctly as long as we have `osql->xerr.errval` correct (see `rc_sqlite_to_client()`).

riverszhang89: I added the flag so that we don't print an error on a fake single-statement verify-retry (see [here](https://github.com/bloomberg/comdb2/pull/1422/files#diff-0197042d98ced8aebf540e5e208649afR1150)).

ldennis: Is it easier to return nonzero rcode here?

riverszhang89:
The overhead of saving an OSQL opcode is now negligible with https://github.com/bloomberg/comdb2/pull/1521. Closing.

PR #1423, opened 2019-02-01
Handling of TK_FLOAT for remote cursors

mistachkin:
These change are designed to allow the TK_FLOAT token type to be handled for the remote cursor subsystem, i.e. sqlite3ExprDescribe_inner, et al.


dorinhogea:
This looks good. Please add a regression test, thanks!

PR #1427, opened 2019-02-07
Turn eventlog off by default.

mponomar:


adizaimi:
I'd suggest you put this under legacy defaults. This is needed for the testsuite.

PR #1429, opened 2019-02-08
Forbid remote admin

markhannum:
Port the forbid_remote_admin tunable back to R7: by default we will only allow 'admin' connections which originate from localhost.


markhannum:
Eventlog failures are unrelated

markhannum:
This adds a simple tunable, enabled by default, which forbids admin requests from remote machines.  Additionally, this fixes the threadpool logic- rather than consume from the queue, a thread will exit if it determines that the number of threads in the threadpool is beyond the maximum allowed.

PR #1432, opened 2019-02-12
Parallel sql with params

dorinhogea:
This is the next iteration for parallel sql: support for bound parameters.
The patch packages the client provided arguments in per-worker chunks that can be accessed contention-free, and handles the necessary remapping of un-named parameters in the statement string ('?' and ?'NNN'). Supporting named parameters is trivial ('@Name', ':Name', '$Name').


sqlite/src/sqliteInt.h
```diff
@@ -4382,6 +4382,8 @@ u64 sqlite3LogEstToInt(LogEst);
 VList *sqlite3VListAdd(sqlite3*,VList*,const char*,int,int);
 const char *sqlite3VListNumToName(VList*,int);
 int sqlite3VListNameToNum(VList*,const char*,int);
+VList *sqlite3VListClone(const VList*, void* (*)(size_t));
+void sqlite3VListPrint(loglvl lvl, const VList *pIn);
```
mistachkin: Ideally, these new prototypes should be wrapped in the #if SQLITE_BUILDING_FOR_COMDB2 found elsewhere in the file.

dorinhogea: done

dorinhogea:
thank you

PR #1437, opened 2019-02-13
Auto fix blob inconsistency.

riverszhang89:
Allow verify to fix non-fatal blob inconsistency.

(DRQS 120322231)


riverszhang89:
@akshatsikarwar has reviewed the pull request on bbgh. Thanks!

PR #1440, opened 2019-02-14
sqlite: sleep is not a constant function

akshatsikarwar:
This will make `select sleep(1), * from generate_series limit 10` actually take 10 seconds (1s per row).


riverszhang89:
odh_blobs is intermittently failing on master. I know what the problem is and have fixed it in https://github.com/bloomberg/comdb2/pull/1378/commits/b8768aed15348a3fff6d0d218a67e7ae2f225b53.

PR #1442, opened 2019-02-17
Add osql snapinfo hash for currently running osql session

adizaimi:
Add osql snapinfo hash for currentnly running osql session

### DIFF
--- a/db/comdb2.h
+++ b/db/comdb2.h
@@ -1216,7 +1216,11 @@ typedef struct snap_uid {
     char key[MAX_SNAP_KEY_LEN]; /* cnonce */
 } snap_uid_t;
 
-enum { SNAP_UID_LENGTH = 8 + 4 + (4 * 5) + 4 + 64 };
+enum { SNAP_UID_LENGTH = 16 + 4 + (4 * 5) + 4 + 64 };
+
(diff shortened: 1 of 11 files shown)


db/comdb2.h
```diff
@@ -1194,7 +1194,11 @@ typedef struct snap_uid {
     char key[MAX_SNAP_KEY_LEN]; /* cnonce */
 } snap_uid_t;
 
-enum { SNAP_UID_LENGTH = 8 + 4 + (4 * 5) + 4 + 64 };
+enum { SNAP_UID_LENGTH = 16 + 4 + (4 * 5) + 4 + 64 };
+
+BB_COMPILE_TIME_ASSERT(snap_uid_size,
+                       sizeof(snap_uid_t) == SNAP_UID_LENGTH);
+
```
markhannum: wait, this breaks compatibility, no?

markhannum: !! uid mode is disabled by legacy defaults .. nice ..

adizaimi:
/skipsysbench

trying roborivers one more time

db/comdb2.h
adizaimi: Just to clarify for the record, SNAP_UID_LENGTH was set to the wrong value and here I am just fixing it and adding the compile time assert.

PR #1444, opened 2019-02-19
NEWSI: free pglogs_queue_hash to fix memory leak

ldennis:
.


riverszhang89:
Does hash_free imply hash_clear?

PR #1445, opened 2019-02-20
Port fix string in indexes

dorinhogea:
Fixes:
Changes in upstream sqlite Expr for TK_COLUMN broke remote sql in R7.
Errors parsing to auto-generated strings are now passed back to local server and to the client.
Quoting in predicates containing string values are now properly handled.

### DIFF
--- a/sqlite/src/expr.c
+++ b/sqlite/src/expr.c
@@ -6282,9 +6282,10 @@ default_prec:
     }
     case TK_COLUMN: {
       if( atRuntime ){
-        return sqlite3_mprintf("\"%s\"",  pExpr->u.zToken);
+        assert(pExpr->y.pTab && pExpr->y.pTab->nCol > pExpr->iColumn);
+        return sqlite3_mprintf("\"%s\"",  pExpr->y.pTab->aCol[pExpr->iColumn].zName);
       }else{
-        return sqlite3_mprintf("%s",  pExpr->u.zToken);
+        return sqlite3_mprintf("%s",  pExpr->y.pTab->aCol[pExpr->iColumn].zName);
       }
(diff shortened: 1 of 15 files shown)


dorinhogea:
Follow-up from R6 https://bbgithub.dev.bloomberg.com/comdb2/comdb2/pull/1014.

sqlite/src/expr.c
```diff
@@ -6282,9 +6282,10 @@ static char* sqlite3ExprDescribe_inner(
     }
     case TK_COLUMN: {
       if( atRuntime ){
-        return sqlite3_mprintf("\"%s\"",  pExpr->u.zToken);
+        assert(pExpr->y.pTab && pExpr->y.pTab->nCol > pExpr->iColumn);
+        return sqlite3_mprintf("\"%s\"",  pExpr->y.pTab->aCol[pExpr->iColumn].zName);
       }else{
-        return sqlite3_mprintf("%s",  pExpr->u.zToken);
+        return sqlite3_mprintf("%s",  pExpr->y.pTab->aCol[pExpr->iColumn].zName);
```
mistachkin: I think this should also be %q, e.g.:

sqlite> create table t1 ( "x' -- more stuff here" integer );
sqlite> pragma table_info(t1);
0|x' -- more stuff here|integer|0||0

Column names from the schema can include quotes.

dorinhogea: This does not work in comdb2:
cdb2sql db1 local "create table ttttt (\"x' ha\" integer)"
[create table ttttt ("x' ha" integer)] failed with rc 240 ERROR at line   2: syntax error: '

PR #1447, opened 2019-02-21
Fix bdb temptbltest to correctly set bdberr, measure time for write

adizaimi:
Fix bdb temptbltest to correctly set bdberr, measure time for write.
FTR, to write 10k records of size 8KB each, took almost 1s in bdta machines (which is a bit on the slow side). Writing 10k records took (depending on size): 
1KB 0.306s
2KB 0.409s
4KB 0.599s
8KB 0.956s

### DIFF
--- a/bdb/temptable.c
+++ b/bdb/temptable.c
@@ -32,6 +32,7 @@
 #include "bdb_int.h"
 #include <list.h>
 #include <plhash.h>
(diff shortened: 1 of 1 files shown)


riverszhang89:
A temptable has roughly 512KiB cache. The amount you write exceeds the cache size.

PR #1448, opened 2019-02-21
Give up ongoing transactions on downgrade

ldennis:
Make sure block processor is able to give up outstanding transactions on downgrades


ldennis:
cdb2test   Feb 22 10:46:57 2019  in-comdb2db nicer_downgrade.R20190222.1

PR #1450, opened 2019-02-22
Initialize newsqlheader and other structures with zero bytes

adizaimi:
Initialize newsqlheader and other structures with zero bytes


riverszhang89:
Let's preserve the authorship of the original commit.

riverszhang89:
For example: https://github.com/bloomberg/comdb2/commit/3c22c64357da402c92f1c345515a1ab10c447798

adizaimi:
Good point. Entered separate PR https://github.com/bloomberg/comdb2/pull/1452. 
Thanks Rivers.

PR #1453, opened 2019-02-23
cdb2api: allocate on stack protobuf buffer for msg to server

adizaimi:
On the clients side, allocate on stack buffer used for protobuf message to the server,
instead of mallocing that buffer for every query. If buffer is greater than 8k, use malloc.


cdb2api/cdb2api.c
```diff
@@ -2616,6 +2617,9 @@ static int cdb2_send_query(cdb2_hndl_tp *hndl, cdb2_hndl_tp *event_hndl,
                 __func__, __LINE__);
     }
 
+    static __thread unsigned char *staticbuf = NULL;
+    static __thread size_t staticbuf_size = 0;
+
```
riverszhang89: I'd prefer to use alloca to allocate an array on the stack for queries shorter than 8192 bytes (alloca is cheap). No need to realloc over and over again because 1) it is not efficient for a query longer than 8192 bytes (You'd be doing 2 realloc's and that's potentially 2 malloc's + 2 free's); 2) this is no way to free up the memory. This will make it difficult for the leak checker I'm working on to catch memory leaks correctly.

akshatsikarwar: Agreed. This is not clean. At the very least these should be held in cdb2_hnd so I can pass handle between threads.

riverszhang89:
Good to merge. but please fix the format before doing so.

adizaimi:
Thanks Rivers and Akshat.

PR #1455, opened 2019-02-25
Init structs with zeros rather than bzero content

adizaimi:
Init structs with zeros rather than bzero content


adizaimi:
Thank you Joe.

PR #1461, opened 2019-02-28
datetime: Fix drifting timezone

riverszhang89:
After failing to load a timezone, we don't back out the timezone pointer to the previous value,
but leave it point to the most recently loaded timezone. The following queries reproduce the bug:

```SQL
set timezone EST
select cast(0 as datetime)
-- EST is loaded
set timezone UTC
select cast(0 as datetime)
-- UTC is loaded
set timezone EST
select cast(0 as datetime)
-- timezone changed to EST
select cast("1970-01-01T000000.000 MARS" as datetime)
-- MARS is an invalid timezone. We don't change timezone back to EST
-- but leave it point to the most recently loaded timezone, which is UTC
set timezone EST
select now()
-- We end up being 5 hours ahead of EST.
--
```

The fix is to simply back out the timezone pointer to the previous value if it fails to load a timezone.

(DRQS 138630743)

### DIFF
--- a/tests/datetime.test/t5_01.req
+++ b/tests/datetime.test/t5_01.req
@@ -5,5 +5,6 @@ select cast(0 as datetime)
 set timezone EST
 select cast(0 as datetime)
 select cast("1970-01-01T000000.000 MARS" as datetime)
+select cast(0 as datetime)
 set timezone EST
 select cast(0 as datetime)
--- a/tests/datetime.test/t5_01.req
+++ b/tests/datetime.test/t5_01.req
@@ -1,0 +1,9 @@
+set timezone EST
+select cast(0 as datetime)
+set timezone UTC
+select cast(0 as datetime)
+set timezone EST
+select cast(0 as datetime)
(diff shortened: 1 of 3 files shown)


ldennis:
whoah

tests/datetime.test/t5_01.req
```diff
@@ -0,0 +1,9 @@
+set timezone EST
+select cast(0 as datetime)
+set timezone UTC
+select cast(0 as datetime)
+set timezone EST
+select cast(0 as datetime)
+select cast("1970-01-01T000000.000 MARS" as datetime)
+set timezone EST
```
nirbhayc: You should drop this SET to prove that the TZ was correctly restored.

riverszhang89: Gotcha Nirbhay.. This is a great suggestion. Sorry it took me a while to understand your point.

riverszhang89:
@nirbhayc @mistachkin Thank you for the review.

PR #1462, opened 2019-02-28
DDL: Fix to retrieve/escape index on expressions correctly

nirbhayc:
.


nirbhayc:
@mistachkin Thanks for reporting this issue!

PR #1463, opened 2019-02-28
CSC2: Fix csc2 parser to accept named constraint and modifiers together

nirbhayc:
.


nirbhayc:
@ldennis Thanks for reporting this issue.

PR #1464, opened 2019-03-01
Prevent getdbrelpath() from causing a SEGFAULT when lrlname is not set

mistachkin:


adizaimi:
How can lrlname be null (under what condition)? Is there a testcase we can add?
Looks good in general.
Thanks.

mistachkin:
It can be null when initresourceman() has never been successfully completed.

PR #1468, opened 2019-03-05
Always allow subsequent queries from open transactions to be queued.

riverszhang89:
If a transaction has begun on a replicant, always allow subsequent queries from the transaction to be queued on the replicant.

(DRQS 138604354)


riverszhang89:
Mike reviewed it internally. Thanks!

PR #1471, opened 2019-03-07
Wrap few more newly introduced locks

adizaimi:
Wrap few more newly introduced locks.
Also remove duplicate file assert_locks which does the same thing as locks_wrap.h


adizaimi:
The insert_lots test timed out because was too short a timeout.

PR #1472, opened 2019-03-08
Add ability to whitelist databases allowed to connect to a given db

adizaimi:
Add ability to whitelist databases allowed to connect to a given db.
Db list is passed in lrl in the form `remsql_whitelist databases d1 d2 d3`. If an other DB makes a connection to *this* DB, we will only continue processing request if that DB's name is in the whitelist, otherwise it will receive an error. 
Also added ability to send message to db to add/delete/dump to whitelist content.


db/config.c
```diff
@@ -690,6 +691,31 @@ static int read_lrl_option(struct dbenv *dbenv, char *line,
             dbenv->sibling_port[0][NET_REPLICATION] = port;
             dbenv->sibling_port[0][NET_SQL] = port;
         }
+    } else if (tokcmp(tok, ltok, "fdb_whitelist") == 0) {
+        /* expected parse line: 
+         * fdb_whitelist nodes node1 node2 ... databases db1 db2 ...
+         */
+        tok = segtok(line, len, &st, &ltok);
+#if 0
```
riverszhang89: Can we please get rid of commented-out code?

db/fdb_access.c
```diff
@@ -311,3 +317,65 @@ static int _access_control_clear(fdb_access_t *acc)
...
+
+    hash_for(fdb_dbname_hash, (hashforfunc_t *)dump_whitelist, NULL);
+}
+
+/* Check if parameter dbname is in whitelist
+ * Note that dbname can be dbname or uri (dbname@hostname) 
+ */
+int fdb_is_dbname_in_whitelist(const char *name)
+{
+    logmsg(LOGMSG_ERROR, "%s: name=%s\n", __func__, name);
```
riverszhang89: You probably do not want to print a line of trace every time this function gets hit.

db/fdb_fend.c
```diff
@@ -2181,6 +2181,15 @@ static int _fdb_send_open_retries(struct sqlclntstate *clnt, fdb_t *fdb,
         }
 
         if ((rc = _fdb_remote_reconnect(fdb, psb, host, (fdbc)?1:0)) == FDB_NOERR) {
+
+            /*
```
riverszhang89: No commented-out code.

db/process_message.c
```diff
@@ -75,11 +75,12 @@ extern int __berkdb_read_alarm_ms;
 #include "crc32c.h"
 #include "ssl_bend.h"
 #include "dohsql.h"
+#include "fdb_whitelist.h"
 
-#include <trigger.h>
-#include <sc_stripes.h>
-#include <sc_global.h>
-#include <logmsg.h>
+#include "trigger.h"
```
riverszhang89: Why do we replace brackets with quotes here? Are these files in db directory?

plugins/remsql/fdb_comm.c
```diff
@@ -688,7 +690,7 @@ void fdb_msg_clean_message(fdb_msg_t *msg)
         break;
 
     case FDB_MSG_CURSOR_OPEN:
-        if (msg->co.srcname != gbl_myhostname) {
+        if (msg->co.srcname != gbl_myhostname && msg->co.srcname != gbl_myuri) {
```
riverszhang89: just if (msg->co.srcname != gbl_myuri)

db/process_message.c
adizaimi: They are local to the comdb2 dir. I believe good practice is to have only system includes in <>.

riverszhang89: Not really. The convention is to use quotes for headers in the same directory as the source file and use brackets for the rest.

adizaimi: http://www.cplusplus.com/forum/general/24564/
<> tells compiler  to look for file in system dirs first.

riverszhang89: https://github.com/torvalds/linux/blob/master/drivers/android/binder.c#L54-L83

db/config.c
adizaimi: Yep.

db/fdb_access.c
adizaimi: agreed.

db/fdb_fend.c
adizaimi: Yep.

plugins/remsql/fdb_comm.c
adizaimi: I think i was crashing without msg->co.srcname != gbl_myhostname, but it's harmless, so i'm just going to leave it if I may.

plugins/remsql/fdb_comm.c
```diff
@@ -1014,6 +1016,9 @@ int fdb_msg_read_message(SBUF2 *sb, fdb_msg_t *msg, enum recv_flags flags)
         } else {
             msg->co.srcname = NULL;
         }
+
+        if (!fdb_is_dbname_in_whitelist(msg->co.srcname))
+            return -1;
```
dorinhogea: this has to pass an error and that gets converted to SQLITE_ACCESS

db/comdb2.c
```diff
@@ -633,6 +634,7 @@ int gbl_broken_max_rec_sz = 0;
 int gbl_private_blkseq = 1;
 int gbl_use_blkseq = 1;
 int gbl_reorder_socksql_no_deadlock = 1;
+int gbl_reorder_idx_writes = 1;
```
dorinhogea: what is this?

db/config.c
```diff
@@ -690,6 +691,15 @@ static int read_lrl_option(struct dbenv *dbenv, char *line,
             dbenv->sibling_port[0][NET_REPLICATION] = port;
             dbenv->sibling_port[0][NET_SQL] = port;
         }
+    } else if (tokcmp(tok, ltok, "remsql_whitelist") == 0) {
```
dorinhogea: description mentions:
db_whitelist databases d1 d2 d3

db/fdb_access.c
```diff
@@ -56,6 +59,9 @@ struct fdb_access {
     listc_t lst;
 };
 
+
+static hash_t *fdb_dbname_hash;
```
dorinhogea: this is not mutex protected; we have traps adding/removing from it, and sql access this concurrently

db/fdb_fend.c
```diff
@@ -702,7 +702,7 @@ int fix_table_stats(fdb_t *fdb, fdb_tbl_t *tbl, const char *stat_name)
     /* we need to move this from ent->tbl->ents to tbl_stat->ents */
     listc_rfl(&stat_ent->tbl->ents, stat_ent);
     stat_ent->tbl = stat_tbl;
-    stat_ent->tbl->version = stat_ent->_version;
+   stat_ent->tbl->version = stat_ent->_version;
```
dorinhogea: format?

db/fdb_fend.c
```diff
@@ -4927,7 +4927,9 @@ int fdb_get_remote_version(const char *dbname, const char *table,
         flags = 0;
     }
 
-    snprintf(sql, sizeof(sql), "select table_version(\'%s\')", table);
+    int tot = snprintf(sql, sizeof(sql), "select table_version(\'%s\')", table);
```
dorinhogea: this is too low to check for user provided dbname length; if you're doing it here, make sure we pass back -3 (syntax error or similar)

db/process_message.c
```diff
@@ -2550,6 +2551,30 @@ int process_command(struct dbenv *dbenv, char *line, int lline, int st)
...
+            logmsg(LOGMSG_ERROR, "Expected action\n");
+            return -1;
+        }
+        if (tokcmp(tok, ltok, "add") == 0) {
+            tok = segtok(line, lline, &st, &ltok);
+            if (ltok == 0) {
+                logmsg(LOGMSG_ERROR, "add: expected db name\n");
+                return -1;
+            }
+            fdb_add_dbname_to_whitelist(tok);
```
dorinhogea: unprotected

db/process_message.c
```diff
@@ -2550,6 +2551,30 @@ int process_command(struct dbenv *dbenv, char *line, int lline, int st)
...
+        } else if (tokcmp(tok, ltok, "del") == 0) {
+            tok = segtok(line, lline, &st, &ltok);
+            if (ltok == 0) {
+                logmsg(LOGMSG_ERROR, "del: expected db name\n");
+                return -1;
+            }
+            fdb_del_dbname_from_whitelist(tok);
```
dorinhogea: unprotected

tests/simple_remsql.test/runit
```diff
@@ -27,6 +27,41 @@ for file in $files ; do
 done
 
 
+if [[ $DBNAME == *"whitelistgenerated"* ]] ; then
```
dorinhogea: test output and error code when access is denied

db/config.c
adizaimi: will change description, thanks.

db/fdb_access.c
adizaimi: Adding mutex. Initially I did not think this is going to be used that much anyway, but of course you are right : )

db/fdb_fend.c
adizaimi: Well, I was just trying to get to protect against bad tablename being passed in. Since it will not be common at all, I figured generic error would suffice. Can we agree to leave it with generic or return FDB_ERR_FDB_TBL_NOTFOUND instead. -3 is not syntax error and, there is no other code that seems close to syntax error.

tests/simple_remsql.test/runit
adizaimi: Added testcase for access denied. Returning correct error msg for access error.

plugins/remsql/fdb_comm.c
adizaimi: Done.

db/fdb_fend.c
adizaimi: fixed.

db/process_message.c
adizaimi: added locks

db/process_message.c
adizaimi: added locks

db/fdb_fend.c
adizaimi: Returning FDB_ERR_FDB_TBL_NOTFOUND.

plugins/remsql/fdb_comm.c
adizaimi: you are right. that works actually.

adizaimi:
Thanks Rivers and Dorin.

PR #1474, opened 2019-03-11
Fix 'REBUILD INDEX' triggering an assert() at the top of add_ongoing_alter().

mistachkin:


ldennis:
Looks good

adizaimi:
This looks good.

PR #1475, opened 2019-03-12
For use running tests on master via the debug build

mistachkin:


db/osqlcomm.c
```diff
@@ -4814,9 +4814,11 @@ int osql_send_commit_by_uuid(char *tohost, uuid_t uuid, int nops,
             sbuf2flush(logsb);
         }
 
+#if 0
```
mistachkin: Without using #if 0 here, this output ends up as stray in one of the test cases (random_osql_replay).

nirbhayc:
Are you waiting on review? If not, could you close this PR if you don't want this to be merged?

PR #1477, opened 2019-03-14
Logical partitions

dorinhogea:
This is an upgrade to existing time partitions.  It replaces the time dependent rollout with a logical clock.
The advantages of this are:
1) using a time cron job to roll the logical clock turns this in the previous time partition scheme
2) table owners can decide to rollout based on a variety of criteria now: ad-hoc, table size, contention, and so on.
3) the logical clock is implemented using a persistently replicated named counter, that can be used for other purposes as well
4) the rollout can be controlled to recreate various race scenarios for testing purposes
5) manual rollouts with a time cron job replaces previous efforts to temporarily stop the rollout (cases: when a backout is restored, or when client wants to accumulate more data in the current shard)
Syntax:
PUT COUNTER <name> SET <value>
PUT COUNTER <name> INCREMENT
Behind the scene this is supported by a regular comdb2 table, comdb2_logical_cron, and upsert operations.


dorinhogea:
Need bb-plugin change:
https://bbgithub.dev.bloomberg.com/comdb2/bb-plugins/pull/71

dorinhogea:
Thank you Mohit!

sqlite/src/parse.y
```diff
@@ -2202,11 +2214,14 @@ cmd ::= createkw RANGE PARTITION ON nm(A) WHERE columnname(B) IN LP exprlist(C)
     comdb2CreateRangePartition(pParse, &A, &B, C);
 }
 
-cmd ::= createkw TIME PARTITION ON nm(A) AS nm(P) PERIOD STRING(D) RETENTION INTEGER(R) START STRING(S). {
+cmd ::= createkw partition_type PARTITION ON nm(A) AS nm(P) PERIOD STRING(D) RETENTION INTEGER(R) START STRING(S). {
```
nirbhayc: @dorinhogea : Looks like you forgot to reflect this change in docs.

PR #1478, opened 2019-03-14
Fix some issues caught by clang

adizaimi:
Fix some issues [and potential bugs] caught by warnings produced by clang in linux.


adizaimi:
note, a more portable byteswap example can be found here: https://github.com/google/cityhash/blob/8af9b8c2b889d80c22d6bc26ba0df1afb79a30db/src/city.cc#L50

sqlite/src/vdbe.h
```diff
@@ -35,7 +35,9 @@ typedef struct Vdbe Vdbe;
 ** The names of the following types declared in vdbeInt.h are required
 ** for the VdbeOp definition.
 */
+#if !defined(SQLITE_BUILDING_FOR_COMDB2)
 typedef struct sqlite3_value Mem;
```
mistachkin: I'm a bit confused about this change.  What was the warning this fixed?

adizaimi: double define of typedef (only C11 feature). Mem was defined in one of the includes.

akshatsikarwar:
Please make sure this builds on macOS before merging.

PR #1479, opened 2019-03-19
Stop using TK_SELECTV from code, instead relying on the 'recording' flag.

mistachkin:
These changes are an experiment to simplify our integration with SQLite.  Given that use of the SELECTV command in SQL always sets the 'recording' flag, we should be able to rely on that.  **It should also be noted that the actual TK_SELECTV token needs to be retained for use by the parser.**


mistachkin:
This pull request passes both test suites.

PR #1482, opened 2019-03-26
Fix osql_shadtbl_usedb_only()

riverszhang89:
The bug causes SELECTV-only transactions to behave incorrectly under DISABLE_SELECTVONLY_TRAN_NOP.
Also port Lingzhi's snapshot-read fix.


riverszhang89:
The pull request passes robomark as well.

PR #1483, opened 2019-04-01
Add new tunable for use in deadlock detection and avoidance

mistachkin:


adizaimi:
This looks good. The testsuite should catch the test added timing out--need fixing.

PR #1486, opened 2019-04-02
Fix parallel sql leak

dorinhogea:
bound parameter are repackaged for each parallel thread, and that was leaked in a few scenarios.

### DIFF
--- a/db/dohast.c
+++ b/db/dohast.c
@@ -321,8 +321,7 @@ static dohsql_node_t *gen_oneselect(Vdbe *v, Select *p, Expr *extraRows,
 
     if (!node->sql) {
         if (node->params) {
-            if (node->params->params)
-                free(node->params->params);
+            free(node->params->params);
             free(node->params);
         }
         free(node);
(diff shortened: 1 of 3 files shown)


db/dohast.c
```diff
@@ -337,6 +342,13 @@ static void node_free(dohsql_node_t **pnode, sqlite3 *db)
         node_free(&node->nodes[i], db);
     }
 
+    /* params */
+    if ((*pnode)->params) {
+        if ((*pnode)->params->params)
+            free((*pnode)->params->params);
```
riverszhang89: You don't need to check for NULL. `free(NULL)` is a no-op.

db/dohsql.c
```diff
@@ -1087,6 +1087,8 @@ static void _shard_disconnect(dohsql_connector_t *conn)
     if (conn->cols)
         free(conn->cols);
 
+    if (conn->params)
+        free(conn->params);
```
riverszhang89: ditto.

db/dohsql.c
```diff
@@ -1879,8 +1881,10 @@ struct params_info *dohsql_params_append(struct params_info **pparams,
     newparam = clnt_find_param(params->clnt, name + 1, index);
     if (!newparam) {
         /* clnt parameters are incorrect, fallback to single thread to err */
-        free(params->params);
+        if (params->params)
+            free(params->params);
```
riverszhang89: ditto

db/dohast.c
dorinhogea: I didn't want to make assumptions about our code that overrides free, but sure.

dorinhogea:
thank you

PR #1489, opened 2019-04-03
set memptricklepercent to 90 to avoid bottleneck and gain some performance

adizaimi:
set memptricklepercent to 90 to avoid bottleneck and gain some performance

### DIFF
--- a/bdb/attr.h
+++ b/bdb/attr.h
@@ -121,7 +121,7 @@ DEF_ATTR(DIRECTIO, directio, BOOLEAN, 1,
    evicted cheaply - if there are only dirty blocks, a read will cause
    a write, as the read has no cache blocks to use, and the eviction will
    result in a write. */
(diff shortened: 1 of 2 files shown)


riverszhang89:
https://github.com/bloomberg/comdb2/pull/1441

In fact the gain wasn't noticeable in sysbench. Mike said that we made it 99 for a reason.

nirbhayc:
oltp_read_only seemingly has gotten worse, please keep an eye on that.

adizaimi:
Hmm. Sysbench probably has other constraints so this change does not improve any of its tests. We can pstack the sysbench tests specifically to know what they need to improve performance. It seems that the read workload may be suffering from this here?

riverszhang89:
I don't think we fully understand the performance impact of memp_trickle yet. When you ran your benchmark, how big was the cache size configured to?

adizaimi:
Was small, 64MB.

Issue #1404, opened 2019-01-25
Auth: Inconsistencies with sqlite_statN tables


nirbhayc:
```
# Create an OP and a non-OP user
nctdb> put password 'aaa' for 'aaa'
[put password 'aaa' for 'aaa'] rc 0
nctdb> put password 'bbb' for 'bbb'
[put password 'bbb' for 'bbb'] rc 0
nctdb> grant op to aaa;
[grant op to aaa] rc 0

# Switch to OP user
nctdb> set user aaa
[set user aaa] rc 0
nctdb> set password aaa
[set password aaa] rc 0
nctdb> create table t1(i int)$$
[create table t1(i int)] rc 0
nctdb> put authentication on
[put authentication on] rc 0
# OP sees both sqlite_stat tables
nctdb> select * from comdb2_tables;
+----------------+
| tablename      |
+----------------+
| 'sqlite_stat1' |
| 'sqlite_stat4' |
| 't1'           |
+----------------+
[select * from comdb2_tables] rc 0
# OP can read sqlite_stat1 (also evident from comdb2_tablepermissions below)
nctdb> select * from sqlite_stat1;
+-----+-----+------+
| tbl | idx | stat |
+-----+-----+------+
[select * from sqlite_stat1] rc 0
nctdb> select * from comdb2_tablepermissions where tablename like 'sqlite%';
+----------------+----------+------+-------+-----+
| tablename      | username | READ | WRITE | DDL |
+----------------+----------+------+-------+-----+
| 'sqlite_stat1' | 'aaa'    | 'Y'  | 'Y'   | 'Y' |
| 'sqlite_stat1' | 'bbb'    | 'N'  | 'N'   | 'N' |
| 'sqlite_stat4' | 'aaa'    | 'Y'  | 'Y'   | 'Y' |
| 'sqlite_stat4' | 'bbb'    | 'N'  | 'N'   | 'N' |
+----------------+----------+------+-------+-----+
[select * from comdb2_tablepermissions where tablename like 'sqlite%'] rc 0

# Switch to non-OP
nctdb> set user bbb
[set user bbb] rc 0
nctdb> set password bbb
[set password bbb] rc 0
# Only sqlite_stat1 shows under comdb2_tables
nctdb> select * from comdb2_tables;
+----------------+
| tablename      |
+----------------+
| 'sqlite_stat1' |
+----------------+
[select * from comdb2_tables] rc 0
# No sqlite_stat tables shows under comdb2_tablepermissions
[...]

PR #1490, opened 2019-04-04
Auth: Fix an inconsistency in system table results

nirbhayc:
Fixes: https://github.com/bloomberg/comdb2/issues/1404


nirbhayc:
Thanks @riverszhang89 !

PR #1497, opened 2019-04-08
remsql: Fix some segfaults

nirbhayc:
.


dorinhogea:
Is there a test case you can add here?

PR #1500, opened 2019-04-10
schemachange: Fix the schema change resumption check for comdb2sc

riverszhang89:
`sorese_info_t.rqid` is always 0 for a schema change from comdb2sc.
Hence the current logic would allow multiple schema changes against the same table, causing deadlocks.

(DRQS 140731509)


ldennis:
whoah

riverszhang89:
Right. Thank you for the explanation and the review!

PR #1502, opened 2019-04-12
Remote db class overrides

dorinhogea:
Remove redundant code when checking machine/database class.  A more generic lrl option to define the classes, and specify the intended class for the database.


dorinhogea:
thank you for review!

PR #1504, opened 2019-04-16
Reduce contention on the private blkseq.

riverszhang89:
Increase PRIVATE_BLKSEQ_STRIPES to 8 (same as dtastripe). Also fix the crashes when PRIVATE_BLKSEQ_STRIPES is greater than 1.

### DIFF
--- a/bdb/bdb_int.h
+++ b/bdb/bdb_int.h
@@ -1019,7 +1019,8 @@ struct bdb_state_tag {
     DB **blkseq[2];
     time_t blkseq_last_roll_time;
     DB_LSN *blkseq_last_lsn[2];
-    LISTC_T(struct seen_blkseq) blkseq_log_list[2];
+    listc_t *blkseq_log_list;
+    int pvt_blkseq_stripes;
     uint32_t genid_format;
 
     /* we keep a per bdb_state copy to enhance locality */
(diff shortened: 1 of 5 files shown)


bdb/bdb_int.h
```diff
@@ -1019,7 +1019,8 @@ struct bdb_state_tag {
     DB **blkseq[2];
     time_t blkseq_last_roll_time;
     DB_LSN *blkseq_last_lsn[2];
-    LISTC_T(struct seen_blkseq) blkseq_log_list[2];
+    listc_t *blkseq_log_list;
+    int pvt_blkseq_stripes;
```
nirbhayc: Why would you need this new variable?

riverszhang89: `bdb_state->pvt_blkseq_stripes` is just a copy from `bdb_state->attr->private_blkseq_stripes`. It is faster than accessing `bdb_state->attr`. It also make sure that we don't end up accessing invalid memory if  `bdb_state->attr->private_blkseq_stripes` is increased at runtime.

riverszhang89:
Roboperf reports significant improvement on sysbench/oltp_insert.

adizaimi:
Nice job Rivers.
Looks good to me.

riverszhang89:
@adizaimi & @nirbhayc Thank you for the reviews!

PR #1507, opened 2019-04-16
Fix mismatching remsql db clases

dorinhogea:
Currently, it is possible to skip class overrides or even remote db names when pointing to a remote table, after an initial correct access:
"select * from LOCAL_dbname.tblname"
"select * from dbname.tblname" 
"select * from tblname"
Worse, this is also possible because the table-lookup fast path skips the class name validation:
"select * from PROD_dbname.tblname"
"select * from NOTYOURFOO_dbname.tblname"
Patch addresses this issues.


dorinhogea:
@mistachkin thank you for review!

PR #1508, opened 2019-04-16
reqlog: less spew from excess client stats.

riverszhang89:
Instead of printing an error for every new client that's beyond `gbl_max_clientstats_cache`, do so every `gbl_max_clientstats_cache` clients.

(DRQS 138722214)


riverszhang89:
Thank you!

PR #1509, opened 2019-04-17
fix tunables that magically passed before

dorinhogea:
PR https://github.com/bloomberg/comdb2/pull/1502 regression misreported this tunables as passing.  Fixing it here.


dorinhogea:
thank you

PR #1511, opened 2019-04-18
Revert "Authorizer callback for use with SQLite, based on PR #1305 (#1501)"

akshatsikarwar:
This reverts commit 618360a5307b41219503cedad2cd56e0993096c8.

Temporary revert to get passing again.


mponomar:
Ok - let's revert it for now.

PR #1512, opened 2019-04-18
Fix pglogs_asof_thread race with clean exit.

riverszhang89:
pglogs_asof_thread accesses bdb_state in a loop. Ensure that clean exit waits for pglogs_asof_thread to exit before shutting down the bdb env.

(DRQS 141223321)


riverszhang89:
Thank you for the review!

PR #1515, opened 2019-04-18
Create osql hash before starting net

mponomar:


riverszhang89:
I've also attempted to fix the crash in https://github.com/bloomberg/comdb2/pull/1513. Yours works too.

db/sqloffload.c
```diff
@@ -154,6 +154,12 @@ int osql_open(struct dbenv *dbenv)
         return -2;
     }
 
+    rc = osql_blkseq_init();
+    if (rc) {
+        fprintf(stderr, "%s failing to init blocksql blockseq module\n",
```
riverszhang89: Looks good. Would you mind changing this to logmsg?

PR #1516, opened 2019-04-18
Reopen a temp cursor if it is closed

riverszhang89:
Recent SQLite upgrade changes the order of temptable-related calls. For instance, temp_table_truncate() may be immediately followed by temp_table_find(). The former closes all open cursors while the latter does not verify whether the cursor is open or not.

To fix this, every time before using a temp cursor, we verify and reopen it if needed.


riverszhang89:
Thank you Mike.

PR #1519, opened 2019-04-22
Have the SIGTERM handler call clean_exit_thd

riverszhang89:
To avoid double clean-exits if the process receives SIGTERM (e.g., from can.tsk) while clean-exiting.

(DRQS 141375609)


riverszhang89:
Mike found the bug and Akshat provided the fix. I simply typed.

riverszhang89:
Thank you for the review!

PR #1520, opened 2019-04-23
use dynamic array for osql

adizaimi:
use dynamic array for osql


riverszhang89:
I'd prefer that you base your dynamic array on my https://github.com/bloomberg/comdb2/pull/1521. It'll be less code and have a unified interface with the rest of temptable implementations.

PR #1525, opened 2019-04-24
Use reentrant version of functions like ctime

adizaimi:
Use reentrant version of functions like ctime
/skipsysbench


adizaimi:
Thanks Mike for looking at this.

PR #1526, opened 2019-04-24
make sure we access dev instance of comdb2db when on dev

dorinhogea:
After refactoring class names, dev clusters try to access prod comdb2db and fail.  This patch fixes it.


dorinhogea:
thank you

PR #1529, opened 2019-04-25
Add 'host' and 'master' to the comdb2_sysinfo SQL function

mistachkin:


sqlite/src/func.c
```diff
@@ -755,6 +757,12 @@ static void comdb2SysinfoFunc(
   zName = (const char *)sqlite3_value_text(argv[0]);
   if( sqlite3_stricmp(zName, "pid")==0 ){
     sqlite3_result_int64(context, (sqlite3_int64)getpid());
+  }else if( sqlite3_stricmp(zName, "master")==0 ){
+    sqlite3_result_text(context, thedb->bdb_env->repinfo->master_host, -1,
+                        SQLITE_TRANSIENT);
+  }else if( sqlite3_stricmp(zName, "host")==0 ){
```
mponomar: We have comdb2_host() that returns this.

mistachkin: Cool.  That will provide a useful way to verify that both functions (old and new) return the correct result, e.g. SELECT comdb2_sysinfo('host') == comdb2_host()

mistachkin: Actually, it might be useful to modify this new implementation so that it attempts to determine the local host name dynamically, which would then be a slightly more robust way of testing the "cached" version returned by comdb2_host().

mponomar: I don't think we expect host names to change live.  That'll wreak havoc with how the database reports itself to other replicants, for one.  No problem with having yet another place to fetch the current name (other than now being 2 places), but I was just pointing out that we already have this information.

mistachkin: Good point.  I've modified the 'host' semantics so it's live (i.e. so that we can actually verify that the host name is not different from what we think it should be), e.g. via something like SELECT comdb2_sysinfo('host') == comdb2_host().

This query can now be used to compare the live host name versus what the database thinks it should be.

PR #1531, opened 2019-04-26
Clean exit only when the bdb env is ready

riverszhang89:
A replicant may receive a SIGTERM (from fixcdb2) while catching up with the master, in which case the SIGTERM handler should not perform clean exit as the bdb env is not ready yet.

(DRQS 141558232)

### DIFF
--- a/db/process_message.c
+++ b/db/process_message.c
@@ -605,6 +605,9 @@ void bdb_newsi_mempool_stat();
 static pthread_mutex_t exiting_lock = PTHREAD_MUTEX_INITIALIZER;
 void *clean_exit_thd(void *unused)
 {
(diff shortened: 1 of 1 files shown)


db/process_message.c
```diff
@@ -605,6 +605,9 @@ void bdb_newsi_mempool_stat();
 static pthread_mutex_t exiting_lock = PTHREAD_MUTEX_INITIALIZER;
 void *clean_exit_thd(void *unused)
 {
+    if (thedb->bdb_env == NULL)
```
adizaimi: check gbl_ready instead/also?

riverszhang89: Thanks for the suggestion. Checking gbl_ready is indeed better here.

PR #1533, opened 2019-04-26
Support for VIEWs

nirbhayc:
TODO:
* User schema support (TBD)


db/sql.h
```diff
@@ -102,6 +102,7 @@ struct sqlthdstate {
     int dbopen_gen;
     int analyze_gen;
     int views_gen;
+    int user_views_gen;
```
akshatsikarwar: Lets reuse `dbopen_gen` for this.

sqlite/src/build.c
```diff
@@ -2601,6 +2606,12 @@ void sqlite3EndTable(
       zStmt = sqlite3MPrintf(db, 
           "CREATE %s %.*s", zType2, n, pParse->sNameToken.z
       );
+#if defined(SQLITE_BUILDING_FOR_COMDB2)
+      if (p->pSelect && db->isTimepartView == 0 && iDb != 1) {
+        comdb2_create_view(pParse, pParse->sNameToken.z, pParse->sNameToken.n,
+                           zStmt, 0);
+      }
+#endif
```
akshatsikarwar: `/* defined(SQLITE_BUILDING_FOR_COMDB2) */`

sqlite/src/parse.y
```diff
@@ -1566,8 +1575,11 @@ uniqueflag(A) ::= .        {A = OE_None;}
   ){
     ExprList *p = sqlite3ExprListAppend(pParse, pPrior, 0);
 #if defined(SQLITE_BUILDING_FOR_COMDB2)
-    /* Allow sort order in FK index columns. */
-    if( hasCollate && pParse->db->init.busy==0)
+    /* Allow sort order in FK index columns. Views do not allow sort order.
```
akshatsikarwar: Check why we need this diff from upstream.

nirbhayc: Upstream sqlite touches this code while parsing column lists in FK, CTE & VEIW definitions, where it disallows sort-order altogether.

In Comdb2, unlike upstream, we allow sort orders in FK index column lists, but not in VIEWs and CTE - thus the additional check. (still need to fix this for CTE tough, TODO).

nirbhayc: @akshatsikarwar : As discussed, I have remove the support for use of sort order in DDL FK defs.

bdb/llmeta.c
```diff
@@ -9793,3 +9799,109 @@ int bdb_delete_sc_start_lsn(tran_type *tran, const char *table, int *bdberr)
...
+
+    k = htonl(LLMETA_VIEW);
+    rc = kv_get_keys(t, &k, sizeof(k), (void ***)&v, &n, &bdberr);
+    if (rc || (n == 0)) {
+        *num = 0;
+        return rc;
+    }
+
+    for (int i = 0; i < n; ++i) {
+        names[i] = strdup(v[i]->key.view_name);
```
akshatsikarwar: I don't think you need the strdup here..

bdb/llmeta.c
```diff
@@ -9793,3 +9799,109 @@ int bdb_delete_sc_start_lsn(tran_type *tran, const char *table, int *bdberr)
     }
     return rc;
 }
+
+/* View key */
+struct llmeta_view_key {
+    int file_type;
+    char view_name[LLMETA_TBLLEN];
+    int view_name_len;
```
akshatsikarwar: Remove length from key.

bdb/llmeta.c
```diff
@@ -9793,3 +9799,109 @@ int bdb_delete_sc_start_lsn(tran_type *tran, const char *table, int *bdberr)
...
+    /* Type */
+    u.key.file_type = htonl(LLMETA_VIEW);
+    /* View name */
+    strncpy0(u.key.view_name, view_name, sizeof(u.key.view_name));
+
+    rc = kv_get(t, &u, sizeof(u), (void ***)&view_defs, &num, &bdberr);
+    if (rc == 0) {
+        if (num == 1) {
+            *view_def = view_defs[0];
+        } else { // logical error: there can't be more that one view definition
```
akshatsikarwar: a test with common prefix name to exercise this code without len in key.

nirbhayc:

PR #1535, opened 2019-04-27
Fix potential temptable / object pool deadlock, add tests and diagnostics

mistachkin:


bbinc/object_pool.h
```diff
@@ -335,6 +323,23 @@ int comdb2_objpool_borrow(comdb2_objpool_t op, void **objp);
...
+** op   - an object pool
+** objp - address to be written upon success
+**
+** Return Value
+** 0      - success
+** EPERM  - the pool is stopped
+** EEXIST - object returned by creation function already exists
+** Other  - failed to lock/unlock mutex
+*/
+int comdb2_objpool_available_borrow(comdb2_objpool_t op, void **objp);
```
riverszhang89: force_borrow does exactly this..

bdb/temptable.c
```diff
@@ -79,6 +79,17 @@ struct hashobj {
     unsigned char data[/*len*/];
 };
 
+int gbl_temptable_count;
```
riverszhang89: object pool already keeps track of this.

db/sqlglue.c
```diff
@@ -791,6 +794,8 @@ void done_sql_thread(void)
         destroy_sqlite_master(thd->rootpages, thd->rootpage_nentries);
         free(thd);
     }
+
+    bdb_temp_table_maybe_reset_priority_thread(thedb->bdb_env);
```
riverszhang89: You should do this IFF the thread has requested temptables. For queries that don't need a temp table at all (eg, SELECT * FROM TABLE), this is needless. This is why this is done in bdb_temp_table_close().

bdb/temptable.c
mistachkin: Of course, that only works when the object pool is enabled.  This count works when it is disabled.

util/object_pool.c
```diff
@@ -355,16 +375,9 @@ int comdb2_objpool_return(comdb2_objpool_t op, void *obj)
...
 {
-    int avail;
-
-    Pthread_mutex_lock(&op->data_mutex);
-
-    avail = !exhausted(op) || !full(op);
-
-    Pthread_mutex_unlock(&op->data_mutex);
-    return avail;
+    return objpool_notify_int(op, force);
```
riverszhang89: If comdb2_objpool_notify is merely a call to objpool_notify_int, let's get rid of it.

util/object_pool.c
```diff
@@ -655,6 +671,29 @@ static void *eviction_thread(void *arg)
...
+    return 0;
+}
+
+static int objpool_notify_int(comdb2_objpool_t op, int force)
+{
+    int rc;
+    assert(op != NULL);
+    OP_DBG(op, "notify pool");
+    Pthread_mutex_lock(&op->data_mutex);
+    if (force || (op->nborrowwaits != 0)) {
```
riverszhang89: Only signal when there's waiters (ie, op->nborrowwaits != 0).

util/object_pool.c
```diff
@@ -260,6 +274,12 @@ int comdb2_objpool_resume(comdb2_objpool_t op)
     return 0;
 }
 
+int comdb2_objpool_clear(comdb2_objpool_t op)
+{
+    objpool_evict_all_int(op);
```
riverszhang89: You need to hold evict_mutex here, and then call op->evict_impl(). You don't need to write your own objpool_evict_all_int(). They're already implemented (different pool types have different eviction routines).

util/object_pool.c
```diff
@@ -1063,6 +1128,40 @@ static void objpool_rand_clear(comdb2_objpool_t op)
 }
 // ^^^^^^get/put impl
 
+static void objpool_evict_all_int(comdb2_objpool_t op)
```
riverszhang89: Don't need this.

bdb/temptable.c
```diff
@@ -620,6 +666,17 @@ int bdb_temp_table_create_pool_wrapper(void **tblp, void *bdb_state_arg)
     return bdberr;
 }
 
+int bdb_temp_table_notify_pool_wrapper(void **tblp, void *bdb_state_arg)
+{
+    bdb_state_type *bdb_state = (bdb_state_type *)bdb_state_arg;
+    int rc1 = recover_deadlock_simple(bdb_state);
```
riverszhang89: We definitely do not want to unconditionally recover deadlock every time we request a temp table - it is rare that we run out of temptables, hence most of the time we'd be doing unnecessary work here. Maybe only do this when the pool is both full (# items reaches upper limit) and exhausted (no idle items) -- ie when comdb2_objpool_available() is false.

riverszhang89: Joe pointed out that we get in here exactly when the pool is both full and exhausted and the caller chooses to wait. So this is sane.

mponomar: Don't call this if you're the priority thread - no point in releasing.

util/object_pool.c
mistachkin: I assume you mean get rid of objpool_notify_int(), since the public function is called from another subsystem.

util/object_pool.c
mistachkin: Yeah, I agree with that.  I was just originally nervous about relying on nborrowwaits...  It seems there wouldn't be real harm either way?

util/object_pool.c
mistachkin: Perhaps the objpool_evict_all_int() function is misnamed.  Really, it should be called something like objpool_clear()...  The intent is that it always completely empties out the pool of all objects that are NOT in use.  It seems the evict_impl() functions (potentially) have some additional criteria?

util/object_pool.c
mistachkin: See comment above.  If we truly don't need it, that's cool.  I just need a way to make sure that all objects in the pool are emptied out (i.e. if they aren't in use).

util/object_pool.c
```diff
@@ -355,16 +374,19 @@ int comdb2_objpool_return(comdb2_objpool_t op, void *obj)
...
-
+    int rc;
+    assert(op != NULL);
+    OP_DBG(op, "notify pool");
     Pthread_mutex_lock(&op->data_mutex);
-
-    avail = !exhausted(op) || !full(op);
-
+    if (force || (op->nborrowwaits != 0)) {
+        rc = objpool_signal_unexhausted_int(op);
```
markhannum: broadcast rather than signal ..

util/object_pool.c
riverszhang89: Gotcha. I misunderstood this. This is cool.

util/object_pool.c
```diff
@@ -1063,6 +1122,40 @@ static void objpool_rand_clear(comdb2_objpool_t op)
...
+        free(rec);
+        if (op->del_fn != NULL)
+            op->del_fn(object, op->del_arg);
+        --op->nobjs;
+
+        logmsg(LOGMSG_INFO, "destroyed a pool %s object %p\n", op->name, object);
+        OP_DBG(op, "object evicted");
+    }
+
+    if (indx != 0) {
```
markhannum: ?

mistachkin: Apparently, this code is correct.

util/object_pool.c
```diff
@@ -1063,6 +1121,40 @@ static void objpool_rand_clear(comdb2_objpool_t op)
...
+        memcpy(op->objs, op->objs + indx, sizeof(void *) * (op->in - indx));
```
riverszhang89: memmove

mistachkin: Fixed.

PR #1537, opened 2019-04-29
Add timing accounting: accumulate time spent in a given function

adizaimi:
Add timing accounting: accumulate time spent in a given function. 
Make it simple to get timing information how much time was spent in a given function.
This is enabled only for debug mode.

### DIFF
--- a/db/debug.c
+++ b/db/debug.c
@@ -263,7 +263,8 @@ void debug_trap(char *line, int lline)
         logmsg(LOGMSG_USER, "getvers table        - get schema version for table (or all)\n");
         logmsg(LOGMSG_USER, "putvers table num    - set schema version for table\n");
         logmsg(LOGMSG_USER, "delsc   table tag    - delete a tag\n");
-        logmsg(LOGMSG_USER, "timings              - print all accumulated timing measurements \n");
+        logmsg(LOGMSG_USER, "timings              - print all accumulated "
+                            "timing measurements \n");
     } else {
         logmsg(LOGMSG_ERROR, "Unknown debug command <%.*s>\n", ltok, tok);
     }
--- a/db/debug.c
+++ b/db/debug.c
@@ -21,14 +21,15 @@
 
 #include <segstr.h>
 #include <stdarg.h>
+#include <compat.h>
 
 #include "comdb2.h"
 #include "tag.h"
 #include "net.h"
 #include "nodemap.h"
 #include "logmsg.h"
-#include <intern_strings.h>
-#include <compat.h>
+#include "time_accounting.h"
+#include "intern_strings.h"
 
 int osql_disable_net_test(void);
 int osql_enable_net_test(int testnum);
@@ -79,7 +80,7 @@ static void nettest_usage(void)
 
 char *tcmtest_routecpu_down_node;
 
-/* parse debug trap */
+/* parse debug trap @send debug <cmd> */
 void debug_trap(char *line, int lline)
 {
     char table[MAXTABLELEN];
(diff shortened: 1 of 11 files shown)


util/time_accounting.h
```diff
@@ -0,0 +1,50 @@
...
+#define _time_accounting_h
+
+
+#ifndef NDEBUG
+enum { CHR_IXADDK, CHR_DATADD, CHR_TMPSVOP, CHR_MAX };
+
+#define CHRONO_START()   \
+{                        \
+    struct timeval __tv; \
+    gettimeofday(&__tv, NULL);
```
riverszhang89: gettimeofday isn't cheap. We should at least not do this by default.

adizaimi: It is off by default. On only if DEBUG is on.

riverszhang89: Gotcha. This can be done by just 1 macro though. For example, you'd have a macro like the following:

```c
#define TIMER(code) do { struct timeval __tv; gettimeofday(&__tv, NULL); X; accumulate_time(name, chrono_stop(&__tv)); } while (0)
```

And you'd time a code block like this:

```c
TIMER(
    // line 1
    // line 2
    // line 3
)
```

db/debug.c
```diff
@@ -254,14 +255,16 @@ void debug_trap(char *line, int lline)
         logmsg(LOGMSG_USER, "nodes:\n");
         for (int i = 0; i < numnodes; i++)
             logmsg(LOGMSG_USER, "  %s %d\n", hosts[i], nodeix(hosts[i]));
+    } else if (tokcmp(tok, ltok, "printchron") == 0) {
```
riverszhang89: The command is a bit confusing as "chron" isn't a real word.

db/debug.c
```diff
@@ -254,14 +255,16 @@ void debug_trap(char *line, int lline)
...
+        print_all_time_accounting();
     } else if (tokcmp(tok, ltok, "help") == 0) {
         logmsg(LOGMSG_USER, "tcmtest <test>       - enable a cdb2tcm test\n");
         logmsg(LOGMSG_USER, "tcmtest list         - list cdb2tcm tests\n");
         logmsg(LOGMSG_USER, "getvers table        - get schema version for table (or all)\n");
         logmsg(LOGMSG_USER, "putvers table num    - set schema version for table\n");
         logmsg(LOGMSG_USER, "delsc   table tag    - delete a tag\n");
+        logmsg(LOGMSG_USER, "printchron           - print all chron timer accumulation \n");
```
riverszhang89: chron timer -> chronometer?

adizaimi:
Thanks Rivers. I incorporated the changes you suggested. I think the code looks a bit strange with code inside ACCUMULATE_TIMING(); but I don't mind it too much.

riverszhang89:
Thanks. It is cleaner and forms its own code block. I think it is better than having 2 incomplete macros (opening brace in one and closing brace in the other).

util/time_accounting.h
```diff
@@ -0,0 +1,51 @@
...
+
+void print_time_accounting(int el);
+void print_all_time_accounting();
+
+void reset_time_accounting(int el);
+void reset_all_time_accounting();
+
+#else
+#define ACCUMULATE_TIMING(NAME, CODE) do { CODE; } while(0);
+#define print_all_time_accounting() {}
```
riverszhang89: Since you already use the `do {} while (0)` idiom for ACCUMULATE_TIMING you might also do the same to this macro.

util/time_accounting.h
```diff
@@ -0,0 +1,51 @@
...
+#define cleanup_time_accounting() {}
```
riverszhang89: Where is this defined otherwise?

util/time_accounting.h
```diff
@@ -0,0 +1,51 @@
...
+    accumulate_time(NAME, __sec_part + __usec_part); \
+} while(0); 
+
+void accumulate_time(int el, int us);
+
+void print_time_accounting(int el);
+void print_all_time_accounting();
+
+void reset_time_accounting(int el);
+void reset_all_time_accounting();
```
riverszhang89: No definition?

util/time_accounting.h
```diff
@@ -0,0 +1,51 @@
...
+    int __usec_part = (__tv2.tv_usec - __tv1.tv_usec); \
+    accumulate_time(NAME, __sec_part + __usec_part); \
+} while(0); 
+
+void accumulate_time(int el, int us);
+
+void print_time_accounting(int el);
+void print_all_time_accounting();
+
+void reset_time_accounting(int el);
```
riverszhang89: Not used.

adizaimi: yes, removing from header.

util/time_accounting.h
adizaimi: yes, maybe useful in future, but for now removing from header.

util/time_accounting.h
adizaimi: ok

PR #1539, opened 2019-04-29
Safely remove an element from a SQLite hashtable while looping over it.

riverszhang89:
Make sure we don't read from freed memory.

(DRQS 141427848)


riverszhang89:
Thank you so much for the review!

PR #1544, opened 2019-05-02
Fix a double free in VDBE Datetime handling

riverszhang89:
Set Mem.szMalloc to 0 to prevent releaseMemArray() from double freeing.

(DRQS 141874636)


tests/datetime.test/t6_01.req
```diff
@@ -0,0 +1 @@
+select (cast(cast(now() as text) as datetime) - cast(cast(now() as text) as datetime)) as intv
```
nirbhayc: In this case, would sqlite compute NOW() only once? If not, the result could be non-zero.

nirbhayc: Actually it does. It stores NOW() in a register and reuses it.

riverszhang89: Yep. The bug seems to be hit only when we cast `now()` to text first and cast the text back to datetime again. I use the expression (essentially `now() = now()`) so I can always get the same output.

PR #1547, opened 2019-05-06
net: Fix self-deadlock on a failed write_connect_message.

riverszhang89:
We already hold `host_node_ptr->lock`. Since `host_node_ptr->lock` isn't recursive, relocking it would hang.

(DRQS 141973376)


riverszhang89:
Thank you Akshat. Thank you Joe. I will merge the pull request once it is passing the tests.

PR #1548, opened 2019-05-07
Fixes for SQL identifier and string literal quoting issues

mistachkin:
These changes attempt to address some issues with the quoting of SQL identifier names and string literal values embedded within dynamically built SQL query strings.


db/fdb_fend.c
```diff
@@ -964,30 +964,30 @@ static int check_table_fdb(fdb_t *fdb, fdb_tbl_t *tbl, int initial,
...
                      "select * from sqlite_master"
-                     " where tbl_name='%s' or tbl_name='sqlite_stat1' or "
+                     " where tbl_name='%q' or tbl_name='sqlite_stat1' or "
                      "tbl_name='sqlite_stat4' collate nocase",
                      tbl->name);
         } else {
-            snprintf(sql, sizeof(sql),
+            sql = sqlite3_mprintf(
                      "select * from sqlite_master"
-                     " where tbl_name='%s' collate nocase",
```
mponomar: We end up leaking sql if we hit one of the `goto run` cases below (eg: ending up on a node with a lower FDB version)

db/sqlanalyze.c
```diff
@@ -808,30 +830,32 @@ static int analyze_table_int(table_descriptor_t *td,
         sampled_table = 1;
         rc = sample_indicies(td, &clnt, tbl, td->scale, td->sb);
         if (rc) {
-            snprintf(sql, sizeof(sql), "Sampling table '%s'", td->table);
+            snprintf(zErrTab, sizeof(zErrTab), "Sampling table '%s'", td->table);
```
mponomar: Joe, can you please double-chec? I don't think zErrTab is populated in all reachable cases.

sqlite/src/build.c
```diff
@@ -5511,21 +5519,29 @@ char *sqlite3DescribeIndexOrder(
      }
      isExpr = 1;
   }
-  ret = sqlite3_mprintf("(%s)%s",
-    isExpr?pExprDesc:pTbl->aCol[pIdx->aiColumn[0]].zName, pDesc);
-
+  if( isExpr ){
+    ret = sqlite3_mprintf("(%s) %s", pExprDesc, pDesc);
```
mponomar: Not sure - Dorin can comment - are pExprDesc already correctly quoted? (in this routine and other similar uses)

sqlite/src/expr.c
```diff
@@ -5550,7 +5550,7 @@ static char *print_mem(Mem *m){
       }
       key[2*m->n] = '\0';
 
-      return sqlite3_mprintf("x'%s'", key);
+      return sqlite3_mprintf("x'%q'", key);
```
mponomar: I don't think we can have quote characters inside a hex literal (not that this case is any worse than without the change)

PR #1552, opened 2019-05-08
Fix some warnings

akshatsikarwar:


akshatsikarwar:
`cdb2test   May  8 17:40:15 2019  in-comdb2db cleanup.R20190508.5`

datetime/localtimedb.c
```diff
@@ -2207,13 +2208,9 @@ static struct db_state *find_tz(const char *name)
...
 {
-    char key[NAME_KEY_MAX];
-    tz_hash_entry_type *hash_entry_ptr;
-
-    bzero(key, NAME_KEY_MAX);
-    strcpy(key, name);
-
-    hash_entry_ptr = malloc(sizeof(tz_hash_entry_type));
+    char key[NAME_KEY_MAX] = {0};
+    strncpy0(key, name, sizeof(key));
```
riverszhang89: I don't think we need `key` here. We could copy `name` directly into `hash_entry_ptr->key`.

akshatsikarwar: Good point.

PR #1559, opened 2019-05-09
More temptable tests

mistachkin:
These new tests are designed to help isolate issues with the temptable subsystem.  In particular, to show that it can end up in a state where no SQL threads can proceed (i.e. "deadlock").


mistachkin:
This tests added by this pull request are already part of #1535.  This branch was created to make it easy to test against current master.

PR #1560, opened 2019-05-09
Fix a race between clean-exit and net queue stat

riverszhang89:
clean-exit might have destroyed the net before we get in `net_enque_write_rtn()`. Protect us from the race by checking the exiting flag in the net first.

(DRQS 142229812)


riverszhang89:
Thank you Joe!

PR #1562, opened 2019-05-09
Fix temptable segfault when a trigger attempts to open a temporary btree via SQL

mistachkin:
These changes fix a crash that can occur when a trigger attempts to open a temporary btree via SQL.


mistachkin:
I've now added a new test case to this PR in order address this gap in our test coverage.

akshatsikarwar:
These should be part of reset_clnt and cleanup_clnt instead.

PR #1563, opened 2019-05-10
Empty blobs shouldn't be bound like NULL

nirbhayc:
DRQS 142258096


akshatsikarwar:
`fastsql` needs the same?

nirbhayc:
It very well could. I'm having a look.

PR #1564, opened 2019-05-10
Cosmetic changes to fix warning from code analysis.

riverszhang89:
Use of freed values, undefined externs, etc.

### DIFF
--- a/bdb/bdb_verify.c
+++ b/bdb/bdb_verify.c
@@ -815,7 +815,7 @@ static int bdb_verify_ll(
 
                         if (rc == 0) {
                             realblobsz[blobno] = dbt_blob_data.size;
-                            if (blobsizes[blobno] == -1 && rc != DB_NOTFOUND) {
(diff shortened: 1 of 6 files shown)


bdb/bdb_verify.c
```diff
@@ -815,7 +815,7 @@ static int bdb_verify_ll(
 
                         if (rc == 0) {
                             realblobsz[blobno] = dbt_blob_data.size;
-                            if (blobsizes[blobno] == -1 && rc != DB_NOTFOUND) {
```
nirbhayc: What was the warning here?

riverszhang89: `rc != DB_NOTFOUND` will always to true as `if (rc == 0)` guarantees it.

nirbhayc: right

riverszhang89:
Thank you for the review Nirbhay!

PR #1566, opened 2019-05-10
handle column-less unions of tables

dorinhogea:
Fix queries of type:
select 1 from (select * from t1 union select * from t2) where col="foo"


db/dohast.c
```diff
@@ -161,6 +161,10 @@ char *sqlite_struct_to_string(Vdbe *v, Select *p, Expr *extraRows,
         sqlite3_free(where);
         return NULL;
     }
+    if (!tbl && p->pSrc->nSrc) {
+        /* select 1 from tbl */
+        tbl = (const char*)p->pSrc->a[0].zName;
+    }
 
     if (unlikely(!tbl)) {
```
mponomar: Is this case still hit?  I guess it would be for cases like `select 1` but those won't be candidates for parallelization?

dorinhogea: Yes, the case is still hit, for example "select a from t1 union all select 1"

PR #1567, opened 2019-05-11
Sockpool mutex r7

markhannum:
Port cdb2api sockpool_fd pool to r7.


markhannum:
The cdb2api failure was fixed under https://github.com/bloomberg/comdb2/pull/1572

PR #1570, opened 2019-05-13
Simplify temp table mgmt

akshatsikarwar:
Remove unnecessary complexity caused by sharing tmptable hash  between SQL threads.

### DIFF
--- a/db/sql.h
+++ b/db/sql.h
@@ -572,10 +572,6 @@ struct sqlclntstate {
     int want_stored_procedure_debug;
     char spname[MAX_SPNAME + 1];
     struct spversion_t spversion;
-    int n_lua_stmt;
-    int max_lua_stmt;
-    pthread_mutex_t *temp_table_mtx; /* for "sp.c" temp table subsystem */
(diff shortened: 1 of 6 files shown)


db/sql.h
```diff
@@ -752,6 +748,15 @@ struct query_path_component {
     LINKC_T(struct query_path_component) lnk;
 };
 
+struct temptable {
+    int rootpage;
+    struct temp_cursor *cursor;
+    struct temp_table *tbl;
```
nirbhayc: IMO, `struct temptable` vs `struct temp_table` is a bit confusing.

akshatsikarwar: Agreed. I can cleanup in subsequent PR (struct wasn't introduced in the PR, that is what it was called already).

akshatsikarwar: I remember when both the structs were called `struct temp_table`. One day I couldn't take it anymore and renamed the one in `sqlglue.c`!

nirbhayc: https://news.ycombinator.com/item?id=14598201

akshatsikarwar: Indeed!

PR #1576, opened 2019-05-15
Free normalized sql/locks allocated by dohast threads

mponomar:
Fixes memory leak for parallelized queries.

### DIFF
--- a/db/dohsql.c
+++ b/db/dohsql.c
@@ -1266,6 +1266,13 @@ int dohsql_end_distribute(struct sqlclntstate *clnt, struct reqlogger *logger)
                 conns->stats.max_queue_bytes =
                     conns->conns[i].stats.max_queue_bytes;
         }
+        if (conns->conns[i].clnt) {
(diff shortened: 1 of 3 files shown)


db/dohsql.c
```diff
@@ -1266,6 +1266,13 @@ int dohsql_end_distribute(struct sqlclntstate *clnt, struct reqlogger *logger)
                 conns->stats.max_queue_bytes =
                     conns->conns[i].stats.max_queue_bytes;
         }
+        if (conns->conns[i].clnt) {
```
dorinhogea: Please make this part of _shard_disconnect, which is responsible for cleanup up sql. Even better, can we make this part of cleanup_clnt routine?   This should not be specific to dohsql.

mponomar: It's only specific to dohsql because it's the only code path where we have a clnt that doesn't get cleaned up by normal means.  But it's still a good point to stick this into cleanup_clnt.

mponomar:
Moved to cleanup_clnt

PR #1577, opened 2019-05-15
fix count(*) per sqlite Check-in [05897ca4]

dorinhogea:
Simple fix for crash in:
"select count(*) from (select * from t union all select * from t) where col=1"


dorinhogea:
Thank you Mike

PR #1582, opened 2019-05-20
Fix new abort on sqlite mutex

dorinhogea:
Currently we require sqlite_mutex to be held when calling alloc and free in sqlite (not needed though).   Fix one more aborting path.


dorinhogea:
thank you

PR #1583, opened 2019-05-21
Create berk tmptable when spilling a hash temp tbl

adizaimi:
Fix bug: create berk tmptable when spilling a hash temp tbl

### DIFF
--- a/bdb/temptable.c
+++ b/bdb/temptable.c
@@ -389,6 +389,11 @@ static int bdb_hash_table_copy_to_temp_db(bdb_state_type *bdb_state,
     unsigned int hash_cur_buk;
     char *data;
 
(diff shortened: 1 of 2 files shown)


adizaimi:
Thanks Akshat.

bdb/temptable.c
```diff
@@ -389,6 +389,11 @@ static int bdb_hash_table_copy_to_temp_db(bdb_state_type *bdb_state,
     unsigned int hash_cur_buk;
     char *data;
 
+    if (tbl->dbenv_temp == NULL &&
+        create_temp_db_env(bdb_state, tbl, bdberr) != 0) {
```
nirbhayc: A comment here would be better.

PR #1584, opened 2019-05-21
Fix rename with legacy config

dorinhogea:
Legacy settings use a different a schema change call stack, and rename was not working in this case.  Here is the patch.

### DIFF
--- a/bdb/bdb_schemachange.c
+++ b/bdb/bdb_schemachange.c
@@ -222,7 +222,7 @@ retry:
 }
 
 static int do_llog(bdb_state_type *bdb_state, scdone_t sctype, char *tbl,
-                   int wait, int *bdberr)
+                   int wait, const char *origtable, int *bdberr)
 {
     DBT *dtbl = NULL;
     if (tbl) {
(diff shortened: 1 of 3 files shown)


bdb/bdb_schemachange.c
```diff
@@ -222,14 +222,25 @@ static int do_llog_int(bdb_state_type *ch_bdb_state, DBT *tbl, DBT *type,
...
-                   int wait, int *bdberr)
+                   int wait, const char *origtable, int *bdberr)
 {
     DBT *dtbl = NULL;
     if (tbl) {
         dtbl = alloca(sizeof(DBT));
         bzero(dtbl, sizeof(DBT));
         dtbl->data = tbl;
         dtbl->size = strlen(tbl) + 1;
+        dtbl->size = strlen(bdb_state->name) + 1;
```
nirbhayc: Do we need this? Also, its overwriting the above assignment.

dorinhogea: typo, removed

PR #1345, opened 2018-12-27
Reorder index add/upd/del defered

adizaimi:
Reorder index operations (add/upd/del) by table and index value and perform them deferred at the end of the osql processing. This ensures that we avoid deadlocking by touching (locking) pages in an increasing order from all the transactions on master node.

PR #1587, opened 2019-05-23
TEST ONLY: test with delayed keyadds always on

adizaimi:
test with delayed keyadds always on -- need to test behavior of sysbench

### DIFF
--- a/bdb/bdb_fetch.h
+++ b/bdb/bdb_fetch.h
@@ -67,8 +67,9 @@
 
 typedef struct {
     uint8_t ver;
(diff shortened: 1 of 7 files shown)


adizaimi:
Seems that oltp_read_write is not affected by delaying the key adds, nor by getting writelock a bit earlier for index verification (updrec/delrec). This test shows that https://github.com/bloomberg/comdb2/pull/1345 oltp_read_write  has issues with oltp_read_write for other reasons rather than the two above.

PR #1589, opened 2019-05-28
indices.c: ix find to get write lock in updrec and delrec

adizaimi:
In upd_rec_indices and del_rec_indices we call ix find in readl lock first to verify that records exists, then subsequently get write lock to update the record. This is the cause for a deadlocks and instead, by getting the record in write lock from the beginning, we turn the deadlocks into waits, thus improving performance by 20% in some sysbench tests, but more importantly by not busying the cpu undoing work in the case of a deadlock (which aborts one of the transactions).
This was part of https://github.com/bloomberg/comdb2/pull/1345 but checking it in separately because it is relevant on its own.

### DIFF
--- a/bdb/bdb_fetch.h
+++ b/bdb/bdb_fetch.h
@@ -67,8 +67,9 @@
 
 typedef struct {
     uint8_t ver;
-    int ignore_incoherent;
-    int page_order;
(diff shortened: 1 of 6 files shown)


markhannum:
Yes awesome .. okay to merge

PR #1592, opened 2019-05-29
 Remove thedb->exiting because it is never set 

adizaimi:
Remove thedb->exiting because it is never set 
Call db_is_stopped() accessor instead of checking thedb->stopped directly


adizaimi:
Thank you Mike. 
gbl_exit is used for exiting immediately without bringing up much of the server, say when doing --create.

PR #1595, opened 2019-05-29
Don't clear gbl_cron in views.

mponomar:
Partition system clears all scheduled events - it should ignore events scheduled on gbl_cron that it didn't schedule.


adizaimi:
Cool, Dorin only restarts timepartition crons in https://github.com/bloomberg/comdb2/pull/1596.

PR #1596, opened 2019-05-30
only clear logical and timepart schedulers

dorinhogea:
currently code clears a few other schedulers (gbl_sched, memdump) on master swing, which don't restart.  the master swing should only clear logical and time-partition schedulers, as here.


adizaimi:
manual_partition test seems a expected output needs fixing.
odh_blob seems to randomly fail for master as well.
I dont know about tablelocks test.

PR #1597, opened 2019-05-30
fix: sp error string should not be null for fastsql codepath

adizaimi:
Fastsql codepath crashes when lua sp err str is null. 
Make sure we set it in a few codepaths that were previously missed.
Also set it to generic error in case there are other paths that don't set it.

### DIFF
--- a/lua/sp.c
+++ b/lua/sp.c
@@ -6207,7 +6207,8 @@ static int exec_procedure_int(struct sqlthdstate *thd,
     if ((rc = commit_sp(L, err)) != 0) return rc;
 
     if (sprc) {
-        *err = strdup("emit_result error");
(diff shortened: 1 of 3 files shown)


lua/sp.c
```diff
@@ -6200,7 +6206,10 @@ static int exec_procedure_int(struct sqlthdstate *thd,
 
     if ((rc = commit_sp(L, err)) != 0) return rc;
 
-    if (sprc) return sprc;
+    if (sprc) {
+        *err = strdup("emit_result error");
```
mohitkhullar: if (*err == NULL)?

adizaimi: You are right, fixing. thanks

PR #1599, opened 2019-05-30
Clean exit: wait for all the threads to terminate

adizaimi:
Use thread manager to account for all running threads in order to exit cleanly.
Upon exiting, the db will be left with only the main thread which waits for all the other threads to exit first.
(In case of a thread failing to exit, main will be stuck and timeout will eventually kill the DB.)
This clean exit will allow us to detect memory leaks by monitoring memory usage on process exit.


bdb/attr.h
```diff
@@ -357,7 +357,7 @@ DEF_ATTR_2(
 DEF_ATTR(PAGEDEADLOCK_MAXPOLL, pagedeadlock_maxpoll, QUANTITY, 5,
          "If retrying on deadlock (see pagedeadlock_retries), poll up to this "
          "many ms on each retry.")
-DEF_ATTR(ENABLE_TEMPTABLE_CLEAN_EXIT, temp_table_clean_exit, BOOLEAN, 0,
+DEF_ATTR(ENABLE_TEMPTABLE_CLEAN_EXIT, temp_table_clean_exit, BOOLEAN, 1,
```
riverszhang89: This seems unnecessary ...

db/comdb2.h
```diff
@@ -3614,4 +3613,6 @@ extern int gbl_disable_tpsc_tblvers;
 extern int gbl_osql_odh_blob;
 extern int gbl_pbkdf2_iterations;
 extern int gbl_bpfunc_auth_gen;
+extern int gbl_thread_count;
```
riverszhang89: We already have code to wait for threads to exit on clean-exit. https://github.com/bloomberg/comdb2/blob/master/db/thrman.c#L535

Why don't we just register these threads to the thread manager?

adizaimi: Only some of the threads of our system go through thrman.c. 
Most of the ones that linger at the end are simply pthread_created so unfortunately I have to account for all of them myself. The upside is that if we structure our threads which loop to check for exit with a period of 1 second, then counting all of them is not necessary (ie counting a subset suffices because they exit at the same time because of the 1 per second check).

bdb/attr.h
adizaimi: It is only necessary to cleanup memory on exit. Not necessary otherwise.

riverszhang89: Why is it necessary to clean up temptables on clean exit?

db/comdb2.h
riverszhang89: I know. That's why I'm suggesting that you register these threads to the thread manager so it will take care of the wait-for-completion logic for you.

bdb/attr.h
adizaimi: The goal is to have clean memory as reported by valgrind. I suppose we could only do so on DEBUG build, instead of Release build. Not sure it is worth spending time to differentiate tho.

db/comdb2.h
adizaimi: I didn't understand you earlier, I can look into it I suppose.

adizaimi:
All tests are exiting cleanly with only one thread showing up on gstack:
```
pno-comdb2btda1:/bb/comdb2a/roborivers/tests/2268535d$ find | grep onexit | xargs grep "Thread 1 " | wc
    818
pno-comdb2btda1:/bb/comdb2a/roborivers/tests/2268535d$ find | grep onexit | xargs grep "Thread 2 " | wc
      0
```

adizaimi:
Noticed this earlier in test /bb/comdb2a/roborivers/tests/460c2d99> less common/c1/db/logs/ddlnocascadedrop5547.*db

2019/07/30 18:23:11 heartbeat check thread exiting!
2019/07/30 18:23:11 7f3d2fbe4700 [offloadsql m2.54cb94b2_primary fd -1] connect_thread: netinfo->exiting
2019/07/30 18:23:11 [ERROR] WARNING: REMOVED WRONG ITEM 0x7f3d3a3a4528 FROM LIST 0x2196a38
Connection to m1.54cb94b2_primary closed.

adizaimi:
Closing this and opening a new clean PR to continue working on this: https://github.com/bloomberg/comdb2/pull/2775

PR #1600, opened 2019-05-30
berkdb: Fix a memory leak in the rep processor

riverszhang89:
`pglogs` is malloc'd in `bdb_pglogs_key_list_init()` and hence needs freed.

(DRQS 143278402)


adizaimi:
Looks good. I wonder we coerce this leak to happen in your memtest to catch the leak?

PR #1602, opened 2019-05-31
Fix mangled port values with older comdb2

dorinhogea:
R7 mangles port numbers with net id, example:
netinfo_ptr->myport != connect_message.to_portnum 19856 85392
85392 is 19856 | (1<<16), since it comes from a secondary net. 
We need to make sure we remove net id while processing hello messages, otherwise we might populate netinfo with wrong port values (like, when older comdb2 passes back hello messages that have ports mangled like that).


dorinhogea:
haha, thanks!

PR #1607, opened 2019-06-05
Selectv writelock

markhannum:
Grab a writelock for a selectv'd record that will later be modified.

### DIFF
--- a/db/db_tunables.c
+++ b/db/db_tunables.c
@@ -290,6 +290,7 @@ extern int gbl_osql_check_replicant_numops;
 extern int gbl_abort_on_missing_osql_session;
 extern int gbl_abort_irregular_set_durable_lsn;
 extern int gbl_legacy_schema;
+extern int gbl_selectv_writelock_on_update;
 
 int gbl_debug_tmptbl_corrupt_mem;
 
--- a/db/db_tunables.c
+++ b/db/db_tunables.c
@@ -291,6 +291,7 @@ extern int gbl_abort_on_missing_osql_session;
 extern int gbl_abort_irregular_set_durable_lsn;
 extern int gbl_legacy_schema;
 extern int gbl_selectv_writelock_on_update;
+extern int gbl_selectv_writelock;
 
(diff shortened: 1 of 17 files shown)


db/db_tunables.c
```diff
@@ -290,6 +290,8 @@ extern int gbl_osql_check_replicant_numops;
 extern int gbl_abort_on_missing_osql_session;
 extern int gbl_abort_irregular_set_durable_lsn;
 extern int gbl_legacy_schema;
+extern int gbl_selectv_writelock_on_update;
+extern int gbl_selectv_writelock;
```
adizaimi: This one seems to be a duplicate so can be deleted.

db/osqlcomm.c
```diff
@@ -7487,7 +7503,10 @@ int osql_process_packet(struct ireq *iq, unsigned long long rqid, uuid_t uuid,
 
         lclgenid = bdb_genid_to_host_order(dt.genid);
 
-        rc = ix_check_genid(iq, trans, dt.genid, &bdberr);
+        if (gbl_selectv_writelock)
```
adizaimi: Using wrong variable which is never set. Why does the testcase work?

tests/selectv_deadlock.test/lrl.options
```diff
@@ -0,0 +1,17 @@
...
+maxosqltransfer 1000000
+enable_new_snapshot
+enable_snapshot_isolation
+berkattr abort_on_replicant_log_write 1
+init_with_genid48
+
+#reorder_socksql_no_deadlock on
+#reorder_idx_writes on
+
+#selectv_writelock_on_update is enabled by default
```
adizaimi: Clean up some of this?

db/db_tunables.c
markhannum: No - selectv_writelock acquires a writelock for every selectv, selectv_writelock_on_update only acquires a writelock for rows that are updated later

PR #1608, opened 2019-06-05
Display deadlock steps aborted and worst wait time

adizaimi:
We already show every second in .statreqs file number of deadlocks (and lockwaits) and number of locks gotten in that interval and average time they took. 
Here we add the following data points for that interval:
  - Aggregate number of locks aborted when transactions aborted because of deadlock.
  - Worst wait time for any lock.
```
    ndeadlocks 362, nlockwaits 1203, vreplays 0, locks aborted 10
    399 locks, avg time 37ms, worst time 432ms
```

### DIFF
--- a/bbinc/thread_stats.h
+++ b/bbinc/thread_stats.h
@@ -1,4 +1,17 @@
 /*
+   Copyright 2015 Bloomberg Finance L.P.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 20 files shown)


bbinc/thread_stats.h
```diff
@@ -0,0 +1,35 @@
+/*
+ */
```
nirbhayc: License?

PR #1609, opened 2019-06-05
Fix handling of the keepsocket flag to prevent a possible double-free and/or leaking appsock connections

mistachkin:


dorinhogea:
Why don't you make close_appsock(SBUF2 **sb), and set sb to NULL inside the call? This way we capture all cases, like:
db/socket_interfaces.c:156:    close_appsock(sb);
plugins/newsql/newsql.c:2436:    close_appsock(sb);
schemachange/sc_struct.c:94:        close_appsock(s->sb);

PR #1611, opened 2019-06-06
fix limit when missing parameters force an early error

dorinhogea:
Corner case: parsing succeeds, but bound parameters are missing.  This generates an early error that triggers a double free.


adizaimi:
Can we add a test case for this please?

PR #1617, opened 2019-06-07
Highly experimental changes to dohast / bound parameter handling

mistachkin:


db/sqlinterfaces.c
```diff
@@ -3313,7 +3316,7 @@ struct param_data *clnt_find_param(struct sqlclntstate *clnt, const char *name,
         if (p->pos > 0 && p->pos == index)
             return p;
 
-        if (name[0] && !strncasecmp(name, p->name, strlen(name) + 1))
+        if (name[0] && !strncmp(name, p->name, strlen(name) + 1))
```
mistachkin: parameter names are always treated as case-sensitive by SQLite...  So, I think that handling is applicable here as well.

dorinhogea: Please provide a test for case sensitive parameters (like @a vs @A)

PR #1618, opened 2019-06-10
fix foreign constraint check in time partition code

dorinhogea:
When a time partition has a foreign constraint on a non-partition table, code abort. Fixing this.


nirbhayc:
Could you also add a test?

dorinhogea:
it had a test, I forgot to push it though

PR #1619, opened 2019-06-10
Cleanup and simplify the check_sql() function...

mistachkin:
These changes radically simplify the check_sql() function, moving most of its prior responsibilities (i.e. those not directly related to EXEC PROCEDURE, which does not go through the SQLite parser) into the authorizer.

Also, the EXECUTE keyword is added as an alternative to EXEC.


db/sql.h
```diff
@@ -79,8 +79,13 @@ typedef struct stmt_hash_entry {
 } stmt_hash_entry_type;
 
 struct sql_authorizer_state {
-    int denyDdl;                       /* non-zero if DDL is forbidden */
-    int numDdls;                       /* number of DDL statements found */
+    struct sqlclntstate *clnt;         /* pointer to current client info */
+    int denyDdl;                       /* DDL is forbidden? */
+    int denyPragma;                    /* PRAGMA is forbidden? */
+    int denyTrigger;                   /* CREATE TRIGGER is forbidden? */
```
nirbhayc: Move them into a single 'flag' variable with OR-able flags?

mistachkin: Good point.  I actually thought about doing that; however, I was afraid it would hurt the readability too much.  I suppose the code will be easier to maintain with just the flags versus having individual boolean fields.

mistachkin: I've made the necessary changes.  While making the changes, I discovered an issue with the authorizer integration that I also fixed (i.e. it assumed that "thd" was always non-NULL).

sqlite/src/main.c
```diff
@@ -3433,13 +3433,15 @@ static int openDatabase(
...
 
 #if defined(SQLITE_BUILDING_FOR_COMDB2)
-  static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
-  pthread_mutex_lock(&mutex);
-  /* these modify global structures */
-  register_lua_sfuncs(db, thd);
-  register_lua_afuncs(db, thd);
-  register_date_functions(db); 
-  pthread_mutex_unlock(&mutex);
+  if( thd!=NULL ){
```
mistachkin: Fix for NULL "thd", part 2.  The registered Lua functions assume "thd" cannot be NULL; therefore, they cannot be added when "thd" is in fact NULL.

db/sqlinterfaces.c
```diff
@@ -911,7 +927,9 @@ int sqlite3_open_serial(const char *filename, sqlite3 **ppDb,
                 */
             }
         }
-        comdb2_setup_authorizer_for_sqlite(*ppDb, &thd->authState, 1);
+        if (thd != NULL) {
```
mistachkin: Fix for NULL "thd", part 1.  Setting up the authorizer for the database requires a non-NULL "thd".

db/sql.h
```diff
@@ -79,8 +79,9 @@ typedef struct stmt_hash_entry {
 } stmt_hash_entry_type;
 
 struct sql_authorizer_state {
-    int denyDdl;                       /* non-zero if DDL is forbidden */
-    int numDdls;                       /* number of DDL statements found */
+    struct sqlclntstate *clnt;         /* pointer to current client info */
+    int flags;                         /* DDL, PRAGMA, CREATE TRIGGER denied? */
```
mistachkin: Per code review feedback, use a flags field instead of individual bits here.

mistachkin:
Yes, I should reference EXECUTE in the docs and tests.

PR #1620, opened 2019-06-10
Don't create a new ast on recursive parse.

mponomar:
Fixes a memory leak on delete/update on time partitions.


mponomar:
Better fix in #1631

PR #1625, opened 2019-06-11
UPSERT: Fix a bug in constraint check

nirbhayc:
While checking for duplicate keys, the code erroneously passed the same buffer for IN & OUT key arguments, causing the underlying logic to incorrectly return IX_FND. Due to this, in some specific cases,
INSERT .. ON CONFLICT .. DO NOTHING ended up ignoring a record that could have been inserted otherwise.


nirbhayc:
Thanks @akshatsikarwar!

PR #1626, opened 2019-06-11
cdb2api: Honor per-handle default_type config in comdb2db DNS lookup

riverszhang89:
Prefer the per-handle default_type over the global config when forming comdb2db domain name.


mohitkhullar:
cdb2_default_cluster should be based on machine config.

riverszhang89:
Withdrawing this. Comdb2db destinations should remain a global config.

PR #1631, opened 2019-06-14
Fix recursive parsing dohast

dorinhogea:
There are two types of recursive parsing: nested parsing that reusing the Parse object, and another used currently by triggers and upserts that relies on a tree of Parse objects connected by pToplevel pointers.
Fixing the tree-based recursive parsing is similar with https://github.com/bloomberg/comdb2/pull/1620, but code makes sure the dohsql is disabled for the statement.
Fixing nested parsing (for which I don't have yet a statement) requires moving the ast member from the to-be-zeroed area in Parse object to the recursive/constant part.

### DIFF
--- a/db/ast.h
+++ b/db/ast.h
@@ -30,9 +30,10 @@ enum ast_type {
     AST_TYPE_UPDATE = 6
 };
 
-ast_t *ast_init(void);
 struct Vdbe;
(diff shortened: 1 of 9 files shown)


dorinhogea:
Thank you for review

PR #1636, opened 2019-06-18
Refactor bdb_osql_cache_table_versions

markhannum:
Don't hold the children lock while reading from berkley.


markhannum:
Eventlog test failure addressed in https://github.com/bloomberg/comdb2/pull/1641

PR #1638, opened 2019-06-18
Make clean exit on SIGTERM optional

mponomar:
Make orderly database shutdown on SIGTERM optional.  Tunable toggles behavior.  Legacy options disable by default.

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -768,6 +768,8 @@ int64_t gbl_temptable_spills;
 
 int gbl_osql_odh_blob = 1;
 
+int gbl_clean_exit_on_sigterm = 0;
+
 comdb2_tunables *gbl_tunables; /* All registered tunables */
 int init_gbl_tunables();
 int free_gbl_tunables();
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -768,7 +768,7 @@ int64_t gbl_temptable_spills;
 
 int gbl_osql_odh_blob = 1;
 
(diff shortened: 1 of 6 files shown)


db/db_tunables.c
```diff
@@ -710,6 +712,19 @@ static int deadlock_policy_override_update(void *context, void *value)
...
 
+extern void clean_exit_sigwrap(int signum);
+
+static int update_clean_exit_on_sigterm(void *context, void *value) {
+    int val = *(int *)value;
+    if (val)
+        signal(SIGTERM, clean_exit_sigwrap);
+    else
+        signal(SIGTERM, SIG_DFL);
+    printf("turning SIGTERM handling %s\n", val ? "on" : "off");
```
nirbhayc: logmsg?

db/comdb2.c
```diff
@@ -768,6 +768,8 @@ int64_t gbl_temptable_spills;
 
 int gbl_osql_odh_blob = 1;
 
+int gbl_clean_exit_on_sigterm = 0;
```
nirbhayc: Should be 1 by default.

db/db_tunables.c
mponomar: Good catch, thanks.  I meant to take it out altogether.

db/comdb2.c
mponomar: It ends up being on anyway because we set up signal handling before we set up tunables.  But certainly less confusing to set it to the default value.

PR #1645, opened 2019-06-20
cdb2api protobuf allocator

mohitkhullar:
from bbgithub


riverszhang89:
Can server also use this allocator?

mohitkhullar:
Thanks

PR #1646, opened 2019-06-20
Close old connections

mohitkhullar:
on reaching appsock limit, instead of giving error on new connections, close old connections.


mohitkhullar:
11:55:47> [cldeadlock-2] Failed to run update statement, rc=402, Client api should change nodes
there was crash after this because of race in this which appsock thread would cleanup the client structure and sql thread would still be working on it.

db/sqlinterfaces.c
```diff
@@ -4106,6 +4108,62 @@ __func__, cnonce);
...
+int check_active_appsock_connections(struct sqlclntstate *clnt)
+{
+    int max_appsock_conns =
+        bdb_attr_get(thedb->bdb_attr, BDB_ATTR_MAXAPPSOCKSLIMIT);
+    if (active_appsock_conns > max_appsock_conns) {
+        int num_retry = 0;
+        int rc = -1;
+    retry:
+        Pthread_mutex_lock(&clnt_lk);
+        Pthread_mutex_lock(&appsock_conn_lk);
```
akshatsikarwar: Would it be possible to use one lock + `ATOMIC_ADD` for counter?

mohitkhullar: active_appsock_conns needs to be under lock for this logic but got separate lock to reduce contention, as we will be inside this loop rarely.

PR #1648, opened 2019-06-22
Selectv rcodes

markhannum:
Fix rcodes for selectv and serializable.


markhannum:
sp test started breaking a while ago- I have bisected it to a commit- not related to this patch.

PR #1649, opened 2019-06-24
watcher_thread: Warn of long log truncation

riverszhang89:
Only warn of and abort for long log truncation.

(DRQS 144192348)


riverszhang89:
Thank you Mark!

PR #1650, opened 2019-06-24
Stop checking 'allow_lua_exec_with_ddl' tunable after Lua is setup.

mistachkin:


lua/sp.c
```diff
@@ -2086,12 +2086,7 @@ static int lua_prepare_sql_int(Lua L, SP sp, const char *sql,
 
 static int lua_get_prepare_flags()
 {
-    int prepFlags = PREPARE_DENY_DDL | PREPARE_IGNORE_ERR;
-
-    if (gbl_allow_lua_exec_with_ddl)
-        prepFlags &= ~PREPARE_DENY_DDL;
```
nirbhayc: This looked about right. What was the problem with this code?

mistachkin: That was actually part of the problem.  It was right, with the previous design (i.e. before the "allow_lua_exec_with_ddl" tunable was checked at Lua interpreter creation time).  Currently, the design is supposed to be that once the Lua interpreter is created and setup (i.e. perhaps with an "exec_with_ddl" command, based on the value of the tunable), the value of the tunable is not consulted for it again.

PR #1652, opened 2019-06-24
Faster Analyze

riverszhang89:
Significantly speed up analyze (2.5x ~ 3x) by reducing the cost of sorting. Instead of sampling individual entries and sorting them in a temptable, we sample and sort pages by their first key. This way we leverage the fact that entries on the same page are already sorted and are able to greatly reduce sorting cost.


nirbhayc:
Go go go!

riverszhang89:

PR #1653, opened 2019-06-25
Don't skip sqlite_stat* tables on a repoplrl

mponomar:
Legacy database creation needs to have those tables present.


mponomar:
Thanks Rivers.

PR #1654, opened 2019-06-25
Genid-based selectv rcodes should take precedence over serializable e…

markhannum:
My "simple" code to check serializability on the replicant broke return codes for genid-based selectv's.


markhannum:
These tests aren't failing for me .. (?)

PR #1657, opened 2019-06-26
Serial verify retry

markhannum:
Automatically retry snapshot and serial transactions if verify-retry is enabled, and if the client has not received any information (selects or effects).  This is controlled by the "snapshot_serial_verify_retry" tunable, which is enabled by default.


markhannum:
sc_blob_update is a real bug- difficult to reproduce and not related to this PR.  verify_del_constraints on the master is accessing a Btree which hasn't been opened.  Another thread is finalizing a schema change.  I've posted the stack-trace here:

https://gist.github.com/markhannum/c7e6da8607cb76753219b0c391db4028

tests/tools/selectv_rcode.c
```diff
@@ -179,21 +215,28 @@ int selectv_update_query(cdb2_hndl_tp *db)
...
             fprintf(stderr, "line %d error running commit: %d %s\n", __LINE__, rc, cdb2_errstr(db));
             exit(1);
         } else
             got_error = 1;
     }
 
     if ((rc = cdb2_next_record(db)) != CDB2_OK_DONE) {
-        if (rc != -103) {
+        if (is_common_acceptable_error(db, rc)) {
+        } else if (rc != -103) {
```
markhannum: Use CDB2ERR_CONSTRAINTS

tests/tools/selectv_rcode.c
```diff
@@ -70,27 +93,44 @@ static char *type_to_str(int type)
...
             break;
     }
 }
 
+int is_common_acceptable_error(cdb2_hndl_tp *db, int rc)
+{
+    switch(rc) {
+        /* Master lost transaction */
+        case -109:
+        case 203:
```
markhannum: CDB2ERR_DEADLOCK instead of 203 .. -109 is not referenced in cdb2api.h (maybe a different pr)

markhannum:
odh_blobs error is reproducible against master

PR #1659, opened 2019-06-27
Fix an incorrect usage of pthread_mutex_lock / pthread_mutex_unlock...

mistachkin:
These changes correct an incorrect (according to POSIX) usage of the pthread_mutex_lock and pthread_mutex_unlock APIs, in the schema change subsystem, i.e. a mutex must be unlocked on the same thread that locked it.


schemachange/schemachange.c
```diff
@@ -334,9 +333,13 @@ int start_schema_change_tran(struct ireq *iq, tran_type *trans)
             rc = pthread_create(&tid, &gbl_pthread_attr_detached,
                                 (void *(*)(void *))do_schema_change_locked, s);
         } else {
-            Pthread_mutex_lock(&s->mtx);
             rc = pthread_create(&tid, &gbl_pthread_attr_detached,
                                 (void *(*)(void *))do_schema_change_tran, arg);
+            if (rc == 0) {
+                Pthread_mutex_lock(&s->mtxStart);
```
akshatsikarwar: Should lock mtxStart before creating thread.

mistachkin: I don't understand.

PR #1660, opened 2019-06-27
Enhance handling of the Lua maximum instruction count

mistachkin:
Permit zero to mean 'unlimited' for the Lua maximum instruction count. Also, increase the limit for the 'send' command.


mohitkhullar:
Looks good

PR #1661, opened 2019-06-28
Adjust the workaround for the asynchronous nature of CREATE TIME PARTITION in the 'sp' test.

mistachkin:


mponomar:
Also need to adjust the test, it expects 20 in the output

PR #1664, opened 2019-07-03
Refactor code to run each stripe/key/blob verify in its own function

adizaimi:


adizaimi:
Performance is pretty decent, in NJ-BDTA4 on a table of 25GB and 6 indices, we get about 3-4 times performance gain:
```
time build/tools/cdb2sql/cdb2sql eqbasdb "EXEC PROCEDURE sys.cmd.verify('sec_monthly','parallel')" &> /dev/shm/out2

real    40m47.934s
user    0m0.054s
sys     0m0.070s

<ild/tools/cdb2sql/cdb2sql eqbasdb "EXEC PROCEDURE sys.cmd.verify('sec_monthly','parallel')" &> /dev/shm/out2

real    12m10.048s
user    0m0.019s
sys     0m0.025s
```

adizaimi:
Another datapoint with a larger table of about 65GB:

```
table = sec_daily, sz = 65.79GB  57% (dta 15.03GB, ix0 11.85GB, ix1 21.19GB, ix2 9.54GB, ix3 8.19GB)')

time build/tools/cdb2sql/cdb2sql eqbasdb "EXEC PROCEDURE sys.cmd.verify('sec_daily')" &> /dev/shm/out3

real    158m1.540s
user    0m0.287s
sys     0m0.223s

time build/tools/cdb2sql/cdb2sql eqbasdb "EXEC PROCEDURE sys.cmd.verify('sec_daily','parallel')" &> /dev/shm/out4

real    49m14.235s
user    0m0.076s
sys     0m0.078s
```

bdb/bdb_verify.c
```diff
@@ -521,564 +798,451 @@ static int bdb_verify_ll(
...
+ */
+void bdb_verify_handler(td_processing_info_t *info) 
+{
+    verify_common_t *par = info->common_params;
+    bdb_state_type *bdb_state = par->bdb_state;
+    int rc;
+    unsigned int lid;
+    BDB_READLOCK("bdb_verify");
+    
+    if ((rc = bdb_state->dbenv->lock_id_flags(bdb_state->dbenv, &lid,
```
mponomar: When you run this without parallel, this is a second lock you get in the same thread of control (tran was the first).

bdb/bdb_verify.c
```diff
@@ -273,245 +213,582 @@ static void printhex(SBUF2 *sb, int (*lua_callback)(void *, const char *),
...
+
+    while (rc == 0 && !par->client_dropped_connection) {
+        ATOMIC_ADD(par->records_processed, 1);
+        par->nrecs_progress++;
+
+        now = comdb2_time_epochms();
+        print_verify_progress(par, now);
+
+        /* check if comdb2sc is killed every 1000ms */
+        if (par->verify_mode == VERIFY_DEFAULT && (now - last) > 1000) {
```
mponomar: We still want to check this in parallel mode as well.

bdb/bdb_verify.c
```diff
@@ -273,245 +213,582 @@ static void printhex(SBUF2 *sb, int (*lua_callback)(void *, const char *),
...
+        return 0;
+    }
+    rc = ckey->c_get(ckey, &dbt_key, &dbt_data, DB_FIRST);
+    if (rc && rc != DB_NOTFOUND) {
+        par->verify_status = 1;
+        locprint(par->sb, par->lua_callback, par->lua_params, "!ix %d first rc %d\n", ix, rc);
+    }
+    while (rc == 0 && !par->client_dropped_connection) {
+        ATOMIC_ADD(par->records_processed, 1);
+        par->nrecs_progress++;
```
mponomar: Double counting records?

bdb/bdb_verify.c
```diff
@@ -273,245 +213,582 @@ static void printhex(SBUF2 *sb, int (*lua_callback)(void *, const char *),
...
+
+                    if (blobsizes[blobno] >= 0 && realblobsz[blobno] >= 0) {
+                        rc = par->add_blob_buffer_callback(
+                            blob_buf, dbt_blob_data.data,
+                            dbt_blob_data.size, blobno);
+                        if (rc)
+                            goto err;
+                    }
+
+                    if (dbt_blob_data.data && had_errors == 0)
```
mponomar: This may be a much older bug, but we always want to free.

bdb/bdb_verify.c
```diff
@@ -521,564 +798,451 @@ static int bdb_verify_ll(
                             free(dbt_blob_data.data);
```
mponomar: Also always want to free here.

bdb/bdb_verify.c
```diff
@@ -521,564 +798,451 @@ static int bdb_verify_ll(
...
-                locprint(sb, lua_callback, lua_params,
-                    "!verifying index %d, did %lld records, %d per second\n",
-                    ix, nrecs, (int)(nrecs_progress / progress_report_seconds));
-                last = now;
-                nrecs_progress = 0;
-                sbuf2flush(sb);
-            }
+    rc = cblob->c_get(cblob, &dbt_key, &dbt_data, DB_FIRST);
+    while (rc == 0 && !par->client_dropped_connection) {
+        ATOMIC_ADD(par->records_processed, 1);
```
mponomar: Counting the record again?

bdb/bdb_verify.c
adizaimi: Ok, will move into print_verify_progress. Somehow I did not think it was needed in parallel mode -- not sure why.

bdb/bdb_verify.c
adizaimi: Yes, just like nrecs++, I guess we want to know how many raw records each of these (data, blobs, indices) is processing so at the end we get ```nrecs = N * nRecords (with N = 1 (data) + B (numblobs) + I (numIdx)) ``` 
Just pointing out that i am not changing that behavior.

bdb/bdb_verify.c
adizaimi: This is the same as what we did before -- that logic did not change with this PR.

db/verify.c
```diff
@@ -390,3 +411,98 @@ int verify_table(const char *table, SBUF2 *sb, int progress_report_seconds,
...
+
+    if (rc) {
+        logmsg(LOGMSG_INFO, "Readlock table %s %d\n", table, rc);
+        if (sb)
+            sbuf2printf(sb, "?Readlock table %s rc %d\n", table, rc);
+        rc = 1;
+        goto done;
+    }
+
+    if (!gbl_verify_thdpool) {
```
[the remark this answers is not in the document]
adizaimi: Ok.

db/verify.c
```diff
@@ -390,3 +411,98 @@ int verify_table(const char *table, SBUF2 *sb, int progress_report_seconds,
...
+        .verify_mode = mode,
+    };
+
+    td_processing_info_t info = { .common_params =  &par };
+
+    // enqueue work to the threadpool queue
+    rc = bdb_verify_enqueue(&info, gbl_verify_thdpool);
+
+    // wait for all our enqueued work items to complete for this verify
+    while(par.threads_spawned > par.threads_completed) {
```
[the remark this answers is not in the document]
adizaimi: The accounting is atomic for the spawned and completed, I don't think cond_signal is needed. Why add one more global if this is enough?

bdb/bdb_verify.c
adizaimi: ok. fixing.

db/verify.c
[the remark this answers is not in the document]
mponomar: Ok.  This isn't a critical code path, no point in making it more complicated.

bdb/bdb_verify.c
mponomar: Sounds good.  As long as the output makes sense.  Want to avoid confusion of someone running select *, seeing 1000 records, then seeing progress reported as 1500 records.

adizaimi: I changed the output so that the 'default' way prints exactly what it did before, with same accounting of records (which is really bad if you study it--we zero the nrecs counter and progress counter for every stripe/index and not for blob entries).

adizaimi: Also for the parallel mode, we count 'items' instead of records now. A row will be counted multiple times with this method.

PR #1667, opened 2019-07-03
cdb2_load

akshatsikarwar:


adizaimi:
We should add a test case for these tools.

PR #1680, opened 2019-07-10
Materialized system tables

mponomar:
Adds the ability to create "system" tables that are backed by real tables.  Has a simple way to hook into changes to these tables and update internal data structures.  Not really meant for tables that get updated a great deal (eg: wouldn't use this for sequences, or at least not with the default callback mechanism).  Would really love to see more of llmeta migrate to use a system like this at some point.

This is incomplete.  Needs: (1) tests, (2) a way to upgrade the tables automatically, (3) some cleanup (remove my docker test changes maybe?)

Entering mostly as a RFC, feedback welcome.


sqlite/src/parse.y
```diff
@@ -282,7 +282,7 @@ columnname(A) ::= nm(A) typetoken(Y). {sqlite3AddColumn(pParse,&A,&Y);}
   BLOBFIELD BULKIMPORT
   CHECK COMMITSLEEP CONSUMER CONVERTSLEEP COUNTER COVERAGE CRLE
   DATA DATABLOB DATACOPY DBPAD DEFERRABLE DISABLE DISTRIBUTION DRYRUN
-  ENABLE FUNCTION GENID48 GET GRANT INCREMENT IPU ISC KW LUA LZ4 NONE
+  ENABLE FUNCTION GENID48 GET GRANT INCREMENT IPU ISC KW LUA LZ4 NONE DTASTRIPE
   ODH OFF OP OPTION OPTIONS
```
mistachkin: Minor thing: it would be nice to keep the keyword list in alphabetical order here.

tests/docker/Dockerfile.install
```diff
@@ -37,20 +38,23 @@ RUN apt-get update && \
...
 RUN cd /comdb2 &&  \
     rm -rf build && \
     mkdir build && \
     cd build && \
-    cmake -DWITH_TCL=1 .. && \
-    make -j4 && \
-    make -j4 test-tools && \
-    make install &&  \
+    cmake -DWITH_TCL=1 -DCMAKE_BUILD_TYPE=Debug -G Ninja .. && \
+    time ninja && \
```
mistachkin: I have a tendency to add the Debug build type here as well.  It would be nice if there were a way to do this without having to hard-wire it into this Docker file.

mponomar: We can have a separate Dockerfile for debug builds (maybe FROM a container that has all the prerequisites installed).

schemachange/sc_callbacks.c
```diff
@@ -186,6 +186,7 @@ unsigned long long get_genid_stripe_pointer(unsigned long long genid,
                                             unsigned long long *sc_genids)
 {
     int stripe = get_dtafile_from_genid(genid);
+    // TODO: can this ever be out of range?
```
mistachkin: It appears that get_dtafile_from_genid() can return a negative value if the genid is synthetic.  Using that negative value here would indeed be bad, given that it's used as an array index.  Related: since gbl_dtastripe is based on a tunable, I wonder how it is validated?

mponomar: gbl_dtastripe should be a readonly tunable - it's set at db create time.  There's a 'morestripe' command to expand it.  It's rarely done.

sqlite/ext/comdb2/fingerprints.c
```diff
@@ -113,18 +115,18 @@ int systblFingerprintsInit(sqlite3 *db)
...
         CDB2_INTEGER, "total_time", -1,
-        offsetof(struct fingerprint_track, time),
+        offsetof(struct fingerprint_track_systbl, time),
         CDB2_INTEGER, "total_rows", -1,
-        offsetof(struct fingerprint_track, rows),
+        offsetof(struct fingerprint_track_systbl, rows),
         CDB2_CSTRING, "normalized_sql", -1,
-        offsetof(struct fingerprint_track, zNormSql),
+        offsetof(struct fingerprint_track_systbl, zNormSql),
         SYSTABLE_END_OF_FIELDS);
```
mistachkin: It would be nice to include the new longreqLimit datum here as well.

db/comdb2.c
```diff
@@ -2123,14 +2127,18 @@ static int llmeta_load_tables(struct dbenv *dbenv, char *dbname, void *tran)
...
+        nstripes = db_get_dtastripe_by_name(tblnames[i], tran);
+        is_systable = db_get_is_systable_by_name(tblnames[i], tran);
+        tbl = newdb_from_schema(dbenv, tblnames[i], NULL, dbnums[i], i, 0, nstripes);
         if (tbl == NULL) {
             logmsg(LOGMSG_ERROR, "newdb_from_schema failed %s:%d\n", __FILE__,
                     __LINE__);
             rc = 1;
             break;
         }
         tbl->schema_version = ver;
```
adizaimi: What if we passed is_systable as a parameter to newdb_from_schema()? 
You don't need tbl->nstripes = nstripes; because nstripes is already set.

PR #1683, opened 2019-07-10
Don't do comdb2_row* fields for synthetic tables

mponomar:
Also fixes a memory leak from dispatching to a prefault pool that doesn't exist.


mponomar:
Akshat points out that I'm wrong about the memory leak.

sqlite/src/resolve.c
```diff
@@ -415,11 +415,11 @@ static int lookupName(
     ** the vm sniffs this out when it runs OP_Rowid and executes the comdb2
     ** backend call to get the rrn+genid.
     */
-    else if( cnt==0 && cntTab==1 && sqlite3IsComdb2Rowid(zCol) ){
+    else if( cnt==0 && cntTab==1 && sqlite3IsComdb2Rowid(pMatch->pTab, zCol) ){
        cnt = 1;
```
mistachkin: Upon closer examination, I think both changes here need "&& pMatch" just prior to the sqlite3IsComdb2*() calls...

mponomar: Thanks!

nirbhayc:
Could you also add a test please?

mistachkin:
That's a good point, it would be nice to have test coverage for this.  Also, having a test where pMatch ends up being false (e.g. column not found?) would be nice as well.

mponomar:
Ah, I had a test for this, that I somehow forgot to include (small comdb2sys change).

adizaimi:
I think t7 and t8 subtests are not run anymore after this change.

PR #1685, opened 2019-07-11
newsql: log out-of-memory message

nirbhayc:
After logging, we deliberately let the node to crash and reboot, subsequently, allowing the client to pick  up another (potentially healthier) node.
(DRQS 144692998)


nirbhayc:
As discussed with @mponomar & @riverszhang89, I will introduce a tunable to limit amount of memory to a connection, also catch and report OOM to the client.

PR #1688, opened 2019-07-11
 Fix random bad query plans

mohitkhullar:


db/sqlglue.c
```diff
@@ -574,6 +574,18 @@ static inline int check_recover_deadlock(struct sqlclntstate *clnt)
     return rc < 0 ? SQLITE_BUSY : rc;
 }
 
+static int isPrepare(BtCursor *pCur)
+{
```
mistachkin: Given what this function actually checks, the name seems a bit confusing.  Perhaps it can be named something like "cursor_in_db_init()" ?

sqlite/src/prepare.c
```diff
@@ -385,7 +385,7 @@ int sqlite3InitOne(sqlite3 *db, int iDb, char **pzErrMsg, u32 mFlags){
     sqlite3DbFree(db, zSql);
 #ifndef SQLITE_OMIT_ANALYZE
     if( rc==SQLITE_OK ){
-      sqlite3AnalysisLoad(db, iDb);
+      rc = sqlite3AnalysisLoad(db, iDb);
```
mistachkin: To help keep the comdb2 changes isolated, can this be changed to the following?

```
      rc = sqlite3AnalysisLoad(db, iDb);
#else /* defined(SQLITE_BUILDING_FOR_COMDB2) */
      sqlite3AnalysisLoad(db, iDb);
#endif /* defined(SQLITE_BUILDING_FOR_COMDB2) */
```

mistachkin:
Looks good to me.

mohitkhullar:
Thanks Joe and Mike

PR #1689, opened 2019-07-11
berkdb: Throttle REP_LOG messages.

riverszhang89:
The master easily runs out of memory if a replicant is quite behind, because we don't throttle `REP_LOG` responses to replicants' `ALL_REQ` and `LOG_REQ` requests.

The patch sends `REP_LOG_LOGPUT` messages instead for `ALL_REQ` and `LOG_REQ` requests. `REP_LOG_LOGPUT` is almost identical to `REP_LOG` except that it can be throttled. The type gets reverted to `REP_LOG` in rep_send so a replicant knows how to process these messages.

The patch also includes minor refactoring. Two new wrapper around `__rep_send_message` are introduced:
1) `__rep_send_log_more`, which forcibly sends `LOG_MORE` if the net queue is full;
2) `__rep_time_send_message`, which times __rep_send_message.

(DRQS 145076276)

### DIFF
--- a/berkdb/rep/rep_record.c
+++ b/berkdb/rep/rep_record.c
@@ -453,7 +453,6 @@ int gbl_verbose_fills = 0;
 
 static int queue_log_more_count = 0;
 static int queue_log_fill_count = 0;
-int gbl_trace_repmore_reqs = 0;
 int gbl_verbose_repdups = 0;
 uint64_t subtract_lsn(void *bdb_state, DB_LSN *lsn1, DB_LSN *lsn2);
 void comdb2_early_ack(DB_ENV *, DB_LSN, uint32_t generation);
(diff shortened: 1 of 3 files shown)


berkdb/rep/rep_record.c
```diff
@@ -1406,30 +1407,12 @@ __rep_process_message(dbenv, control, rec, eidp, ret_lsnp, commit_gen, online)
...
-						logmsg(LOGMSG_USER, "%s line %d failed to send "
-								"newfile to %s for LSN %d:%d %d\n",
-								__func__, __LINE__, *eidp, oldfilelsn.file,
-								oldfilelsn.offset, rc);
-					}
+				if ((rc = __rep_time_send_message(dbenv,
+								*eidp, REP_NEWFILE, &oldfilelsn, NULL,
+								sendflags, NULL, &sendtime)) != 0) {
+					/* Net queue could be full. Send LOG_MORE to throttle. */
+					goto more;
```
riverszhang89: remove this .. we'll never come here.

riverszhang89:
Thank you for the review Mark!

PR #1693, opened 2019-07-15
Add a verify callback for the 'dtastripe' tunable parameter.

mistachkin:
These changes should prevent an invalid 'dtastripe' tunable value (i.e. less than one or greater than sixteen) from being used.


nirbhayc:
Also add this check for `private_blkseq_stripes` ? It's defined in bdb sub-system.

mistachkin:
> private_blkseq_stripes

Good catch, thank you.

PR #1704, opened 2019-07-17
Print memp_fgets and memp_pgs in statreqs

adizaimi:
Print memp_fgets and memp_pgs in statreqs

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -4559,7 +4559,7 @@ void *statthd(void *p)
                     unsigned nwaits = cur_bdb_stats.n_lock_waits -
                                       last_bdb_stats.n_lock_waits;
                     reqlog_logf(statlogger, REQL_INFO,
-                                "%u locks, avg time %ums, worst time=%ums\n",
+                                "%u locks, avg time %ums, worst time %ums\n",
                                 nwaits,
                                 U2M(cur_bdb_stats.lock_wait_time_us -
                                     last_bdb_stats.lock_wait_time_us) /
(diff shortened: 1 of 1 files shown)


db/comdb2.c
```diff
@@ -4559,7 +4559,7 @@ void *statthd(void *p)
                     unsigned nwaits = cur_bdb_stats.n_lock_waits -
                                       last_bdb_stats.n_lock_waits;
                     reqlog_logf(statlogger, REQL_INFO,
-                                "%u locks, avg time %ums, worst time %ums\n",
+                                "%u locks, avg time %ums, worst time=%ums\n",
```
riverszhang89: Why do we need the `=` here?

db/comdb2.c
```diff
@@ -4589,6 +4589,16 @@ void *statthd(void *p)
                                     last_bdb_stats.pwrite_time_us) /
                                     npwrites);
                 }
+                if (cur_bdb_stats.n_memp_fgets > last_bdb_stats.n_memp_fgets) {
+                    unsigned n_memp_fgets = cur_bdb_stats.n_memp_fgets - last_bdb_stats.n_memp_fgets;
+                    unsigned us = cur_bdb_stats.memp_fget_time_us - last_bdb_stats.memp_fget_time_us;
+                    reqlog_logf(statlogger, REQL_INFO, "n_memp_fgets=%u, avg time=%ums\n", U2M(us) / n_memp_fgets);
```
riverszhang89: The format has 2 arguments, but only 1 is provided? Also maybe rename avg time to something else to not confuse with the one above.

db/comdb2.c
adizaimi: leftover from other related work--removing. thanks

db/comdb2.c
adizaimi: Fixing (leftover from other branch). Thanks.

PR #1711, opened 2019-07-18
Fix cdb2_verify crashes: was missing Pthread_key_create

adizaimi:
cdb2_verify was crashing because it was missing a call to Pthread_key_create for DBG_FREE_CURSOR. This PR fixes the crash.


akshatsikarwar:
Why do we have `#ifndef BERKDB_46` inside of code from berkdb 4.2? Can we remove those - it will never be set.

Also do the other tools (dump/load/stat) also need this?

PR #1715, opened 2019-07-19
Send an dummy response for COMMIT/ROLLBACK for a failed tx

nirbhayc:
This fixes a regression where old client APIs expect a response from Comdb2.

### DIFF
--- a/db/sql.h
+++ b/db/sql.h
@@ -198,16 +198,23 @@ typedef struct osqlstate {
 } osqlstate_t;
 
 enum ctrl_sqleng {
-    SQLENG_NORMAL_PROCESS =
-        0, /* no user specified transactions, i.e. begin/commit */
-    SQLENG_PRE_STRT_STATE =
-        1, /* "begin" was submitted, mark this as user transaction begin */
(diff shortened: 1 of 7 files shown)


mohitkhullar:
Looks good

akshatsikarwar:
What is the old client API?

nirbhayc:
go-cdb2, IIUC it's implementation was based on R6's cdb2api.

akshatsikarwar:
cdb2api should be the same regardless of which implementation, no?

akshatsikarwar:
Sorry, when you say old client API, I think fastsql..

nirbhayc:
> cdb2api should be the same regardless of which implementation, no?

yea, and also the wire protocol.

nirbhayc:
This PR is about fixing a fairly recent change in the wire protocol (0fe3cb0 & 89fb44d). go-cdb2 API, since its inception, expected a response on COMMIT/ROLLBACK even for failed transactions. But, due to the regression, Comdb2 no longer sends one for failed transactions.

nirbhayc:
Cherry-picked to 7.0

PR #1719, opened 2019-07-23
Testsuite: some small changes to tests

adizaimi:
Some small changes to tests: 
- parameter to start DB in setup script
- move a testcase from sc_insert to verify_db
- clean a bit verify_db


adizaimi:
selectv test failure is something I observed before and unrelated to this:
18:35:49> line 189 error running commit: 230 transaction is not serializable

PR #1725, opened 2019-07-24
Use Pthread wrappers for berkely mutex lock functions

adizaimi:
Use Pthread wrappers for berkely mutex lock functions.


markhannum:
Looks great- okay to merge

adizaimi:
Thank you Mark.

PR #1726, opened 2019-07-24
Way to banish misbehaving nodes

mponomar:
Also banish master after enough attempts to tell them to downgrade.


mponomar:
Needs test case.

PR #1727, opened 2019-07-25
berkdb: use separate free_mutexp to add to free cursor list

adizaimi:
Berkdb uses mutexp to remove from active list and same mutex
to add to free cursor list. That same mutex is used for other
operations done on cursor.
Using a new mutex to add to free cursor list should reduce 
observed contention on mutexp.

### DIFF
--- a/berkdb/build/db.h
+++ b/berkdb/build/db.h
@@ -1381,6 +1381,7 @@ struct __db {
 	DB_MPOOLFILE *mpf;		/* Backing buffer pool. */
 
 	DB_MUTEX *mutexp;		/* Synchronization for free threading */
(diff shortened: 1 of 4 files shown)


adizaimi:
The read performance gain from this simple change is pretty significant (assuming repeat tests prove that the above gains are real and not just random sysbench fluctuation).

nirbhayc:
Amazing find Adi!! It looks like a real gain to me. I have resubmitted just go be sure. Cheers!

adizaimi:
Thanks Nirbhay, looks like the second time sysbench showed same gain pattern. 
I wonder how many other similar contention points we might have.

adizaimi:
Looks like this is a real gain. Checking this in. There might be other ways of making this even faster that i will look at in the future.

PR #1733, opened 2019-07-26
Add experimental support for precise timing with GCC/x86 (on Linux).

mistachkin:
These changes introduce two (experimental) functions that can be used to precisely time operations within the database.  Also, example usage is included within dispatch_sql_query().

### DIFF
--- a/db/sqlinterfaces.c
+++ b/db/sqlinterfaces.c
@@ -110,6 +110,10 @@
 
 #include "dohsql.h"
 
+#if defined(__GNUC__) && (defined(i386) || defined(__i386__) || defined(__x86_64__))
+#include <x86intrin.h>
+#endif
+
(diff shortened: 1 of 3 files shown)


db/sqlinterfaces.c
```diff
@@ -236,6 +240,31 @@ static int sql_debug_logf_int(struct sqlclntstate *clnt, const char *func,
...
+#endif
+}
+
+double get_cpu_cycle_freq()
+{
+    static double freq = 0.0;
+    if (freq == 0.0) {
+        struct timespec req;
+        memset(&req, 0, sizeof(struct timespec));
+        req.tv_nsec = 100000; /* 10ms */
```
dorinhogea: math does not look right to me

mistachkin: Thanks...  I think it's better now.

PR #1735, opened 2019-07-29
LIST_EMPTY: Fix conflicting definition (#1734)

akshatsikarwar:


akshatsikarwar:
Fixes conflict with https://github.com/bloomberg/comdb2/blob/master/berkdb/dbinc/queue.h#L405

PR #1743, opened 2019-07-31
Small refactoring in rep.c

adizaimi:
Small refactoring in rep.c


bdb/rep.c
```diff
@@ -2116,11 +2113,12 @@ void bdb_get_notcoherent_list(bdb_state_type *bdb_state,
...
-                    nodes_list[*num_notcoherent] = hostlist[i];
-                (*num_notcoherent)++;
-            }
+            if ((*num_notcoherent) < max_nodes)
+                nodes_list[*num_notcoherent] = hostlist[i];
+            (*num_notcoherent)++;
+
+            if (bdb_state->repinfo->skipsinceepoch == 0)
+                abort();
         }
```
mistachkin: There are no circumstances where skipsinceepoch can legitimately be zero here?

bdb/rep.c
```diff
@@ -2843,25 +2848,25 @@ static void bdb_slow_replicant_check(bdb_state_type *bdb_state,
...
-                malloc(sizeof(wait_for_lsn_list));
-            listc_init(bdb_state->seqnum_info->waitlist[nodeix(host)],
-                       offsetof(struct waiting_for_lsn, lnk));
-        }
+    struct waiting_for_lsn *waitforlsn;
+    int node_ix = nodeix(host);
+    if (bdb_state->seqnum_info->waitlist[node_ix] == NULL) {
+        bdb_state->seqnum_info->waitlist[node_ix] =
+            malloc(sizeof(wait_for_lsn_list));
+        listc_init(bdb_state->seqnum_info->waitlist[node_ix],
```
mistachkin: Perhaps this malloc should be a calloc?

bdb/rep.c
adizaimi: I guess i was trying to assert that skipsinceepoch is non zero. This check is not needed, i'm removing it -- we don't gain anything from it. skipsinceepoch is set every time a node is not coherent.

bdb/rep.c
adizaimi: Subsequent listc_init() sets all members of the listc_t so we don't need calloc here.

PR #1754, opened 2019-08-06
fix default fallback since it is part of our sys tables

dorinhogea:
Currently we use "default" as a column name in comdb2_procedures system table, but that is a keyword with no fallback. Fix that.


sqlite/src/parse.y
```diff
@@ -254,7 +254,7 @@ columnname(A) ::= nm(A) typetoken(Y). {sqlite3AddColumn(pParse,&A,&Y);}
 //
 %fallback ID
   ABORT ACTION AFTER ANALYZE ASC ATTACH BEFORE BEGIN BY CASCADE CAST COLUMNKW
-  CONFLICT DATABASE DEFERRED DESC DETACH DO
+  CONFLICT DATABASE DEFAULT DEFERRED DESC DETACH DO
   EACH END EXCLUSIVE EXPLAIN FAIL FOR
   IGNORE IMMEDIATE INITIALLY INSTEAD LIKE_KW MATCH NO PLAN
```
mistachkin: Would it be possible to move this into the Comdb2 specific portion of the fallback list (starting at line 281)?

PR #1757, opened 2019-08-07
comdb2_user() function

nirbhayc:
comdb2_user() returns the name of the current authenticated user.

### DIFF
--- a/sqlite/src/func.c
+++ b/sqlite/src/func.c
@@ -804,9 +804,7 @@ static void comdb2CtxinfoFunc(
 
   zName = (const char *)sqlite3_value_text(argv[0]);
   if( sqlite3_stricmp(zName, "parallel")==0 ){
-    if (clnt && clnt->conns) {
-      sqlite3_result_int(context, 1);
-    }
+    sqlite3_result_int(context, clnt!=NULL && clnt->conns!=NULL);
   } else if( sqlite3_stricmp(zName, "user")==0 ){
     if (clnt) {
       sqlite3_result_text(context, get_current_user(clnt), -1, SQLITE_STATIC);
--- a/sqlite/src/func.c
+++ b/sqlite/src/func.c
@@ -24,6 +24,8 @@
 #include "comdb2.h"
 #include "sql.h"
 #include "bdb_int.h"
+
+char *get_current_user(struct sqlclntstate *clnt);
 #endif /* defined(SQLITE_BUILDING_FOR_COMDB2) */
 
 /*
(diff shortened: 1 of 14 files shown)


mistachkin:
Couldn't we just add this to comdb2_ctxinfo()?

nirbhayc:
Yes, I have add it to comdb2_ctxinfo() as well.

sqlite/src/func.c
```diff
@@ -796,11 +798,16 @@ static void comdb2CtxinfoFunc(
...
+  struct sql_thread *thd = pthread_getspecific(query_info_key);
+  struct sqlclntstate *clnt = thd!=NULL ? thd->clnt : NULL;
+
   zName = (const char *)sqlite3_value_text(argv[0]);
   if( sqlite3_stricmp(zName, "parallel")==0 ){
-    struct sql_thread *thd = pthread_getspecific(query_info_key);
-    struct sqlclntstate *clnt = thd!=NULL ? thd->clnt : NULL;
     sqlite3_result_int(context, clnt!=NULL && clnt->conns!=NULL);
+  } else if( sqlite3_stricmp(zName, "user")==0 ){
+    sqlite3_result_text(context, (clnt!=NULL) ? get_current_user(clnt) : "",
```
mistachkin: If the clnt is NULL or the return value of get_current_user is NULL, simply skipping the call to sqlite3_result_text may be good here as that would return a SQL NULL to the caller?

sqlite/src/func.c
```diff
@@ -796,11 +798,17 @@ static void comdb2CtxinfoFunc(
...
+  struct sqlclntstate *clnt = thd!=NULL ? thd->clnt : NULL;
+
   zName = (const char *)sqlite3_value_text(argv[0]);
   if( sqlite3_stricmp(zName, "parallel")==0 ){
-    struct sql_thread *thd = pthread_getspecific(query_info_key);
-    struct sqlclntstate *clnt = thd!=NULL ? thd->clnt : NULL;
     sqlite3_result_int(context, clnt!=NULL && clnt->conns!=NULL);
+  } else if( sqlite3_stricmp(zName, "user")==0 ){
+    if (clnt) {
+      sqlite3_result_text(context, get_current_user(clnt), -1, SQLITE_STATIC);
```
mistachkin: What if get_current_user returns null?

sqlite/src/func.c
```diff
@@ -796,11 +798,17 @@ static void comdb2CtxinfoFunc(
...
   }
+
+  struct sql_thread *thd = pthread_getspecific(query_info_key);
+  struct sqlclntstate *clnt = thd!=NULL ? thd->clnt : NULL;
+
   zName = (const char *)sqlite3_value_text(argv[0]);
   if( sqlite3_stricmp(zName, "parallel")==0 ){
-    struct sql_thread *thd = pthread_getspecific(query_info_key);
-    struct sqlclntstate *clnt = thd!=NULL ? thd->clnt : NULL;
     sqlite3_result_int(context, clnt!=NULL && clnt->conns!=NULL);
```
mistachkin: For parallel, I was thinking you could use the same "if (clnt) to wrap the sqlite3_result_int call...", e.g.:

```
if (clnt) {
  sqlite3_result_int(context, clnt->conns!=NULL);
}
```

sqlite/src/func.c
nirbhayc: null = "no user", I have tests that covers the scenario.

sqlite/src/func.c
nirbhayc: Skipping the call would result in NULL getting returned client, that, in addition to 0 and 1. Do you really want that? Also, it looks like clnt would never be null here.

mistachkin: Yes, it would return a SQL NULL in the event that clnt is NULL.  Currently, as you point out, this should be "impossible"; however, that may not always be the case.

sqlite/src/func.c
mistachkin: Ah, I see.  The original code returned an empty string if clnt was NULL; however, it returned NULL when get_current_user returned NULL, which would then translate to a SQL NULL.  Given that, my snippet above is unnecessary.

PR #1761, opened 2019-08-09
getmyseqnum can legitimately return a zero'd lsn: don't wait-for-seqn…

markhannum:
toblock calls "trans_wait_for_last_seqnum" when aborting a transaction, which calls bdb_get_myseqnum to retrieve the most recent lsn.  The problem is that bdb_get_myseqnum may legitimately return 0:0 if the node is catching up, or if the database is exiting.  Simple solution: don't call wait_for_seqnum if bdb_get_myseqnum returns a 0:0 LSN.


markhannum:
Thanks Nirbhay!

PR #1765, opened 2019-08-12
pmux.cpp: Changed NULLs to nullptrs.

levinem:
This is a small feature update.  
This is not a bugfix.
This is not a bugfix.
The current pmux.cpp used NULL pointers, which can lead to undefined behavior.  This PR changes the usages of NULL to nullptr in line with C++11 recommendations.
The pmux test passes with this change, and using nullptr instead of NULL is a recommended practice starting with C++11.


levinem:
Not sure if this will work, but /runtests

PR #1772, opened 2019-08-13
Fixes to prevent unintentional duplicate writes within a stored procedure.

mistachkin:


akshatsikarwar:
This is different from #2544 (already reviewed and merged into master)?

PR #1773, opened 2019-08-14
Auth: Reject unauthorized access for direct count

nirbhayc:
Fixes the following issue where unauthorized access was allowed for count(*).
```
nctdb> select * from t1;
[select * from t1] failed with rc -106 Read access denied to t1 for user foo bdberr=15
nctdb> select count(*) from t1;
+----------+
| count(*) |
+----------+
| 3        |
+----------+
[select count(*) from t1] rc 0
```


nirbhayc:
Thanks @mponomar !

nirbhayc:
Cherry-picked to 7.0

PR #1777, opened 2019-08-20
cdb2sql: Replaced while loops for level 1 and charatglyph words usage with range based for loops. Removed NULL from the arrays.

levinem:
This is a code refactoring.
This is not a bugfix.
This is not a bugfix.
The current behavior uses a nullptr and a while loop along with additional variable assignments to range through the two arrays.  The refactor includes the removal of NULL from the end of two word arrays and the updating of the while loop into a range-based for loop for iterating through those arrays.
The new code is more concise, follows more modern conventions (using a range-based for loop for iterating through arrays, and eliminates the need for some of the interim variables.

### DIFF
--- a/tools/cdb2sql/cdb2sql.cpp
+++ b/tools/cdb2sql/cdb2sql.cpp
@@ -199,11 +199,10 @@ const char *char_atglyph_words[] = {
 static char *char_atglyph_generator(const char *text, int state)
 {
     static int len;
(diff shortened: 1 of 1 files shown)


riverszhang89:
Thank you!

riverszhang89:
This looks cleaner than before. Good to merge.

PR #1780, opened 2019-08-22
newsql: Do not drop connection on error

nirbhayc:
.


nirbhayc:
This was an experiment to check the performance impact; no visible gains, will revisit later.

PR #1785, opened 2019-08-24
Disallow an incoherent node to process newsql queries

riverszhang89:
The change is needed to address the sockpool connection issue when nodes in 1 data center are up but not serving requests (due to incoherence, catching up with the master, etc.)


plugins/newsql/newsql.c
```diff
@@ -2344,7 +2344,7 @@ static int handle_newsql_request(comdb2_appsock_arg_t *arg)
         /* avoid new accepting new queries/transaction on opened connections
            if we are incoherent (and not in a transaction). */
         if (!clnt.admin && clnt.ignore_coherency == 0 &&
-            !bdb_am_i_coherent(thedb->bdb_env) &&
+            !gbl_allow_incoherent_sql && !bdb_am_i_coherent(thedb->bdb_env) &&
```
nirbhayc: I’d move this and other similar checks into a func or macro.

PR #1787, opened 2019-08-27
Smartbeat

riverszhang89:
When a query is taking too much time, we don't know whether this is due
to its complex nature or a hung backend. To know this, We enhance our
heartbeats by embedding the node state (e.g., whether the node is progressing)
in a heartbeat packet. The node state can help us tell an expensive query and
a stuck query apart, through a monitoring tool built around our API
event-handling mechanism (e.g., hungserv).

### DIFF
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -598,6 +598,9 @@ static int sql_tick(struct sql_thread *thd)
     if (clnt == NULL)
         return 0;
 
+    /* Increment per-clnt sqltick */
+    ++clnt->sqltick;
+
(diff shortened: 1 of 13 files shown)


db/sqlglue.c
```diff
@@ -598,14 +598,17 @@ static int sql_tick(struct sql_thread *thd)
 
     gbl_sqltick++;
 
+    if (thd == NULL)
```
riverszhang89: ??$$

riverszhang89:
@nirbhayc Thanks for the suggestion! 

PR #1789, opened 2019-08-27
Fix invalid transaction state transition

nirbhayc:
On preemptive abort, do not set clnt's transaction state to SQLENG_FNSH_ABORTED_STATE, if we are not in a client transaction.


nirbhayc:
@adizaimi Thanks!

nirbhayc:
Cherry-picked to 7.0.

PR #1795, opened 2019-09-04
csc2: Simplify the GC logic

riverszhang89:
We keep track of malloc chunks using a linkedlist and free them all when done, in the csc2 subsystem. The logic can be implemented by simply using a dlmalloc allocator. Also get rid of the per-thread CSC2 allocator as it's needless now.


riverszhang89:
This is passing robomark. @mponomar Thank you for the review!

PR #1799, opened 2019-09-09
Add 'strict_double_quotes' tunable to forbid use of double-quoted string literals

mistachkin:
These changes add a new tunable named 'strict_double_quotes'.  When this tunable is enabled, any attempt to use a double-quoted string literal will result in a SQL syntax error.


tests/strict_dbl_quotes.test/t02.req.out
```diff
@@ -0,0 +1,21 @@
+('------- PHASE 1 -------'='------- PHASE 1 -------')
+("1"='1')
+("x"='x')
+(x='1')
+("y"='y')
+('------- PHASE 2 -------'='------- PHASE 2 -------')
+[SELECT "1"] failed with rc -3 double-quoted string literal: "1"
+[SELECT "x"] failed with rc -3 double-quoted string literal: "x"
+(x='1')
+[SELECT "y" FROM t1 WHERE x = 1] failed with rc -3 double-quoted string literal: "y"
```
nirbhayc: This should have passed. We're only forbidding double-quoted string literals. Double-quoted identifiers are perfectly valid (and also SQL standard).

sqlite/src/resolve.c
```diff
@@ -514,6 +515,12 @@ static int lookupName(
   if( cnt==0 && zTab==0 ){
     assert( pExpr->op==TK_ID );
     if( ExprHasProperty(pExpr,EP_DblQuoted) ){
+#if defined(SQLITE_BUILDING_FOR_COMDB2)
```
nirbhayc: We should only be forbidding double-quoted TK_STRINGs (and not TK_IDs).

tests/strict_dbl_quotes.test/t02.req.out
mistachkin: Which query should have passed here?  The one on line 10?  For that one, the t1 table has no column "y".  Several others are similar.

sqlite/src/resolve.c
mistachkin: If I recall correctly, the token code here has more to do with the expectation than reality.  I think it's basically just the difference between single and double quotes, as far as the tokenizer is concerned.

tests/strict_dbl_quotes.test/t02.req.out
mistachkin: Just added some more tests to verify that valid double-quoted identifiers are working correctly with the tunable enabled.

nirbhayc: I stand corrected. I thought "y" was a column in t1.

tests/strict_dbl_quotes.test/lrl.options
```diff
@@ -0,0 +1 @@
+enable_sql_stmt_caching NONE
```
nirbhayc: Why?

sqlite/src/resolve.c
nirbhayc: Ok, while looking at this patch, I wrongly assumed that we haven't yet looked up for a matching column name at this point. I'm ok with this change.

tests/strict_dbl_quotes.test/lrl.options
mistachkin: Without this, the previously prepared statements that had double-quoted string literals will not be re-prepared, allowing it to pass incorrectly.

nirbhayc: makes sense.

PR #1801, opened 2019-09-10
replicant_can_retry: Fix check for retryable txns

akshatsikarwar:


adizaimi:
Looks good, but are we adding a testcase?

akshatsikarwar:
> Looks good, but are we adding a testcase?
Missed adding to commit - its there now.

PR #1804, opened 2019-09-11
comdb2ar: Refactor modes_enum into an enum class.

levinem:
It is a refactoring.
It is not a bugfix.
It is not a bugfix.
The behavior should be largely the same.
Modern C++ recommends using enum classes over regular enums to help minimize the possibility of traditional enums being converted to ints unintentionally.  Also, I made all the enum class elements lower case to avoid potential clashes with current/future macros.

### DIFF
--- a/tools/comdb2ar/comdb2ar.cpp
+++ b/tools/comdb2ar/comdb2ar.cpp
@@ -89,8 +89,8 @@ int main(int argc, char *argv[])
     extern char *optarg;
     extern int optind, optopt;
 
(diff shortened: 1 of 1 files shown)


levinem:
I was able to build comdb2 with my changes.  Please let me know if you want me to run clang_format on the entire file, parts of the file, etc.  Thanks!

riverszhang89:
Thank you for the patch!

levinem:
No problem @riverszhang89 , glad I could help!

PR #1809, opened 2019-09-12
Auth: A special commandline option to bypass access checks

nirbhayc:
With this change, Comdb2 can be started with an extra argument **--insecure**, in order to temporarily bypass the access controls. This would come in handy when one needs to reset 'op' password.


nirbhayc:
Cherry-picked to 7.0.

nirbhayc:
@akshatsikarwar Thanks for the review!

PR #1810, opened 2019-09-13
Reorder index add del operations to avoid deadlocks

adizaimi:
Reorder index operations (add/upd/del) by table and index value and perform them deferred at the end of the osql processing. This ensures that we avoid deadlocking by touching (locking) pages in an increasing order from all the transactions on master node.
Same as PR https://github.com/bloomberg/comdb2/pull/1345 which was closed.


nirbhayc:
Do we know why oltp_read_write numbers dropped?

adizaimi:
FTR, we investigated why the numbers dropped for oltp_read_write and it stems from the fact that there is an *increased* number of deadlocks for this test: read statements first lock index pages then lock data pages, and that is the exact opposite order of reordered transaction writes which lock data pages first and index pages last. We are investigating ways to revert the order of the later in replicants to avoid deadlocks.

adizaimi:
Mark is trying get some performance estimates with this reordering off in PR https://github.com/bloomberg/comdb2/pull/2148.

PR #1819, opened 2019-09-23
Detect and reject use of BEGIN, COMMIT, ROLLBACK, SAVEPOINT, and RELEASE through the SQLite parser

mistachkin:
These changes are designed to prevent any direct use of the `BEGIN`, `COMMIT`, `ROLLBACK`, `SAVEPOINT`, and `RELEASE` commands via the SQLite parser.  This has the effect of preventing any callers of sqlite3_prepare*() from using them, including Lua stored procedures.


akshatsikarwar:
LGTM 👍

PR #1820, opened 2019-09-24
Fix Apple/Darwin build

prontidisbb:
Currently cmake fails on Darwin platforms. This PR aims to fix the build.


akshatsikarwar:
Thank you!

PR #1822, opened 2019-09-25
Add 'class' to the comdb2_sysinfo() SQL function

mistachkin:
These changes permit easier programmatic introspection of the class associated with a particular running database instance.

### DIFF
--- a/sqlite/src/func.c
+++ b/sqlite/src/func.c
@@ -821,6 +821,8 @@ static void comdb2SysinfoFunc(
     }else{
       sqlite3_result_error(context, "unable to obtain host name", -1);
     }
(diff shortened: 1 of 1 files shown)


adizaimi:
I think this should be in plugins if possible, and only exposed to bb.

mistachkin:
> 
> 
> I think this should be in plugins if possible, and only exposed to bb.

Even though the core functionality behind this is in the core itself?  There are certain use cases where being able to introspect this information could be very important.

adizaimi:
I guess others can use functionality or not, as long as we have a default, this is ok.

mohitkhullar:
looks good, thanks

PR #1824, opened 2019-09-27
Ported to R7: parallel verify

adizaimi:
Use a threadpool to perform parallel verify via command exec procedure sys.cmd.verify('t1', 'parallel'):
* Split out existing code to verify in a separate function datastripe, blob, and index
* Create threadpool to perform parallel verify of a table
* Use same threadpool ability to only verify data stripes, only verify blobs, and only verify indices by changing the way we call lua procedure to exec procedure sys.cmd.verify('t1', 'data') -- use 'blobs' or 'indices' for blob or index verification.


akshatsikarwar:
whats going on?

@@ sysbench/select_random_ranges @@
+9.68%
@@ sysbench/oltp_read_only @@
+19.97%

adizaimi:
R7 results compare with sysbench performance of master branch, which is 19% slower because of reorder indices.

PR #1827, opened 2019-09-27
Add 'group_concat_memory_limit' tunable

nirbhayc:
* Caps memory used by group_concat function
* Can also be changed for the session via SET command
* Defaults to 255MiB (=MAXBLOBSIZE)

### DIFF
--- a/db/db_tunables.c
+++ b/db/db_tunables.c
@@ -313,6 +313,7 @@ extern int gbl_selectv_writelock_on_update;
 extern int gbl_selectv_writelock;
 
 int gbl_debug_tmptbl_corrupt_mem;
+int gbl_group_concat_mem_limit = MAXBLOBLENGTH; // = 255 MiB
 
 extern int gbl_clean_exit_on_sigterm;
 extern int gbl_debug_omit_dta_write;
--- a/plugins/newsql/newsql.c
+++ b/plugins/newsql/newsql.c
@@ -1765,6 +1765,20 @@ static int process_set_commands(struct dbenv *dbenv, struct sqlclntstate *clnt,
                 printf("setting clnt->osql_max_trans to %d\n",
                        clnt->osql_max_trans);
 #endif
+            } else if (strncasecmp(sqlstr, "groupconcatmemlimit",
+                                   sizeof("groupconcatmemlimit") - 1) == 0) {
+                sqlstr += sizeof("groupconcatmemlimit");
+                int sz = strtol(sqlstr, &endp, 10);
+                if (endp != sqlstr && sz >= 0)
+                    clnt->group_concat_mem_limit = sz;
+                else
+                    logmsg(LOGMSG_ERROR,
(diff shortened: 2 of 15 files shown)


sqlite/src/func.c
```diff
@@ -2254,7 +2254,21 @@ static void groupConcatStep(
   if( pAccum ){
     sqlite3 *db = sqlite3_context_db_handle(context);
     int firstTerm = pAccum->mxAlloc==0;
+#if defined(SQLITE_BUILDING_FOR_COMDB2)
+    struct sql_thread *thd = pthread_getspecific(query_info_key);
+    struct sqlclntstate *clnt = thd!=NULL ? thd->clnt : NULL;
+
```
mistachkin: FYI, for these two lines, which are a commonly used pattern, I created a function get_sql_clnt().

sqlite/src/printf.c
```diff
@@ -893,7 +893,7 @@ static int sqlite3StrAccumEnlarge(StrAccum *p, int N){
     char *zOld = isMalloced(p) ? p->zText : 0;
     i64 szNew = p->nChar;
     szNew += N + 1;
-    if( szNew+p->nChar<=p->mxAlloc ){
+    if( (p->noExtraAlloc==0) && (szNew+p->nChar<=p->mxAlloc) ){
       /* Force exponential buffer size growth as long as it does not overflow,
```
mistachkin: Can we wrap this customization in an #ifdef block to prevent merge conflicts later?

sqlite/src/func.c
nirbhayc: Fixed, thank you.

sqlite/src/printf.c
nirbhayc: done

plugins/newsql/newsql.c
```diff
@@ -446,21 +446,31 @@ static struct newsql_appdata *get_newsql_appdata(struct sqlclntstate *clnt,
...
         clnt->appdata = appdata;
+        if (!appdata)
+            goto oom;
         appdata->capacity = n;
     }
     appdata->count = ncols;
     return appdata;
+oom:
+    logmsg(LOGMSG_ERROR, "%s:%d failed to (re)alloc %lu bytes\n", __func__,
+           __LINE__, alloc_sz);
```
riverszhang89: strerror maybe?

db/db_tunables.c
```diff
@@ -313,6 +313,7 @@ extern int gbl_selectv_writelock_on_update;
 extern int gbl_selectv_writelock;
 
 int gbl_debug_tmptbl_corrupt_mem;
+int gbl_group_concat_mem_limit = MAXBLOBLENGTH; // = 255 MiB
```
riverszhang89: Make it 0 by default

nirbhayc:
@riverszhang89 @mistachkin Thanks for the review!

PR #1831, opened 2019-09-30
Early verify in update with limit

mohitkhullar:


dorinhogea:
Please add a test case, otherwise looks ok.  Seekrowid was introduced after the early verify patch in a new sqlite.

PR #1832, opened 2019-10-01
Add constraint checking to verify

adizaimi:
With this change, verify will now support checking correctness of foreign constraints.
Added testcase that simulates failures [to check foreign constraints when inserting], 
and makes sure the inconsistency is caught by verify.


db/db_tunables.h
```diff
@@ -1236,6 +1236,10 @@ REGISTER_TUNABLE("debug.omit_blob_write",
                  "Deliberately corrupt insertion randomly to debug db_verify", TUNABLE_BOOLEAN,
                  &gbl_debug_omit_blob_write, INTERNAL, NULL, NULL, NULL,
                  NULL);
+REGISTER_TUNABLE("debug.allow_constraint_violation",
+                 "Deliberately allow insertion without constraint chec to debug db_verify", TUNABLE_BOOLEAN,
+                 &gbl_debug_allow_constraint_violation, INTERNAL, NULL, NULL, NULL,
```
mistachkin: typo here?  'chec' should be 'check' ?

db/glue.c
```diff
@@ -247,6 +247,13 @@ void init_fake_ireq_auxdb(struct dbenv *dbenv, struct ireq *iq, int auxdb)
     iq->use_handle = get_bdb_handle_ireq(iq, auxdb);
 }
 
+struct ireq *get_fake_ireq()
+{
+    struct ireq *iq = malloc(sizeof(struct ireq));
+    init_fake_ireq(thedb, iq);
```
mistachkin: for extra safety, calloc() instead?

adizaimi: sure

db/db_tunables.h
adizaimi: yep

mistachkin: I was just thinking that maybe the name of this tunable should more closely reflect what it actually does... Perhaps something like "debug.skip_constraints_on_insert"...  I don't know.

adizaimi: sure...will do

db/glue.c
akshatsikarwar: eh - `init_fake_ireq` is a constructor and already bzeros

mistachkin: It does appear to bzero portions of the ireq; however, since it does not bzero the entire thing, it's difficult to verify at a glance that the struct will be in a well-defined initial state upon return.

adizaimi:
Thank you Mike and Joe.

PR #1836, opened 2019-10-02
Add tunable for file database file permissions.

mponomar:


db/db_tunables.c
```diff
@@ -869,6 +871,19 @@ static int page_order_table_scan_update(void *context, void *value)
...
+static void* file_permissions_value(void *context) {
+    static char val[15];
+    snprintf(val, sizeof(val), "0%o", gbl_file_permissions);
+    return val;
+}
+
+static int file_permissions_update(void *context, void *value) {
+    (void)context;
+    char *in = (char*) value;
+    gbl_file_permissions = strtol(in, NULL, 8);
```
nirbhayc: I think we should validate (errno, valid permission) the input before updating the global variable. Also log a line about this change.

db/db_tunables.h
```diff
@@ -1839,4 +1839,8 @@ REGISTER_TUNABLE("strict_double_quotes",
 REGISTER_TUNABLE("eventlog_nkeep", "Keep this many eventlog files (Default: 2)",
                  TUNABLE_INTEGER, &eventlog_nkeep, 0, NULL, NULL, NULL, NULL);
 
+REGISTER_TUNABLE("file_perms", "Default filesystem permissions for database files",
+                 TUNABLE_STRING, NULL, 0, file_permissions_value,
+                 NULL, file_permissions_update, NULL);
```
nirbhayc: Make it read-only? I'd think changing it while db's running is a great idea (or serves any purpose).

tests/fileperms.test/runit
```diff
@@ -0,0 +1,21 @@
...
+${CDB2SQL_EXE} ${CDB2_OPTIONS} $db 'create table testtable1(a int)'
+
+ls -l ${DBDIR}/testtable1* | cut -d' ' -f1 | grep -v -- '-rw-rw---'
+if [[ $? -ne 1 ]]; then 
+    echo "Incorrect perms, expected -rw-rw---, got:" 
+    ls -l ${DBDIR}/testtable1*
+    exit 1
+fi
+
+${CDB2SQL_EXE} ${CDB2_OPTIONS} $db 'put tunable file_perms 0664'
```
nirbhayc: If we decide to make it readonly, we will have to move it to the lrl.

tests/tools/CMakeLists.txt
```diff
@@ -31,7 +31,6 @@ add_exe(hatest hatest.c)
 add_exe(cldeadlock cldeadlock.c)
 add_exe(insert_lots_mt insert_lots_mt.cpp)
 add_exe(leakcheck leakcheck.c)
-add_exe(localrep localrep.c strbuf.c)
```
nirbhayc: Why?

PR #1837, opened 2019-10-02
Seqnum fix

markhannum:
We report seqnum's back to the master from multiple threads.  This permits races where seqnum's will arrive on the master out-of-order.  This causes trouble for the distributed commit phase of a transaction if a seqnum issued by the replication thread is subsequently clobbered by a lower seqnum issued by the watcher thread: the committing transaction which hasn't seen the higher seqnum is forced to wait until the next periodic seqnum update.

The effect is that we periodically wait (needlessly) an extra half-second for a write to commit.  The fix is to only copy "larger" seqnums when the seqnum's generation is available.

### DIFF
--- a/bdb/rep.c
+++ b/bdb/rep.c
@@ -2359,6 +2359,43 @@ int verify_master_leases(bdb_state_type *bdb_state, const char *func,
 int gbl_catchup_window_trace = 0;
 extern int gbl_set_seqnum_trace;
 
+static inline int copy_seqnum(bdb_state_type *bdb_state, int seqnum_generations, 
+        seqnum_type *seqnum, int node_ix)
+{
+    int trace = bdb_state->attr->wait_for_seqnum_trace, now;
+    static int lastpr = 0;
+    if (!seqnum_generations) {
+        return 1;
+    }
+
+    int last_generation = bdb_state->seqnum_info->seqnums[node_ix].generation;
+    if (seqnum->generation > last_generation) {
+        return 1;
+    }
+
+    if (seqnum->generation < last_generation) {
+        if (trace && (now = time(NULL)) > lastpr) {
+            logmsg(LOGMSG_USER, "seqnum-generation %d < last_generation %d, not"
+                    "copying\n", seqnum->generation, last_generation);
+            lastpr = now;
+        }
+        return 0;
+    }
+
+    DB_LSN last_lsn = bdb_state->seqnum_info->seqnums[node_ix].lsn;
+    if (log_compare(&last_lsn, &seqnum->lsn) > 0) {
+        if (trace && (now = time(NULL)) > lastpr) {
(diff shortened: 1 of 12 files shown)


bdb/rep.c
```diff
@@ -2397,16 +2440,29 @@ static void got_new_seqnum_from_node(bdb_state_type *bdb_state,
...
+                       "%s: rejecting seqnum from %s because gen is "
+                       "%u (low), i want %u, count=%llu\n",
+                       __func__, host, seqnum->generation, mygen, count);
                 lastpr = now;
             }
             return;
         }
 
         if (seqnum->generation > mygen) {
+            static unsigned long long count = 0;
```
adizaimi: To avoid needless statics, can this new section move above, at which point the diff would be to change the above if:
if (seqnum->generation != mygen) {...(fix low high)...}

bdb/rep.c
```diff
@@ -2397,16 +2440,29 @@ static void got_new_seqnum_from_node(bdb_state_type *bdb_state,
             static unsigned long long count = 0;
             count++;
```
adizaimi: We may want to atomic_inc(count);

bdb/rep.c
adizaimi: Nevermind, we return in the section above.

adizaimi: grr, i think the two can be merged : ) but use your judgement, sorry for noise.

bdb/rep.c
```diff
@@ -2359,6 +2359,48 @@ int verify_master_leases(bdb_state_type *bdb_state, const char *func,
 int gbl_catchup_window_trace = 0;
 extern int gbl_set_seqnum_trace;
 
+static inline int copy_seqnum(bdb_state_type *bdb_state, seqnum_type *seqnum,
```
adizaimi: Misnamed function? check_if_should_copy_seqnum() maybe?

bdb/rep.c
```diff
@@ -2359,6 +2359,48 @@ int verify_master_leases(bdb_state_type *bdb_state, const char *func,
...
+                              int node_ix)
+{
+    int trace = bdb_state->attr->wait_for_seqnum_trace, now;
+    int last_generation = bdb_state->seqnum_info->seqnums[node_ix].generation;
```
adizaimi: Pass bdb_state->seqnum_info->seqnums[node_ix] and you only need 2 parameters to this function.
In fact you can do the memcopy here at the end of the func before returning.

bdb/rep.c
markhannum: Doesn't have to be perfect- this is used only for wait_for_seqnum_trace flag

bdb/rep.c
markhannum: cool - yeah ..

adizaimi:
Looks good.

PR #1840, opened 2019-10-03
cdb2jdbc: Fix PreparedStatement.setNull()

riverszhang89:
Don't pass an SQL type. Just pass CDB2_INTEGER as a placeholder type for a NULL value.


riverszhang89:
Thank you Nirbhay!

PR #1843, opened 2019-10-04
Verify: Add verbose parameter, no extra eol in verbose mode

adizaimi:
Verify: Add verbose parameter, no extra eol in verbose mode

### DIFF
--- a/lua/syssp.c
+++ b/lua/syssp.c
@@ -205,17 +205,23 @@ static int db_comdb_analyze(Lua L) {
 }
 
 
+/* verify() can take up to 3 parameters (table, mode, verbose):
+ * when we check we look for lua_isstring(L, 1), 2 and 3
+ * when we fetch the string we look for lua_tostring(L, -3), -2, -1
+ */
 static int db_comdb_verify(Lua L) {
     SP sp = getsp(L);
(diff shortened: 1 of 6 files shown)


lua/syssp.c
```diff
@@ -228,8 +234,23 @@ static int db_comdb_verify(Lua L) {
         } else if (strcmp(m, "blobs") == 0) {
             mode = VERIFY_BLOBS;
             logmsg(LOGMSG_INFO, "Verify ONLY blobs for table %s\n", tblname);
-        } else
-            logmsg(LOGMSG_INFO, "Verify table %s\n", tblname);
+        } else if (strcmp(m, "serial") == 0) {
+            mode = VERIFY_SERIAL;
+            logmsg(LOGMSG_INFO, "Verify in serial mode table %s\n", tblname);
+        } else if (strcmp(m, "verbose") == 0) {
```
mponomar: May want to read all the args and apply flags?  Otherwise this allows "verbose", "verbose" for example.

adizaimi: Yes, this does allow "verbose", "verbose", however i wanted to be able to omit the second parameter (ie assume some default, which for now is "serial") thus the best think i could think of is to allow the second param to be "verbose". What did you have in mind exactly?

PR #1844, opened 2019-10-04
Fix DOHSQL memory leak

riverszhang89:
DOHSQL may decide to reuse a cached row, which is essentially an SQLite Mem object. DOHSQL should free malloc'd memory in the Mem object before reusing it.

(DRQS 141973376)

### DIFF
--- a/sqlite/src/vdbeaux.c
+++ b/sqlite/src/vdbeaux.c
@@ -5759,7 +5759,7 @@ Mem *sqlite3CloneResult(
   }else{
     *pSize -= memRowSize(pMem, nCols);
     for(i=0; i<nCols; ++i){
-      sqlite3_value_free_inplace(pMem + i);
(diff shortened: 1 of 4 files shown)


sqlite/src/vdbeaux.c
```diff
@@ -5758,6 +5758,9 @@ Mem *sqlite3CloneResult(
     if( !pMem ) return 0;
   }else{
     *pSize -= memRowSize(pMem, nCols);
+    for(i=0; i<nCols; ++i){
+      sqlite3_value_free_inplace(pMem + i);
```
mistachkin: For  consistency with the sqlite3CloneResultFree function, maybe this?

sqlite3_value_free_inplace(&pMem[i]);

riverszhang89: Okay

PR #1845, opened 2019-10-04
quick-fix for reorder index

markhannum:
Refuse to allow index reordering directly from the block processor.  This is still being actively researched, but I'm checking in a quick-fix to get the builds working again.


markhannum:
Adi is reworking this

PR #1846, opened 2019-10-04
Fix reorder indices logic to work with tags

adizaimi:
Simplify logic how reorder_idx is turned on by setting iq flag,
this way all calls will have it off, unless codepath from osql.


adizaimi:
Reviewed with Mark.

PR #1847, opened 2019-10-07
Always perform extended verify checking for tran_rows > 1

adizaimi:
Always perform extended verify checking for tran_rows > 1 and transaction level below read committed.

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -438,10 +438,8 @@ int gbl_enable_cache_internal_nodes = 1;
 int gbl_use_appsock_as_sqlthread = 0;
 int gbl_rep_process_txn_time = 0;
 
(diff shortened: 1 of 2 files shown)


adizaimi:
This does not seem to bring any performance benefit so probably will withdraw.

nirbhayc:
roboperf failures were due to "failed with rc 402 Client api should run query against a different node", it should perhaps retry on such errors.

nirbhayc:
huh.. I think cdb2api should have retried instead of push it to the client.

adizaimi:
I am not sure that there is a benefit to perform this check first time around.

adizaimi:
Closing this. There is no clear benefit from doing this.

PR #1850, opened 2019-10-08
Use dynamic array for osql

adizaimi:
Use dynamic array to store osql rows.


adizaimi:
Similar to #1520 which fails to run tests under roborivers.

adizaimi:
Btw, the lousy results above are because this is being run with debug-mode build.

PR #1851, opened 2019-10-08
oldblobs already contain arrays in them, should not be arrays

adizaimi:
oldblobs already contain arrays in them, should not be arrays


adizaimi:
Thank you Joe.

PR #1853, opened 2019-10-08
Read committed master

markhannum:
Replace rwlock (dbenv->ser_lk) with a mutex, condition variable, and a counter.  The original design assumed incorrectly that it was okay to release a read-lock from a thread which did not originally acquire it.


markhannum:
W H O A H

markhannum:
Bah failing robomark

PR #1854, opened 2019-10-08
make sure we don't overcount decimal in indexes

dorinhogea:
Compound indexes including decimal columns are not working due to a bug introduced early on while porting changes to R7.


dorinhogea:
Thank you!

PR #1859, opened 2019-10-10
Can't read from comdb2_sc_status system table

nirbhayc:
Problem: The system table implementation was getting stuck while trying to destroy an uninitialized condition variable.


nirbhayc:
Thanks @adizaimi !

PR #1863, opened 2019-10-15
Disable a SQLite assert that does not apply to Comdb2

mistachkin:


mistachkin:
The test failures seen are a red herring.  Merging.

PR #1876, opened 2019-10-22
Fix SQL mspace race condition in DOHSQL

riverszhang89:
An SQlite engine's mspace isn't thread-safe because it's supposed to be accessed by the engine only. Unfortunately DOHSQL breaks the assumption:
While the master engine is casting a Mem object which needs to reallocate zMalloc from its mspace, a child engine may decide to reuse or discard a cached Mem object and therefore free zMalloc in the Mem object which is also allocated from the master's mspace. They can race with each other, as shown below.

```
                                      +----------------+
                                      | Child Engine 1 |
                                      +----------------+
                                              |
                                       inner_row (dohsql.c)
                                              |
                                       add_row (dohsql.c)
                                        * add Mem M to queue
+-------------------+                         |
| Master SQL engine |                         |
+-------------------+                         |
        |                                     |
next_row (sqlinterfaces.c)                    |
        |                                     |
dohsql_dist_next_row (dohsql.c)               |
        |                                     |
 * grab M from Child Engine 1 ----------------+
        |                                     |
send_row (sqlinterfaces.c)                    |
        |                                     |
column_value(M) (vdbeapi.c)                   |
        |                                     |
apply affinity to M (vdbemem.c)               |
 * e.g., sqlite3VdbeMemStringify              |
        |                                     |
sqlite3Malloc (malloc.c)                      |       +----------------+
        |                                     |       | Child Engine 2 |
mspace_malloc (dlmalloc.c)                    |       +----------------+
        |                                     |               |
        |                                     |        inner_row (dohsql.c)
next_row (sqlinterfaces.c)                    |               |
        |                                     |        add_row (dohsql.c)
dohsql_dist_next_row (dohsql.c)               |         * add Mem N to queue
 * Return M to Child Engine 1 ----------------+               |
 * grab N from Child Engine 2 ----------------)---------------+
        |                                     |
        |                              inner_row (dohsql.c)
        |                                     |
        |                              trimQue (dohsql.c)
send_row (sqlinterfaces.c)                    |
        |                              sqlite3CloneResultFree(M) (vdbeaux.c)
column_value(N) (vdbeapi.c)                   |
        |                                     |
apply affinity to N (vdbemem.c)               |
 * e.g., sqlite3VdbeMemStringify              |
        |                                     |
sqlite3Malloc (malloc.c)               sqlite3_free (malloc.c)
        |                                     |
mspace_malloc (dlmalloc.c)             mspace_free (dlmalloc.c)
                         \            /
! Race condition: Non thread-safe mspace shared between 2 engines
```

An obvious fix to this is to make an SQL mspace thread-safe. However this would incur unnecessary locking overhead for non-parallelizable queries. Instead we use a mutex to guard the Mem objects. We just need to ensure that we always hold the mutex whenever casting, reusing or discarding a Mem object inside the DOHSQL subsystem.

(DRQS 150091313)

### DIFF
--- a/db/dohsql.c
+++ b/db/dohsql.c
@@ -387,7 +387,7 @@ static int inner_row(struct sqlclntstate *clnt, struct response_data *resp,
     oldrow = queue_next(conn->que_free);
     if (oldrow && gbl_dohsql_verbose)
         logmsg(LOGMSG_DEBUG, "%lx %s retrieved older row\n", pthread_self(),
-                __func__);
+               __func__);
     Pthread_mutex_unlock(&conn->mtx);
 
     if (oldrow)
--- a/db/dohsql.c
+++ b/db/dohsql.c
@@ -126,6 +126,20 @@ pthread_mutex_t dohsql_stats_mtx = PTHREAD_MUTEX_INITIALIZER;
 dohsql_stats_t gbl_dohsql_stats;       /* updated only on request completion */
 dohsql_stats_t gbl_dohsql_stats_dirty; /* updated dynamically, unlocked */
 
+/* An SQlite engine's mspace isn't thread-safe because it's supposed to
+   be accessed by the engine only. DOHSQL breaks the assumption:
+   While the master engine is casting a Mem object which needs to reallocate
+   zMalloc from its mspace, a child engine may decide to reuse or discard a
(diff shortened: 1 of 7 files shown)


riverszhang89:

PR #1877, opened 2019-10-22
Fix DOHAST memory leak

riverszhang89:
Also fix log spew from fdb (discovered under the same ticket).


riverszhang89:

PR #1878, opened 2019-10-22
Check for cur->db before dereferencing it

adizaimi:
Check that cur->db is not NULL before accessing its members.


adizaimi:
Thank you Joe.

PR #1881, opened 2019-10-23
Remove mpreg

mponomar:
Trying to get rid of 2 bottlenecks.
1) We scan a list of mpreg routines while holding the buffer pool lock.  This list is small, but lots of things are contending for the lock.
2) We scan a list of all open files while holding the buffer pool lock to find a file descriptor we can use for writing.   Here we can deviate from upstream - we never have a shared buffer pool, so we can just store a list of __db_mpoolfiles per __mpoolfile.

### DIFF
--- a/berkdb/mp/mp_bh.c
+++ b/berkdb/mp/mp_bh.c
@@ -1164,7 +1164,7 @@ __dir_pg(dbmfp, pgno, buf, is_pgin)
 		bb_memp_pg_hit(start_time_us);
 	return (0);
 
-err:	MUTEX_THREAD_UNLOCK(dbenv, dbmp->mutexp);
+err:
 	__db_err(dbenv, "%s: %s failed for page %lu",
 	    __memp_fn(dbmfp), is_pgin ? "pgin" : "pgout", (u_long) pgno);
 	if (gbl_bb_berkdb_enable_memp_pg_timing)
--- a/berkdb/mp/mp_bh.c
+++ b/berkdb/mp/mp_bh.c
@@ -1143,6 +1143,9 @@ __dir_pg(dbmfp, pgno, buf, is_pgin)
 		start_time_us = bb_berkdb_fasttime();
 
 	ftype = mfp->ftype;
+    if (ftype == DB_FTYPE_SET)
+        ftype =  DB_UNKNOWN;
+
     if (mfp->pgcookie_len == 0)
         dbtp = NULL;
     else {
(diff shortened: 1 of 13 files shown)


berkdb/mp/mp_bh.c
```diff
@@ -1173,41 +1142,29 @@ __dir_pg(dbmfp, pgno, buf, is_pgin)
...
-		MUTEX_THREAD_UNLOCK(dbenv, dbmp->mutexp);
+    if (mfp->pgcookie_len == 0)
+        dbtp = NULL;
+    else {
+        dbt.size = mfp->pgcookie_len;
+        dbt.data = R_ADDR(dbmp->reginfo, mfp->pgcookie_off);
+        dbtp = &dbt;
+    }
+    if (is_pgin) {
+        if (dbmp->dbenv->pgin[mfp->ftype] != NULL &&
```
nirbhayc: Can ftype be something other than DB_BTREE? If not, turn this into an assert?

mponomar: Thanks Nirbhay.  This is actually badly broken (not in this code path) because ftype can be DB_FTYPE_SET (-1).  This still needs some work.  On the bright side, seeing visible performance improvement just from getting rid of this one extra lock in __dir_pg.

berkdb/mp/mp_register.c
```diff
@@ -64,40 +64,7 @@ __memp_register(dbenv, ftype, pgin, pgout)
...
-
-	mpreg->ftype = ftype;
-	mpreg->pgin = pgin;
-	mpreg->pgout = pgout;
-
-	MUTEX_THREAD_LOCK(dbenv, dbmp->mutexp);
-	LIST_INSERT_HEAD(&dbmp->dbregq, mpreg, q);
-	MUTEX_THREAD_UNLOCK(dbenv, dbmp->mutexp);
-
+    dbenv->pgin[ftype] = pgin;
```
nirbhayc: assert(ftype >= 0) ?

nirbhayc:
https://github.com/bloomberg/comdb2/blob/8cc301782ff8d4a85c0fb42d36f951d5199b90e0/berkdb/db/db.c#L622-L624

IIUC, this would result in ```dbenv->pgin[-1] = pgin```

riverszhang89:
spaces -> tabs

adizaimi:
Looks good to merge.

PR #1883, opened 2019-10-23
Fix a mutex/memory leak if sqlite3_step() fails in create_temp_table() (backport).

mistachkin:


lua/sp.c
```diff
@@ -1116,7 +1116,7 @@ static int create_temp_table(Lua lua, pthread_mutex_t **lk, const char **name)
         return luaL_error(sp->lua, sqlite3ErrStr(SQLITE_SCHEMA));
     }
     // now, actually create the temp table
-    *lk = malloc(sizeof(pthread_mutex_t));
```
mponomar: Why calloc a mutex we're about to initialize?

akshatsikarwar:
> Don't see the need for calloc instead of malloc. Otherwise looks correct.

Agreed.

lua/sp.c
mistachkin: Fixed.

PR #1884, opened 2019-10-23
Fix fingerprinting and its associated accounting for SQL queries executed inside Lua stored procedures (backport of #1682, #1703, #1708, #1965, et al).

mistachkin:
These changes are a backport of changes related to fingerprint handling, primarily from within the Lua stored-procedure subsystem.  Several related groups of changes have been merged into this PR.  

Furthermore, some of the semi-tangential changes (e.g. to the SQLite parser) are not strictly necessary (and could thus be removed); however, since their impact is minimal, well-defined, well-understood, and well-tested, it may be useful to retain them to minimize the delta between 7.0 and 7.x.

/plugin-branch 7.0


sqlite/src/parse.y
```diff
@@ -2340,15 +2351,22 @@ cmd ::= createkw PROCEDURE nm(N) NOSQL(X). {
...
     comdb2CreateProcedure(pParse, &N, &V, &X);
+}
 
 /////////////////////////////// DROP PROCEDURE ////////////////////////////////
-}
+
 cmd ::= DROP PROCEDURE nm(N) INTEGER(V). {
     comdb2DropProcedure(pParse, &N, &V, 0);
 }
+cmd ::= DROP PROCEDURE nm(N) VERSION INTEGER(V). {
```
mponomar, line 2361: This doesn't seem related?

PR #1886, opened 2019-10-24
have comdb2_sc_status table to show seed of running SC

adizaimi:
Have comdb2_sc_status table to show seed of running SC and few smaller changes:
* status_out does not need to be an array of pointers, we can simply have it be an array 
  (however sc_data is still an array of pointers)
* sc var does not need to be inited because we only use it to read in a few of the members from buf
* for `@send llmeta list` print all content of type LLMETA_SC_SEED
* refactor fetch_sc_seed() to make it reusable
* select * from comdb2_sc_status will call fetch_sc_seed() if status is running, ex:
`select name,seed from comdb2_sc_status
(name='t2', seed='0xf53400000000')
`

Unfortunately LLMETA_SC_SEED content gets removed once SC is done, so comdb2_sc_status  will not have a seed for non-running SC.

### DIFF
--- a/bdb/bdb_api.h
+++ b/bdb/bdb_api.h
@@ -1511,7 +1511,7 @@ int bdb_set_schema_change_status(tran_type *input_trans, const char *db_name,
                                  size_t schema_change_data_len, int status,
                                  const char *errstr, int *bdberr);
 
-int bdb_llmeta_get_all_sc_status(llmeta_sc_status_data ***status_out,
+int bdb_llmeta_get_all_sc_status(llmeta_sc_status_data **status_out,
                                  void ***sc_data_out, int *num, int *bdberr);
 
 int bdb_set_high_genid(tran_type *input_trans, const char *db_name,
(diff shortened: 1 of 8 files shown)


nirbhayc:
Please also update the docs: https://bloomberg.github.io/comdb2/system_tables.html#comdb2_sc_status

adizaimi:
Have comdb2_sc_status table to show seed of running SC and few smaller changes:
* status_out does not need to be an array of pointers, we can simply have it be an array 
  (however sc_data is still an array of pointers)
* sc var does not need to be inited because we only use it to read in a few of the members from buf
* for `@send llmeta list` print all content of type LLMETA_SC_SEED
* refactor fetch_sc_seed() to make it reusable
* select * from comdb2_sc_status will call fetch_sc_seed() if status is running, ex:
     select name,seed from comdb2_sc_status
     (name='t2', seed='0xf53400000000')

Unfortunately LLMETA_SC_SEED content gets removed once SC is done, so comdb2_sc_status  will not have a seed for non-running SC.

PR #1890, opened 2019-10-25
Make sqlite PRNG context thread-local

nirbhayc:
Fix provided by @akshatsikarwar

### DIFF
--- a/sqlite/src/random.c
+++ b/sqlite/src/random.c
@@ -21,7 +21,7 @@
 /* All threads share a single random number generator.
 ** This structure is the current state of the generator.
 */
-static SQLITE_WSD struct sqlite3PrngType {
+static __thread SQLITE_WSD struct sqlite3PrngType {
   unsigned char isInit;          /* True if initialized */
   unsigned char i, j;            /* State variables */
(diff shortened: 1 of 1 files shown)


dorinhogea:
testcase?

sqlite/src/random.c
```diff
@@ -21,7 +21,7 @@
 /* All threads share a single random number generator.
 ** This structure is the current state of the generator.
 */
-static SQLITE_WSD struct sqlite3PrngType {
+static __thread SQLITE_WSD struct sqlite3PrngType {
   unsigned char isInit;          /* True if initialized */
```
mistachkin: To avoid future merge conflicts, can we wrap this in the normal #if blocks for SQLite customizations?

mponomar: I'm assuming upstream SQLite doesn't have this bug (due to its locking vs ours?)

mistachkin: Right, canonical SQLite is immune to this issue.

nirbhayc:
Cherry-picked to 7.0.

PR #1891, opened 2019-10-27
Fix warnings when compiling for arm architecture

adizaimi:
Fix warnings when compiling for arm architecture


tests/tools/get_tests_inorder.sh
```diff
@@ -18,4 +18,4 @@ generated=$(for j in `ls */*.testopts` ; do
     echo $gen $time
 done )
 
-echo -e "${custom_times[*]} \n${default_times[*]} \n${generated[*]}" | awk '{print $2,$1}'| sort -n | awk '{print $2}'
+echo -e "${custom_times[*]} \n${default_times[*]} \n${generated[*]}" | awk '{print $2,$1}'| sort -nr | awk '{print $2}'
```
akshatsikarwar: Don't reverse, or yes reverse?

berkdb/lock/lock_deadlock.c
```diff
@@ -1363,7 +1363,7 @@ retry:	count = region->stat.st_nlockers;
 				logmsg(LOGMSG_USER, "Marking valid waiter after increment lip %p id=%x dd_id=%x count=%u master=%x\n",
 					lockerp, lockerp->id, lockerp->dd_id,
 					dd_id_array[lockerp->dd_id].count,
-					(has_master)? lockerp->master_locker : 0);
+					(unsigned int)((has_master)? lockerp->master_locker : 0));
```
akshatsikarwar: `master_locker` is `intptr_t` -- lets use `PRIiPTR`/`PRIxPTR` to print instead of casting to `unsigned`

tests/tools/get_tests_inorder.sh
adizaimi: Sorry for confusion--we don't want to change this from master, i had to change this to run the tests in order from least expensive to most expensive so to finish the quicker tests first. The slower tests do timeout or run out of space more so I know i only will be able to run a portion of them. So far I am at 220/330 tests successful and about 10 out of space or out of time. One actually still gets a segmentation fault--but that will be a different PR.

berkdb/lock/lock_deadlock.c
adizaimi: I can try that.

berkdb/os/os_rw.c
```diff
@@ -725,8 +725,8 @@ __os_physwrite(dbenv, fhp, addr, len, nwp)
 				else {
 					/* We got positioned into the limbo somewhere (how?)
 					 * so it's not safe to proceed. */
-					__db_err(dbenv, "lseek: fd %d off %lu, wanted %lu\n", fhp->fd,
-							rc, off);
+					__db_err(dbenv, "lseek: fd %d off %u, wanted %u\n", fhp->fd,
+							(unsigned int)rc, (unsigned int)off);
```
akshatsikarwar: `off_t` is signed value. cast to int64_t and print using pri64d?

adizaimi: yep, thats good suggestion. thanks.

adizaimi:
Thank you Akshat.

PR #1892, opened 2019-10-28
Bthash fix

markhannum:
Fix bug in bam_split_recover handle where the dbp peer handle references free memory.  This bug is reproducible by checking out 2638790f47fa06a6e1f50d706fd266bc22680967, compiling with -DBTHASH_REPRODUCE_BUG, and running the sc_truncate test in clustered mode.  The fix is to fill "backpointer" references to peers, and to zero these when the reference is no longer valid (because it is closed).

### DIFF
--- a/berkdb/db/db.c
+++ b/berkdb/db/db.c
@@ -537,6 +537,9 @@ __db_dbenv_setup(dbp, txn, fname, id, flags)
 						sizeof(DB *));
 				dbp->revpeer[dbp->revpeer_count-1] = lldbp;
 			} else {
+#if defined BTHASH_REPRODUCE_BUG
+				dbp->peer = lldbp->peer;
+#else
 				if (lldbp->peer && F_ISSET(lldbp->peer, DB_AM_HASH)) {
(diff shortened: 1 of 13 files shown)


berkdb/db/db.c
```diff
@@ -770,17 +785,37 @@ __db_close(dbp, txn, flags)
 	 */
 	MUTEX_THREAD_LOCK(dbenv, dbenv->dblist_mutexp);
 	db_ref = --dbenv->db_ref;
+
```
markhannum: Add a line of comment ..

db/toblock.c
```diff
@@ -2051,6 +2051,10 @@ osql_create_transaction(struct javasp_trans_state *javasp_trans_handle,
                 *trans = bdb_get_physical_tran(iq->sc_logical_tran);
                 irc = create_child_transaction(iq, *trans, &(iq->sc_tran));
             }
```
markhannum: This looks unnecessary .. irc is checked below ..

PR #1895, opened 2019-10-29
fix ibm XCHANGE64 function called, and few warnings

adizaimi:
fix ibm XCHANGE64 function called, and few warnings


akshatsikarwar:
Thanks

PR #1896, opened 2019-10-29
Fix use of comdb2_rowid in WHERE clauses

mistachkin:
These changes fix the refactored sqlite3BtreeGetGenId() function to return the actual string length instead of the allocated string length for the string rowId.  This is necessary for VDBE comparison operations to work properly.


mistachkin:
Some of the extraneous code you mention is related to functionality that has since been removed (e.g. via your commit [5cbe331be31a67a408dab3467bb6dc22173f6f69]).  Do you want me to refactor the code in this pull request or a subsequent one?

mponomar:
There's no need for any refactoring.  There's a lot of asserts that aren't needed.  There's 2 callers, so removing a parameter and deleting asserts would be a quick win.  sqlite3_mprintf seems cleaner than malloc+snprintf, but that's your call.

mponomar:
If you want to get your changes in as is, and then do another PR for cleaning up the routine (or not), that's fine.

mistachkin:
The two remaining RoboRivers test failures are false alarms.

PR #1897, opened 2019-10-29
Fix use of comdb2_rowid in WHERE clauses (backport)

mistachkin:
These changes fix the refactored sqlite3BtreeGetGenId() function to return the actual string length instead of the allocated string length for the string rowId. This is necessary for VDBE comparison operations to work properly.


mistachkin:
The only remaining test failure is a false-alarm.

PR #1900, opened 2019-10-30
Update SQLite to the latest 3.31 branch code...

mistachkin:
These changes are tracking the latest SQLite trunk, which will eventually become the 3.31 release.


nirbhayc:
Could you also merge VERSION file from upstream? will make it easier to get the current checked-in sqlite version.

mistachkin:
> 
> 
> Could you also merge VERSION file from upstream? will make it easier to get the current checked-in sqlite version.

Great idea.  Should be done everywhere now.

PR #1901, opened 2019-10-30
sql_tranlevel: Change level only for valid modes

akshatsikarwar:


nirbhayc:
One of my checkins broke tunables test; should be fixed now.

PR #1906, opened 2019-11-06
Nullify the OUT pointer before passing as arg

nirbhayc:
DRQS 151208736


adizaimi:
Where can contain garbage *only* if rc = -1, in which case we don't strdup. 
Seems setting to 0 does not fix any issue.

nirbhayc:
yep, don't know what I was thinking.. :)

PR #1907, opened 2019-11-07
phys replicant: remove dup in tunables and tiny cleanup

adizaimi:
phys replicant: remove dup in tunables and tiny cleanup


adizaimi:
After talking to Mark, it seems that phys_rep.test is a bit unreliable that's why he did not enable it, and also that truncatesc.test tests a bit the physical replicant feature (I believe that's the one he mentioned).
I also tried running manually phys_rep.test and passed so checking this in.

PR #1909, opened 2019-11-07
LONGREQ: Print bound param values

akshatsikarwar:


adizaimi:
We build logger->bound_param_cson in eventlog.c and populate it in bind_parameters(). 
It would be cool if you could just use that array and print it instead of this work (which is quite elegant otherwise).

PR #1910, opened 2019-11-09
Tests: move common functions in runit_common.sh

adizaimi:
Tests: move common functions in runit_common.sh


adizaimi:
Thank you Nirbhay.

PR #1912, opened 2019-11-10
Maximum row buffer size

riverszhang89:
Only keep a reasonable amount of memory for the row buffer in a clnt.
If the size of the row buffer exceeds a threshold, don't keep it.

(DRQS 150673741)


riverszhang89:

PR #1914, opened 2019-11-11
Add checksum (md5) function.

mponomar:
Found this useful for my own db - function checksums the input string.  If input is not a string, returns NULL.

### DIFF
--- a/sqlite/src/func.c
+++ b/sqlite/src/func.c
@@ -982,12 +982,26 @@ static void md5Func(
     MD5Context ctx = {0};
     char checksum[16];
     char out[33];
+    unsigned int len;
+    int type = sqlite3_value_type(argv[0]);
 
-    if (sqlite3_value_type(argv[0]) != SQLITE_TEXT)
-        return;
-    s = sqlite3_value_text(argv[0]);
+    switch (type) {
+        case SQLITE_TEXT:
+            s = sqlite3_value_text(argv[0]);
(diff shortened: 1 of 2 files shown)


sqlite/src/func.c
```diff
@@ -24,6 +24,8 @@
 #include "comdb2.h"
 #include "sql.h"
 #include "bdb_int.h"
+#include "md5.h"
```
nirbhayc: Enclose the changes in WITH_SSL ?

mponomar: This is SQLite's MD5 implementation, not openssl's.  We use it in other places (eg: fingerprint code), even without WITH_SSL.

nirbhayc: oh ok. I looked at the signature thought they were openssl's.

sqlite/src/func.c
```diff
@@ -971,6 +973,26 @@ static void comdb2StartTimeFunc(
...
+  sqlite3_context *context,
+  int argc,
+  sqlite3_value **argv
+  ){
+    const unsigned char *s;
+    MD5Context ctx = {0};
+    char checksum[16];
+    char out[33];
+
+    if (sqlite3_value_type(argv[0]) != SQLITE_TEXT)
```
mistachkin: It would also be nice to be able to perform the md5sum() function on a blob value.

PR #1915, opened 2019-11-12
Legacy behavior: For compound SELECT column names were picked up from…

nirbhayc:
… the command instead of schema


sqlite/src/select.c
```diff
@@ -1934,7 +1934,12 @@ static void generateColumnNames(
       /* An AS clause always takes first priority */
       char *zName = pEList->a[i].zName;
       sqlite3VdbeSetColName(v, i, COLNAME_NAME, zName, SQLITE_TRANSIENT);
+#if !defined(SQLITE_BUILDING_FOR_COMDB2)
     }else if( srcName && p->op==TK_COLUMN ){
+#else /* defined(SQLITE_BUILDING_FOR_COMDB2) */
```
mistachkin: For this "#else", the "!" character is missing in front of defined(SQLITE_BUILDING_FOR_COMDB2), i.e. to match the associated "#if".

PR #1916, opened 2019-11-12
Allow space in key names (for backward compatibility)

mponomar:


adizaimi:
Can we turn this on only for legacy defaults, or at least on a tunable that is off by default for Open? Then the test can check allow/not depending on the state of the tunable (set .testopts then check like the other tests do ex. tests/reorder_deletes.test/runit: `if [[ $DBNAME == *"noreordergenerated"* ]] ; then...`).

mponomar:
We can.  I don't know of a good reason to block them though.   Do you know of them breaking anything specific?  SQLite, Postgres, MySQL allow them.

PR #1917, opened 2019-11-12
Add experimental SQL function to return qualified table names referenced by an arbitrary SELECT statement

mistachkin:
These changes introduce a new SQL function (tentatively named "table_names") that returns the qualified table names referenced by an arbitrary SELECT statement.

There is one additional wrinkle:

**The referenced database, table, and column names need not actually exist.**

In order to meet this requirement, the prepare-only mode functionality has been enhanced to support short-circuiting the parsing process immediately after the list of qualified table names is made available.


mistachkin:
NOTE: Last remaining RoboRivers test failure is a false alarm.

PR #1918, opened 2019-11-13
On start, only delete files with 'csc2' extension

nirbhayc:
DRQS 151647504

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -1333,8 +1333,7 @@ static void *purge_old_files_thread(void *arg)
     return NULL;
 }
 
-/* remove every file that contains ".csc2" anywhere in its name.
-   this should be safe */
+/* Remove all csc2 files */
 static int clear_csc2_files(void)
 {
     char path[PATH_MAX] = {0};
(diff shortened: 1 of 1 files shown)


akshatsikarwar:
Does putting it in squiggly make it a link? {DRQS 151647504}

adizaimi:
Ok, strrchr will find the last occurence of '.' so we will delete only things that have ".csc2xxx" in them. We might want to make sure that there is no xxx after .csc2 just for completeness.

db/comdb2.c
```diff
@@ -1356,9 +1355,9 @@ static int clear_csc2_files(void)
             if (!strcmp(dp->d_name, ".") || !strcmp(dp->d_name, ".."))
                 continue;
 
-            ptr = strstr(dp->d_name, ".csc2");
+            ptr = strrchr(dp->d_name, '.');
 
-            if (ptr) {
+            if (ptr && (strncmp(ptr, ".csc2", sizeof(".csc2")) == 0)) {
```
adizaimi: sizeof(".csc2") is 6 (includes '\0') so strncmp will compare 6 characters, so will make sure that ptr[5] == '\n', cool.

PR #1920, opened 2019-11-14
Fix smoke test for CentOS 8

riverszhang89:
`protobuf-c-devel` is moved to the PowerTools repository on CentOS 8. Update the docker file accordingly.


riverszhang89:
@akshatsikarwar Thank you!

PR #1922, opened 2019-11-18
Have macc_globals object on heap instead of global memory

adizaimi:
Have macc_globals object on heap instead of global memory. 
This structure groups variables used for csc2 parsing and is about 10MB in size--there is no need for it to be in global memory area.

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -2248,7 +2248,8 @@ static int llmeta_load_tables(struct dbenv *dbenv, char *dbname, void *tran)
             logmsg(LOGMSG_ERROR, "get_csc2_file failed %s:%d\n", __FILE__, __LINE__);
             break;
         }
-        printf("calling dyns_load_schema_string from %s\n", __func__);
+        printf("calling dyns_init_globals from %s\n", __func__);
+        dyns_init_globals();
         rc = dyns_load_schema_string(csc2text, dbname, tblnames[i]);
         free(csc2text);
         csc2text = NULL;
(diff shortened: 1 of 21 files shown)


adizaimi:
cdb2test   Nov 18 22:06:52 2019  in-comdb2db AZ_macc_globals_on_heap.R20191118.6

db/comdb2.c
```diff
@@ -3275,15 +3278,18 @@ static int init_sqlite_table(struct dbenv *dbenv, char *table)
        return -1;
     }
 
+    dyns_init_globals();
     rc = dyns_load_schema_string((char*) schema, dbenv->envname, table);
     if (rc) {
         logmsg(LOGMSG_ERROR, "Can't parse schema for %s\n", table);
-        return -1;
+        rc = -1;
+        goto err;
```
akshatsikarwar: I prefer if caller invoked `dyns_init_globals` and invokes `dyns_cleanup_globals` on non-zero return.

db/config.c
```diff
@@ -966,16 +966,19 @@ static int read_lrl_option(struct dbenv *dbenv, char *line,
                     return -1;
                 }
             }
+            dyns_init_globals();
             rc = dyns_load_schema(fname, (char *)gbl_dbname, tblname);
             if (rc != 0) {
                 logmsg(LOGMSG_ERROR, "Error loading %s schema.\n", tok);
+                dyns_cleanup_globals();
```
akshatsikarwar: Same - I prefer if we don't sprinkle error handling everywhere.

riverszhang89: Agreed.

db/comdb2.c
adizaimi: Well, in this case i tried to keep them as close as possible together that way it appeared as a pattern: dyns_init_globals, dyns_load_schema, dyns_cleanup_globals. 
Anyway, I did reorganize them as much as i could so now most callers will init and cleanup.
There are however places where we do get a lock and this globals init and cleanup needs to be done while the lock is held so I can't clean them all like you wished.

PR #1923, opened 2019-11-18
comdb2_tunables: present int/double tunable values as string

nirbhayc:
.

### DIFF
--- a/sqlite/ext/comdb2/tunables.c
+++ b/sqlite/ext/comdb2/tunables.c
@@ -154,16 +154,20 @@ static int systblTunablesColumn(sqlite3_vtab_cursor *cur, sqlite3_context *ctx,
         switch (tunable->type) {
         case TUNABLE_INTEGER: {
             int val;
+            char buffer[64];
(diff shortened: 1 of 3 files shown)


sqlite/ext/comdb2/tunables.c
```diff
@@ -150,16 +150,20 @@ static int systblTunablesColumn(sqlite3_vtab_cursor *cur, sqlite3_context *ctx,
...
+            sqlite3_result_text(ctx, strdup(buffer), -1, free);
             break;
         }
         case TUNABLE_DOUBLE: {
             double val;
+            char buffer[24];
             val = (tunable->value) ? *(double *)tunable->value(tunable)
                                    : *(double *)tunable->var;
-            sqlite3_result_double(ctx, val);
+            sqlite3_snprintf(sizeof(buffer), buffer, "%f", val);
```
mistachkin: It might be worthwhile to use a double format that is (more?) round-trippable, e.g. "%!.15g", etc.

mistachkin (approved):
Looks good to me.

PR #1925, opened 2019-11-19
Make the result of sqlite3_normalized_sql() survive its statement being reprepared (backport).

mistachkin:
Cherrypick of upstream SQLite trunk fix [4330f0795dbc2ab41].


mistachkin:
The only remaining RoboRivers test failure is a confirmed false alarm.

PR #1926, opened 2019-11-19
Before placing a prepared statement back into the cache, clear its parameter bindings.  Also, detect and log related errors.

mistachkin:


mistachkin:
The last remaining RoboRivers test failure is a confirmed false alarm.

PR #1929, opened 2019-11-19
Fix unportable usage of realpath

riverszhang89:
Some systems (e.g., AIX) do not like resolved_path being NULL: POSIX.1-2001 states that if resolved_name is a null pointer, the behavior of realpath is implementation-defined. So use realpath in a portable way.


riverszhang89:
@mponomar Thank you for the review!

PR #1930, opened 2019-11-19
Simpler and faster GENID48 generation

riverszhang89:
Store GENID48 components separately in `bdb_state`: the high 48 bits (auto-incrementing rowid) are stored in host byte order; the low 16 bits (stripe + update-id) are stored in network byte order. This avoids unnecessary byte-swapping and memcpy.


riverszhang89:

PR #1934, opened 2019-11-20
Provide a tunable to fallback to old column naming behavior

nirbhayc:
Column naming in SELECT result set has changed recently. Introduce a tunable to revert back to old-style names. The tunable is enabled by default. See checked-in tests for example queries that were affected.


mponomar:
Causes an assert in debug build:

```
create table t(a int)
select count(*), a from t group by a

comdb2: ../sqlite/src/select.c:1937: generateColumnNames: Assertion `p->op!=236' failed.
```

nirbhayc:
https://github.com/bloomberg/comdb2/pull/1997

PR #1935, opened 2019-11-21
Free the path returned by realpath in the create_db function.

mistachkin:
Since realpath() returns a malloced buffer that cannot escape the scope of the create_db() function, it should be freed prior to exiting.

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -3348,7 +3348,8 @@ static int create_db(char *dbname, char *dir) {
    logmsg(LOGMSG_INFO, "Creating db in %s\n", dir);
    setenv("COMDB2_DB_DIR", fulldir, 1);
 
-   if (init_db_dir(dbname, dir)) return -1;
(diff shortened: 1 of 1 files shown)


db/comdb2.c
```diff
@@ -3348,7 +3348,8 @@ static int create_db(char *dbname, char *dir) {
    logmsg(LOGMSG_INFO, "Creating db in %s\n", dir);
    setenv("COMDB2_DB_DIR", fulldir, 1);
 
-   if (init_db_dir(dbname, dir)) return -1;
+   if (init_db_dir(dbname, dir)) { free(fulldir); return -1; }
```
adizaimi: how about making this a bit more elegant:
rc = init_db_dir(dbname, dir);
free(fulldir);
if (rc) return -1;

mistachkin: Sure, I'll make that change.

PR #1936, opened 2019-11-21
Use consistent error handling and cleanup semantics for c_close and c_get (fix leak-on-error) in the queuedb subsystem.

mistachkin:
These changes fix a DBC resource leak in the bdb_queuedb_walk function when an unexpected error occurs.  Also, they make the c_close (infinite?) retry handling more consistent.


bdb/queuedb.c
```diff
@@ -642,10 +648,12 @@ int bdb_queuedb_consume(bdb_state_type *bdb_state, tran_type *tran,
         if (crc == DB_LOCK_DEADLOCK) {
             *bdberr = BDBERR_DEADLOCK;
             rc = -1;
+            goto done;
         } else if (crc) {
             logmsg(LOGMSG_ERROR, "%s: c_close berk rc %d\n", __func__, crc);
             *bdberr = BDBERR_MISC;
             rc = -1;
+            goto done;
```
adizaimi: This looks wrong to me -- looks like we are asking for trouble on error -- I think we should not goto done here. I understand that line 283 and 514 do the same thing, and I think they should not.

mistachkin: Initially, that was my perspective as well; however, Mike kinda convinced me that the goto (retry) mechanism is required for close operations...  The code was/is a bit confusing to me.  Clearly, this PR will require a wide audience to review it.

mponomar: I don't see a path (without CDB_LOCKING) where a cursor close can return DEADLOCK.  Unless I'm wrong, that entire code path is impossible.

mponomar: According to the docs (and a quick inspection of the code), we shouldn't be retrying on close, ever.  Mike was wrong.

mponomar: I still don't see how a deadlock can happen, outside of CDB_LOCKING, but it's no different from any other return code - the cursor is unusable.  If it's transactional, we'd need to abort the transaction, but it isn't.

mistachkin:
I'm still waiting for a clean test pass in order to get this merged.  I'm fairly confident that the RoboRivers issues are false alarms.  However, I have no idea what is wrong with the legacy tests at the moment.

mistachkin:
The final two legacy test failures appear to be false alarms as they are also failing on master.

PR #1938, opened 2019-11-21
two rqid in struct ireq, label the dup properly

dorinhogea:
trivial struct field name change, it was confusing to use two rqid-s in same ireq structure (one is in sorese for regular osql writes, the second is used for forwarded tag writes that piggyback in same osql net).


dorinhogea:
thanks

PR #1941, opened 2019-11-22
Show queue size for lua queues

mponomar:
queuesize

### DIFF
--- a/bdb/file.c
+++ b/bdb/file.c
@@ -7092,7 +7092,7 @@ uint64_t bdb_queuedb_size(bdb_state_type *bdb_state)
              bdb_state->name);
     int rc = stat(tmpname, &st);
     if (rc) {
-        fprintf(stderr, "stat(%s) rc %d\n", tmpname, rc);
+        logmsg(LOGMSG_ERROR, "stat(%s) rc %d\n", tmpname, rc);
         return 0;
     }
     return st.st_size;
--- a/bdb/file.c
+++ b/bdb/file.c
@@ -7084,6 +7084,19 @@ static size_t dirent_buf_size(const char *dir)
                                              : sizeof(struct dirent));
 }
 
+uint64_t bdb_queuedb_size(bdb_state_type *bdb_state) 
+{
+    char tmpname[PATH_MAX];
+    struct stat st;
+    snprintf(tmpname, sizeof(tmpname), "%s/%s.queuedb", bdb_state->dir, bdb_state->name);
+    int rc = stat(tmpname, &st);
+    if (rc) {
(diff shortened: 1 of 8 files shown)


bdb/file.c
```diff
@@ -7101,8 +7114,8 @@ uint64_t bdb_queue_size(bdb_state_type *bdb_state, unsigned *num_extents)
 
     *num_extents = 0;
 
-    if (bdb_state->bdbtype != BDBTYPE_QUEUE)
-        return 0;
+    if (bdb_state->bdbtype == BDBTYPE_QUEUEDB)
```
adizaimi: This piece of code assumes that bdbtype here can be only  BDBTYPE_QUEUEDB and BDBTYPE_QUEUE. I would put an assert here for that.
One instance of this func being called is glue.c:5419, and the if clause for that is wrong -- if you can please fix with this pr since these are related.

bdb/file.c
```diff
@@ -7084,6 +7084,19 @@ static size_t dirent_buf_size(const char *dir)
                                              : sizeof(struct dirent));
 }
 
+uint64_t bdb_queuedb_size(bdb_state_type *bdb_state) 
+{
+    char tmpname[PATH_MAX];
+    struct stat st;
+    snprintf(tmpname, sizeof(tmpname), "%s/%s.queuedb", bdb_state->dir, bdb_state->name);
```
adizaimi: Can we make this snprintf a generic function get_queue_filename() and use it from all the places that get the filename?

bdb/file.c
```diff
@@ -7084,6 +7084,19 @@ static size_t dirent_buf_size(const char *dir)
...
+    int rc = stat(tmpname, &st);
+    if (rc) {
+        fprintf(stderr, "stat(%s) rc %d\n", tmpname, rc);
```
adizaimi: logmsg(ERR....)?

bdb/file.c
mponomar: I'm more worried about the fact that calc_table_size and calc_table_size_analyze seem top do the same exact thing.

bdb/file.c
mponomar: This is the only such place, unless you count form_queuedb_name, which has a different set of rules (it forms a loggable name with our usual XXX. kludge).  It can be moved to a different function, sure.

adizaimi: If it is the only place, then leave it : )

bdb/file.c
mponomar: Ah, calc_table_size_analyze skips blob sizes, which makes sense.

mponomar: Added assert, removed copy pasta.

PR #1943, opened 2019-11-25
Fix repopulating a queuedb

riverszhang89:
This PR fixes handling of the `queuedb` directive. `qdbs` is referenced
but is never initialized. The PR properly allocates and resizes `qdbs`
on each `queuedb` configuration.

The last character of a queuedb config is dropped every time it's created
from a repopulated LRL file. The PR fixes this bug too.

(DRQS 152088323)


riverszhang89:

PR #1945, opened 2019-11-25
Use atomics for osql cancelall state instead of cancelall_mtx

adizaimi:
Use atomics for osql cancelall state instead of cancelall_mtx


adizaimi:
Thanks Joe.

PR #1946, opened 2019-11-25
Atomics for clients count instead of mutex

adizaimi:
Atomics for clients count instead of mutex

### DIFF
--- a/db/osqlsession.c
+++ b/db/osqlsession.c
@@ -28,6 +28,7 @@
 #include <net_types.h>
 
 #include "debug_switches.h"
+#include "comdb2_atomic.h"
 
 #include <uuid/uuid.h>
 #include "str0.h"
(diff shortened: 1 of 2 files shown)


db/osqlsession.c
```diff
@@ -206,17 +201,14 @@ inline unsigned long long osql_sess_getrqid(osql_sess_t *sess)
  */
 int osql_sess_addclient(osql_sess_t *sess)
 {
-    Pthread_mutex_lock(&sess->clients_mtx);
 #if 0
    uuidstr_t us;
    comdb2uuidstr(sess->uuid, us);
    fprintf(stderr, "\t\tADDCLNT p_sees=%p rqid=[%llx %s] sess->completed=%llx thread=%d clients=%d p_sess->iq=%p\n",
          sess, sess->rqid, us, sess->completed, pthread_self(), sess->clients+1, sess->iq);
 #endif
```
mistachkin: I realize this code is #if'd out; however, maybe consider using ATOMIC_LOAD here for the reading of sess->clients?

db/osqlsession.c
```diff
@@ -156,10 +153,8 @@ static void _destroy_session(osql_sess_t **prq, int phase)
     case 2:
         Pthread_mutex_destroy(&rq->mtx);
     case 3:
-        Pthread_mutex_destroy(&rq->clients_mtx);
-    case 4:
         Pthread_mutex_destroy(&rq->completed_lock);
-    case 5:
+    case 4:
         free(rq);
```
mistachkin: Would there be any value to atomically setting the rq->clients value to something obviously invalid here (e.g. -1, etc)?

adizaimi: We free rq in next line. I don't think setting client to -1 makes any difference.

db/osqlsession.c
adizaimi: 32 bit variable loading is atomic so in this case that would not be needed at all.

adizaimi:
Thanks for the review Joe.

PR #1947, opened 2019-11-25
fix origname troublemaker change

dorinhogea:
Origname is intended to help hide "new." prefix filename during schema change; the prefix is used to differentiate between the old file and the new file being generated.
When origname is present, even though an in-memory file is identified by "new.tblname", final on-disk version is "tblname", and origname is used to map the in-memory to on-disk filenames.
Unfortunately this breaks file renaming, when table is indexed (schema code changes?), and regression test-case did not cover the scenario.  Origname in this case end up replacing the intended name with the old name (that happens to be saved in origname).
Fix here.

### DIFF
--- a/bdb/file.c
+++ b/bdb/file.c
@@ -1013,6 +1013,7 @@ int bdb_rename_table(bdb_state_type *bdb_state, tran_type *tran, char *newname,
 {
     DB_TXN *tid = tran ? tran->tid : NULL;
     char *orig_name;
+    char *origname; /* certain sc set this, preserve */
     int rc;
 
     rc = close_dbs_flush(bdb_state, tid);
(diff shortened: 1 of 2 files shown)


bdb/file.c
```diff
@@ -1013,6 +1013,7 @@ int bdb_rename_table(bdb_state_type *bdb_state, tran_type *tran, char *newname,
 {
     DB_TXN *tid = tran ? tran->tid : NULL;
     char *orig_name;
+    char *origname; /* certain sc set this, preserve */
```
riverszhang89: I'd choose a different variable name. `origname` and `orig_name` are quite easily confused with each other.

akshatsikarwar: Agreed

dorinhogea:
thank you; the bounce for the test case will be integrated as the last step after each test-case, similar with the verify process, to make sure that the db can come up

PR #1949, opened 2019-11-26
Ckp delete log

markhannum:
Fix day-1 bug in bdb_checkpoint_list_ok_to_delete_log where newer checkpoints can prevent the deletion of older logfiles.


akshatsikarwar:
✓.

adizaimi:
I have noticed logs accumulating and growing bigger and not go away -- hopefully this is the reason and logs will finally go away. 
This seems quite bad -- not ever releasing that mutex in the else path -- thanks for fixing this.

PR #1951, opened 2019-11-26
remove dead code

dorinhogea:
some TRAN_FLAGS that are not in use


adizaimi:
Looks good. Since you are touching this, it seems iq->transflags is not used at all, only assigned.

dorinhogea:
thank you!

PR #1952, opened 2019-11-27
toblock.c reduce lock contention and other cleanup

adizaimi:
* sltdbt.c: only osql_blkseq_unregister in case of nonzero rc
* print blkseq add/rem info only for DEBUG build 
* only decrement with lock if totpen is nonzero 
* replace with atomic inc toblock stats for n_commits


adizaimi:
Thank you Dorin.

PR #1955, opened 2019-12-02
Rewrite check and do less under checkboard->rwlock

adizaimi:
* Check for master does not need to be done under checkboard->rwlock
* Do less work under checkboard->rwlock in _osql_register_sqlthr()
* remove unneeded parameter for osql_chkboard_sqlsession_exists()
* remove unused sess_blocksock member


db/osqlcheckboard.c
```diff
@@ -242,85 +237,81 @@ int osql_unregister_sqlthr(struct sqlclntstate *clnt)
...
+    Pthread_mutex_destroy(&entry->mtx);
+    free(entry);
 
-        /*reset rqid */
-        clnt->osql.rqid = 0;
-    }
+    /*reset rqid */
+    clnt->osql.rqid = 0;
 
     Pthread_rwlock_unlock(&checkboard->rwlock);
```
adizaimi: Maybe move this up after the hash_del.

db/osqlcheckboard.c
```diff
@@ -242,85 +237,81 @@ int osql_unregister_sqlthr(struct sqlclntstate *clnt)
...
-bool osql_chkboard_sqlsession_exists(unsigned long long rqid, uuid_t uuid,
-                                     bool lock)
+bool osql_chkboard_sqlsession_exists(unsigned long long rqid, uuid_t uuid)
 {
     if (!checkboard)
         return 0;
 
-    osql_sqlthr_t *entry = osql_chkboard_fetch_entry(rqid, uuid, lock);
+    osql_sqlthr_t *entry = osql_chkboard_fetch_entry(rqid, uuid, true);
     return entry != NULL;
```
adizaimi: inline the function call in the return stmt.

db/osqlcheckboard.c
adizaimi: Actually there is a subsequent pr that will move this up.

adizaimi:
Thank you Dorin and Joe.

PR #1960, opened 2019-12-03
remove wait sql code

dorinhogea:
Sql transactions are dispatched to worker thread only after the completion message, therefore the whole bplog is available on the master. Remove the code waiting for the bplog to arrive (which was mainly used in defunct blocksql).


dorinhogea:
thank you

dorinhogea:
thank you

PR #1962, opened 2019-12-04
Delete current stats only if backup is available

mohitkhullar:


adizaimi:
I would put a testcase for this to the existing analyze testcase.
Thanks

PR #1965, opened 2019-12-05
Track fingerprints for Lua SQL queries where a statement is explicitly closed

mistachkin:
These changes fix an issue in the SQL query fingerprinting where a query would not be added to the virtual table when a DB statement is closed in Lua without being fully consumed first.


mistachkin:
The final test failure (legacy 'querylimits') appears to be a false alarm.

PR #1967, opened 2019-12-05
Add rc checking to tid->commit in bdb_close_int

mistachkin:
These changes add a minor sanity check to the bdb_close_int function when calling tid->commit.

### DIFF
--- a/bdb/file.c
+++ b/bdb/file.c
@@ -1670,7 +1670,11 @@ static int bdb_close_int(bdb_state_type *bdb_state, int envonly)
     bdb_unlock_children_lock(bdb_state);
 
     /* Commit */
-    tid->commit(tid, 0);
+    rc = tid->commit(tid, 0);
+    assert(rc == 0);
+    if (rc != 0) {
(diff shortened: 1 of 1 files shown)


tests/docker/Dockerfile.install
```diff
@@ -45,7 +45,7 @@ RUN cd /comdb2 &&  \
     rm -rf build && \
     mkdir build && \
     cd build && \
-    cmake -DWITH_TCL=1 .. && \
+    cmake -DCMAKE_BUILD_TYPE=Debug -DWITH_TCL=1 .. && \
```
adizaimi: Will this run tests for roborivers with Debug on? If so, I may be wrong but roboperf uses that same executable, so we will start seeing different performance numbers -- it may not be a good idea to enable here for now.

mistachkin: This change (when I make it) will always be manually backed out prior to merge.

mistachkin:
The RoboRivers tests results for this PR are interesting: http://comdb2.s3-website-us-east-1.amazonaws.com/tests/33942c26/detail.txt

Apparently, this PR may reduce the amount of memory used (i.e. fix a leak)?

PR #1969, opened 2019-12-05
Better handling of very long IN lists

riverszhang89:
When there are `IN` tokens, DOHAST concatenates each one of them to
form a parallelizable child query. It rewrites a new string to replace
the old one on every `IN` token. This is quite inefficient as it suffers
from unnecessary realloc and memcpy.

The pull request changes it to use `sqlite3_str_append*` API to append
the tokens to the output string. It is faster and more memory friendly.

For a `IN` list of roughly 400,000 elements, the change runs 150x faster
than the master branch.

(DRQS 152696248)

### DIFF
--- a/sqlite/src/expr.c
+++ b/sqlite/src/expr.c
@@ -5926,7 +5926,7 @@ static char* sqlite3ExprDescribe_inner(
     }
     case TK_IN: {
       int  i;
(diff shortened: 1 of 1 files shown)


riverszhang89:
Thank you Joe!

PR #1970, opened 2019-12-06
Port ref_sync_wait_txn logic to master

markhannum:
Porting the R6 hotfix logic to master.  If ref-sync fails to clear the ref-sync counter, mp_sync thread grabs a list of active transactions and polls at the end of the loop waiting for them to complete.  This prevents a single transaction from being blocked more than a single time by the sync-thread.


markhannum:
This was reviewed under R6

PR #1975, opened 2019-12-09
Remove sess->initstart because it is unnecessary 

adizaimi:
* remove sess->initstart because it is unnecessary 
* sess->start is renamed to startus


db/osqlsession.c
```diff
@@ -430,8 +430,8 @@ int osql_sess_test_slow(osql_sess_t *sess)
  */
 void osql_sess_getsummary(osql_sess_t *sess, int *tottm, int *rtt, int *rtrs)
 {
-    *tottm = sess->end - sess->initstart;
-    *rtt = sess->end - sess->start;
+    *tottm = U2M(sess->endus - sess->startus);
```
dorinhogea: Please remove redundant argument from call

adizaimi: Done.

adizaimi:
Thank you Dorin.

PR #1978, opened 2019-12-11
fix regression do not try to run bplogs ending with an osql_xerr

dorinhogea:
Regression introduced in https://github.com/bloomberg/comdb2/pull/1960.
Normally, if a bplog ends with an OSQL_XERR packet (because replicant decided to abort the sql transaction), the bplog is not processed and we simply discard it.  After the PR, the log is processed, aborted, db spews (since it is not expected to process an OSQL_XERR), but albeit the extra-work, things are normal (therefore the regression tests did not capture the issue).


dorinhogea:
chat rubberstamped

PR #1983, opened 2019-12-12
Refactor trn repo mtx holding time

adizaimi:
* Try to hold trn_repo_mtx for a smaller amount of time time while inserting trn
* Cleanup trn properly,

### DIFF
--- a/bdb/bdb_osqltrn.c
+++ b/bdb/bdb_osqltrn.c
@@ -340,7 +340,7 @@ bdb_osql_trn_t *bdb_osql_trn_register(bdb_state_type *bdb_state,
     trn = (bdb_osql_trn_t *)calloc(1, sizeof(bdb_osql_trn_t));
     if (!trn) {
         *bdberr = BDBERR_MALLOC;
-        goto done;
+        return NULL;
     }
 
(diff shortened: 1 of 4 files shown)


bdb/bdb_osqltrn.c
```diff
@@ -328,24 +337,24 @@ bdb_osql_trn_t *bdb_osql_trn_register(bdb_state_type *bdb_state,
...
     trn->shadow_tran = shadow_tran;
-    trn->trak = (trak & SQL_DBG_BDBTRN) || (trn_repo->trak);
-
     listc_init(&trn->bkfill_list, offsetof(bdb_osql_log_t, lnk));
 
+    Pthread_mutex_lock(&trn_repo_mtx);
+    if (!trn_repo) {
+        cleanup_trn(trn);
+        return NULL;
+    }
```
mistachkin: I'm a bit confused here.  When is this lock on trn_repo_mtx unlocked?

adizaimi: Bah, fixed. Was too hasty. Good news is it does not get triggered unless trn_repo is null which is at exit only.
Thanks for catching.

adizaimi:
Thank you Joe.

PR #1984, opened 2019-12-13
Remove sorese info

dorinhogea:
Blocksql exorcism part 2 - cutting deeper:
Sorese_info_t is a summary of the osql_sess_t. Its purpose was to allow blocksql code (that had more than one osql_sess_t object), and socksql code (that had only one osql_sess_t), live together during code transition from block to socket protocols.
This PR replaces sorese_info_t with osql_sess_t, which allow us to throw away some cumbersome workarounds.

There is another PR that builds on top of this and abstracts away some session fields, as well as the untouched yet blocksql_tran.

This PR validates the changes in the session dispatch and termination before building more on top of it.


db/comdb2.h
```diff
@@ -1046,20 +1046,6 @@ typedef struct {
     int int3[3];
 } fstblkseq_t;
 
-/* this stores info about the socksql/recom requester */
-typedef struct sorese_info {
-    unsigned long long rqid; /* not null means active */
```
dorinhogea: these were duplicated in osql_sess_t

db/comdb2.h
```diff
@@ -1225,6 +1211,89 @@ typedef struct bpfunc_lstnode {
...
+
+    OSQL_SNAPISOL_REQ = 7,
+    OSQL_SNAP_UID_REQ = 8,
+    OSQL_MAX_REQ = 9,
+};
+
+/* Magic rqid value that means "please use uuid instead" */
+#define OSQL_RQID_USE_UUID 1
+
+struct osql_sess {
```
dorinhogea: this will be split in two, hiding fields that do not need exposure to rest of the world

db/osqlsession.h
```diff
@@ -141,27 +83,6 @@ int osql_sess_getcrtinfo(void *obj, void *arg);
  */
 void *osql_sess_getbptran(osql_sess_t *sess);
 
-/* Lock the session */
```
dorinhogea: all these were a sign of bad encapsulation;  these are gone, since knowledge of the locking and the implied fields is isolated to osqlsession.c

PR #1987, opened 2019-12-17
Fix eventlog: less time under lock, better checks for test

adizaimi:
* reorganize eventlog_add so we get eventlog_lk less often and for less time
* make eventlog_replay.test more robust by making sure to talk to same node for all requests and by comparing eventlog output with a predefined number of times each query fingerprint should appear.


db/toblock.c
```diff
@@ -2558,7 +2558,9 @@ static inline int check_for_node_up(struct ireq *iq, block_state_t *p_blkstate)
 static int toblock_main_int(struct javasp_trans_state *javasp_trans_handle,
                             struct ireq *iq, block_state_t *p_blkstate)
 {
+#ifdef DEBUG_BLKSEQ
```
riverszhang89: The scattered `#ifdef DEBUG_BLKSEQ` blocks make the code somewhat difficult to read. If the trace is considered too verbose even under the DEBUG verbosity, maybe we should get rid of it?

adizaimi: yeah...probably can remove.

dorinhogea:
Can we change the regression test to record the actual eventlog, instead of just validating the number of lines?

adizaimi:
I tried that but there are a lot of member variables I would need to stub out (replace with XXX) in the content such as time, uuid, perf and timing info. So not sure how to get something reliable. I know I should get X selects, Y inserts, Z updates and so on, and check for that?

> Can we change the regression test to record the actual eventlog, instead of just validating the number of lines?

db/eventlog.c
```diff
@@ -454,22 +454,8 @@ static void eventlog_add_newsql(cson_object *obj, const struct reqlogger *logger
...
-    Pthread_mutex_lock(&eventlog_lk);
-    if (eventlog == NULL || !eventlog_enabled) {
-        Pthread_mutex_unlock(&eventlog_lk);
-        return;
-    }
-
-    bool isSql = logger->event_type && (strcmp(logger->event_type, "sql") == 0);
-    bool isSqlErr = logger->error && logger->stmt;
-
-    if ((isSql || isSqlErr) && !hash_find(seen_sql, logger->fingerprint)) {
```
adizaimi: Moved the adding eventlog_add_newsql() to the caller, while it already has the lock so no need to get/release lock here.

db/toblock.c
adizaimi: Left this part as it was before, for now.

adizaimi:
> Can we change the regression test to record the actual eventlog, instead of just validating the number of lines?

Added a terse check that we get the correct number of queries for each fingerprint. It is better than the silly count, and more compressed than checking sql (which can have different values in the insert).

db/eventlog.c
```diff
@@ -547,34 +534,40 @@ static void eventlog_add_int(cson_object *obj, const struct reqlogger *logger)
...
+
+    if ((isSql || isSqlErr) && !hash_find(seen_sql, logger->fingerprint)) {
+        eventlog_add_newsql(logger);
+    }
+}
+
 void eventlog_add(const struct reqlogger *logger)
 {
-    Pthread_mutex_lock(&eventlog_lk);
     if (eventlog == NULL || !eventlog_enabled) {
```
adizaimi: Cursory check should be fine without lock. We do another check below under lock.

db/eventlog.c
```diff
@@ -547,34 +534,40 @@ static void eventlog_add_int(cson_object *obj, const struct reqlogger *logger)
...
         return;
     }
 
-    cson_value *val;
-    cson_object *obj;
-
-    eventlog_count++;
-    if (eventlog_every_n > 1 && eventlog_count % eventlog_every_n != 0) {
-        Pthread_mutex_unlock(&eventlog_lk);
+    int loc_count = ATOMIC_ADD64(eventlog_count, 1);
```
adizaimi: This does not need to be done under lock if counter is atomically incremented.

db/eventlog.c
```diff
@@ -547,34 +534,40 @@ static void eventlog_add_int(cson_object *obj, const struct reqlogger *logger)
...
-    eventlog_add_int(obj, logger);
+    cson_value *val = cson_value_new_object();
+    cson_object *obj = cson_value_get_object(val);
+    populate_obj(obj, logger);
 
     Pthread_mutex_lock(&eventlog_lk);
-    if (eventlog != NULL && eventlog_enabled)
+
+    if (eventlog != NULL && eventlog_enabled) {
+        if (bytes_written > eventlog_rollat) {
```
adizaimi: This will now roll, and add to fingerprints on top of what it was doing earlier. The time spent with the lock held is possibly similar to before, but at least we are lock/release only once instead of several times, so contention and total time should be less.

db/eventlog.c
```diff
@@ -733,12 +726,9 @@ void eventlog_process_message(char *line, int lline, int *toff)
 void log_deadlock_cycle(locker_info *idmap, u_int32_t *deadmap,
                         u_int32_t nlockers, u_int32_t victim)
 {
-    Pthread_mutex_lock(&eventlog_lk);
```
adizaimi: check can be preliminary not under -- we check again later under lock.

db/eventlog.c
markhannum: gotcha

PR #1991, opened 2019-12-20
Micro-optimizations

riverszhang89:
A few micro-optimizations which should bump the TPS of sysbench write tests by 5% to 10%.

### DIFF
--- a/berkdb/log/log_put.c
+++ b/berkdb/log/log_put.c
@@ -1159,13 +1159,10 @@ __log_putr(dblp, lsn, dbt, prev, h)
 	if (lp->num_segments > 1) {
 		tmplsn = *lsn;
 		ret = __log_fill_segments(dblp, lsn, &tmplsn, hdr, nr);
-		assert(tmplsn.offset == lsn->offset + hdr->size);
+		assert(tmplsn.offset == lsn->offset + nr);
 	} else {
 		ret = __log_fill(dblp, lsn, hdr, (u_int32_t)nr);
 	}
-
-	if (LOG_SWAPPED())
-		__log_hdrswap(hdr, CRYPTO_ON(dbenv));
 
 	if (ret != 0)
 		goto err;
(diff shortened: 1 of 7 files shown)


berkdb/log/log_put.c
```diff
@@ -1159,14 +1159,11 @@ __log_putr(dblp, lsn, dbt, prev, h)
 	if (lp->num_segments > 1) {
 		tmplsn = *lsn;
 		ret = __log_fill_segments(dblp, lsn, &tmplsn, hdr, nr);
-		assert(tmplsn.offset == lsn->offset + hdr->size);
+		assert(tmplsn.offset == lsn->offset + nr);
 	} else {
 		ret = __log_fill(dblp, lsn, hdr, (u_int32_t)nr);
 	}
```
markhannum: Can we really git rid of this?

markhannum: oh - it's because the header isn't used beyond this .. ic

util/object_pool.c
```diff
@@ -903,11 +903,12 @@ static int objpool_borrow_int(comdb2_objpool_t op, void **objp, long nanosecs,
 
     /* update access history */
     rec = (pooled_object *)hash_find(op->history, objp);
+    Pthread_mutex_unlock(&op->data_mutex);
+
     rec->active = 1;
     ++rec->nborrows;
     rec->tid = pthread_self();
```
adizaimi: I think that these three need to be done under the lock , otherwise another thread can access this record and find this info inconsistent.
However there is room for improvement to do the creation of rec outside of lock object_create(), and only add it under lock.

berkdb/log/log_put.c
riverszhang89: Right Mark - `__log_hdrswap` doesn't swap `size` either.

util/object_pool.c
riverszhang89: It's okay. `get_impl` makes sure that nothing can grab the same object again after it returns (because it's done while holding the lock), which implies the caller's exclusive access to the object's history.

riverszhang89:
@markhannum @mistachkin Thank you for the review!

PR #1992, opened 2019-12-21
Verify: stop verify when client disconnected and not verbose

adizaimi:
Verify: stop verifying if client disconnected (and verbose is turned off).
The bug is that if we don't specify verbose in the verify command, verify continues to process even after cdb2sql client has disconnected. The reason is because we currently have no way to check whether client is still connected if verbose is off. With this checkin we now pass in sbuf object which we can perform such check. Now we have to avoid printing directly messages to sbuf because that breaks clients connecting via cdb2sql.  
Also in this checkin:
* Added to insert_lots.test checks to ensure all threads in verify thdpool exited when client exits
* Moved sequential verify to be performed by the thdreadpool just like the parallel one


bdb/bdb_verify.c
```diff
@@ -45,37 +45,39 @@ extern int is_comdb2_index_expression(const char *dbname);
...
+    }
+    
     if (par->sb) {
         if (wrote < sizeof(lbuf) - 1)
             strcat(lbuf, "\n");
-        return sbuf2printf(par->sb, lbuf) >= 0 ? 0 : -1;
-    } else if (par->lua_callback)
-        return par->lua_callback(par->lua_params, lbuf);
+        int rc = sbuf2printf(par->sb, lbuf) >= 0 ? 0 : -1;
+        sbuf2flush(par->sb);
```
mponomar: sbuf2flush may also return an error if the client drops

bdb/bdb_verify.c
```diff
@@ -202,47 +204,50 @@ static int fix_blobs(bdb_state_type *bdb_state, DB *db, DBC **cdata,
...
         goto out;
 
-    // enough time has passed, attempt to update
-    int res = CAS32(par->last_reported, last, now);
+    // enough time has passed, attempt to update last_connection_check
+    int res = CAS32(par->last_connection_check, last, t_ms);
     if (!res)
-        goto out; // someonelse updated, get out
+        goto out; // another thread updated, nothing to do
+    printf("AZ checking connection\n");
```
adizaimi: will remove :)

bdb/bdb_verify.c
```diff
@@ -202,47 +204,50 @@ static int fix_blobs(bdb_state_type *bdb_state, DB *db, DBC **cdata,
...
     }
 
-    int rc;
+    // one of the threads gets here every second, and we want to skip printing
+    // if progress_report_seconds is zero or if the time passed is not enough
+    if (!par->progress_report_seconds || 
+        (++par->progress_report_counter) % par->progress_report_seconds != 0)
+        goto out;
+
+    printf("AZ: par->progress_report_counter %d , par->progress_report_seconds %d \n", par->progress_report_counter, par->progress_report_seconds);
```
adizaimi: will remove :)

bdb/bdb_verify.c
```diff
@@ -1247,7 +1254,8 @@ static inline void enqueue_work(td_processing_info_t *work,
             verify_thdpool = NULL;
         }
     }
-    if (!verify_thdpool) {
+
```
adizaimi: thdpool_enqueue() takes a description which helps debugging threads when dumped.

tests/insert_lots.test/README
```diff
@@ -1,2 +1,5 @@
 Insert many rows to db from multithreaded writer, 
 and make sure that they all succeeded.
+
+In this testcas, because there are lots of rows, we will be testing
```
adizaimi: testcas??

tests/insert_lots.test/README
```diff
@@ -1,5 +1,5 @@
 Insert many rows to db from multithreaded writer, 
 and make sure that they all succeeded.
 
-In this testcas, because there are lots of rows, we will be testing 
+In this testcast, because there are lots of rows, we will be testing
```
mistachkin: testcase?

nirbhayc: telecast?

PR #1994, opened 2019-12-23
Rollover queuedbs between one and two files based on a tunable

mistachkin:
These changes add a new tunable 'queuedb_file_threshold', which is the number of megabytes before a second file should be added for a queuedb.  When there are two files for a queuedb, the first one is always being **consumed from** and the second one is alrways being **added to**.  Furthermore, when the first file is empty, it will be automatically cleaned up and the second file will be swapped over to it, thus becoming the new first file.


lua/sp.c
```diff
@@ -6555,11 +6564,22 @@ static int setup_sp_for_trigger(trigger_reg_t *reg, char **err,
...
         return -1;
     }
+    rc = bdb_lock_table_read_fromlid(db->handle,
+              bdb_get_lid_from_cursortran(clnt->dbtran.cursor_tran));
+    if (rc != 0) {
+        unlock_schema_lk();
+        return luaL_error(L, "cannot read-lock queue for sp:%s", spname);
+    }
+    unlock_schema_lk();
+
```
akshatsikarwar: Looks like repetition from `db_consumer`. Perhaps refactor to have one function to initialize?

bdb/llmeta.c
```diff
@@ -1967,6 +1967,159 @@ int bdb_del_file_versions(
     return bdb_chg_file_versions(bdb_state, input_trans, NULL, bdberr);
 }
 
+static int bdb_del_file_version(
```
akshatsikarwar: Consider using `kv_del` in `llmeta.c`. Perpare your key and pass to that routine so you don't have to repeat boilerplate for various CURD on llmeta. There also exists `kv_put` and friends for other ops.

lua/sp.c
mistachkin: Good idea.  There is still much polishing to do on this pull request.

bdb/llmeta.c
mistachkin: I thought about going that route; however, I thought it would be useful to build on the existing LLMETA abstraction layer by providing the "single delete" primitive, using the standard ("well-tested"?) boilerplate.

akshatsikarwar: Indeed. `kv_del` was written with that intention.

mistachkin: Right.  But `kv_del` appears to be generic for any LLMETA key.  What I was after was something that did also encapsulate the "file version" key building and other related logic.

plugins/dbqueuedb/dbqueuedb.c
```diff
@@ -366,15 +366,15 @@ static unsigned long long dbqueue_get_front_genid(struct dbtable *table,
...
     pthread_mutex_t *mu;
     pthread_cond_t *cond;
 
     init_fake_ireq(table->dbenv, &iq);
     iq.usedb = table;
     genid = 0;
 
-    rc = bdb_trigger_subscribe(table->handle, &cond, &mu, &open);
+    rc = bdb_trigger_subscribe(table->handle, &cond, &mu, &status);
     if (rc != 0) {
```
mistachkin: Need special handling for paused here...?

schemachange/sc_queues.c
```diff
@@ -548,14 +542,15 @@ static int perform_trigger_update_int(struct schema_change_type *sc)
...
         rc = trans_start(&iq, NULL, (void *)&tran);
         if (rc) {
             logmsg(LOGMSG_ERROR, "%s: trans_start rc %d\n", __func__, rc);
             goto done;
         }
 
         unsigned long long ver = 0;
-        if (bdb_get_file_version_qdb(db->handle, tran, &ver, &bdberr) == 0) {
+        if (bdb_get_file_version_qdb(db->handle, tran, 0, &ver, &bdberr) == 0) {
             sc_del_unused_files_tran(db, tran);
```
mistachkin: need to drop both files here for queuedb...

schemachange/sc_queues.c
```diff
@@ -599,3 +594,291 @@ int finalize_trigger(struct schema_change_type *s)
...
+               tran_type *tran)
+{
+    struct dbtable *db = getqueuebyname(queue_name);
+    if (db == NULL) {
+        logmsg(LOGMSG_ERROR, "%s: no such queuedb %s\n", __func__,
+               queue_name);
+        return -1;
+    }
+    int rc = bdb_trigger_pause(db->handle);
+    if (rc != 0) return rc;
```
mistachkin: error handling semantics here?  is this fatal?   unpause on failure?

sqlite/src/func.c
```diff
@@ -751,6 +754,33 @@ static void guidFromByteFunc(
...
+  sqlite3_value **argv
+){
+  assert( argc==1 );
+  if( sqlite3_value_type(argv[0])==SQLITE_TEXT ){
+    const unsigned char *zMsg = sqlite3_value_text(argv[0]);
+    if( zMsg ){
+      Pthread_mutex_lock(&test_log_file_mtx);
+      if( gbl_test_log_file!=NULL ){
+        FILE *file = fopen(gbl_test_log_file, "a+");
+        if( file!=NULL ){
```
mistachkin: Change this to use open() system call and sbuf2...

bdb/queuedb.c
```diff
@@ -76,15 +88,210 @@ static uint8_t *queuedb_key_put(struct queuedb_key *p_queuedb_key,
...
+                                 (FCRON)queuedb_cron_event, NULL,
+                                 NULL, NULL, dbenv, NULL, err, NULL);
+        if (p == NULL) {
+            logmsg(LOGMSG_ERROR, "Failed to schedule next queuedb event. "
+                            "rc = %d, errstr = %s\n",
+                    err->errval, err->errstr);
+        }
+    }
+    if (gbl_queuedb_file_threshold <= 0) return NULL;
+    if ((dbenv == NULL) || (dbenv->master != gbl_mynode)) return NULL;
```
mistachkin: grab the BDB lock and then check master-ship.

bdb/queuedb.c
```diff
@@ -76,15 +88,210 @@ static uint8_t *queuedb_key_put(struct queuedb_key *p_queuedb_key,
...
+    bdb_thread_event(dbenv->bdb_env, BDBTHR_EVENT_START_RDWR);
+    for (int i = 0; i < dbenv->num_qdbs; i++) {
```
mistachkin: grab the schema lock for the duration of this loop...

berkdb/env/env_method.c
```diff
@@ -1371,22 +1378,22 @@ __dbenv_get_durable_lsn(dbenv, lsnp, generation)
...
-	const uint8_t **open;
+	const uint8_t **status;
 {
 	int rc = 1;
 	struct __db_trigger_subscription *t;
 	t = __db_get_trigger_subscription(fname);
 	Pthread_mutex_lock(&t->lock);
-	if (t->open) {
+	if (t->status == TRIGGER_SUBSCRIPTION_OPEN) {
 		++t->active;
```
mistachkin: This is probably not correct.  Should check if status is open or paused?

berkdb/env/env_method.c
```diff
@@ -1427,7 +1434,50 @@ __dbenv_trigger_close(dbenv, fname)
...
+
+static int
+__dbenv_trigger_pause(dbenv, fname)
+	DB_ENV *dbenv;
+	const char *fname;
+{
+	struct __db_trigger_subscription *t;
+	t = __db_get_trigger_subscription(fname);
+	Pthread_mutex_lock(&t->lock);
+	t->status = TRIGGER_SUBSCRIPTION_PAUSED;
```
mistachkin: Maybe add an assert that status is not closed at this point?

bdb/queuedb.c
```diff
@@ -201,71 +418,94 @@ int bdb_queuedb_walk(bdb_state_type *bdb_state, int flags, void *lastitem,
...
-        if (rc == DB_LOCK_DEADLOCK) {
-            *bdberr = BDBERR_DEADLOCK;
-            rc = -1;
-            goto done;
-        } else if (rc == DB_NOTFOUND) {
-            /* EOF */
-            rc = 0;
-            goto done;
+        if (flags & BDB_QUEUE_WALK_RESTART) {
+            /* this is a restart, and lastitem containts a copy of the last key when
```
mistachkin: Double check this...  What are the expected semantics for the BDB_QUEUE_WALK_RESTART flag and do we still meet them?

plugins/dbqueuedb/dbqueuedb.c
mistachkin: Given how this function is used, I think returning a zero genid when paused is the correct behavior, i.e. this function is only used to determine when an inactivity warning should be emitted into the log file and I believe that when a queuedb is paused, it should not have these warnings.

schemachange/sc_queues.c
mistachkin: Actually (per comment), I think this code is OK.

lua/sp.c
```diff
@@ -696,19 +706,52 @@ char *sp_column_name(struct response_data *arg, int col)
...
 
+static int grab_qdb_table_read_lock(struct sqlclntstate *clnt,
+                                    struct dbtable *db, trigger_reg_t *reg,
+                                    int have_schema_lock)
+{
+    if (!gbl_queuedb_read_locks) return 0;
+    if ((reg != NULL) && reg->qdb_locked) {
+        return 0; /* we already have the table read lock */
+    }
+    if (!have_schema_lock && (tryrdlock_schema_lk() != 0)) {
```
markhannum, line 717: Since you are making sure to call this only once, you should now be avoiding lock ordering issues with schema-lk and the table-locks .. no?  I think you can block on the schema-lk rather than trylock.

bdb/queuedb.c
```diff
@@ -335,25 +555,45 @@ int bdb_queuedb_stats(bdb_state_type *bdb_state,
...
             rc = -1;
             goto done;
         } else if (rc == DB_NOTFOUND) {
-            /* EOF */
-            rc = 0;
-            goto done;
+            if (!did_dbcp2_first && (dbcp2 != dbcp1)) {
+                rc = bdb_cget_unpack(bdb_state, dbcp2, &dbt_key, &dbt_data, &ver,
+                                     DB_FIRST);
+                did_dbcp2_first = 1; goto chk_db_first_rc;
```
markhannum, line 591: This is wrong- chk_db_first_rc goto target should be repositioned to line 579 .. 

>> this will error out unintentionally if file0 returns NOTFOUND, but file1 returns 0

PR #1998, opened 2019-12-28
Disable nocget optimization as it has issues with dup indices

adizaimi:
Disable skip_cget_in_db_put which was added as an optimization to not crack the btree twice upon write, and also help reduce deadlocks.
Modified insert_lots.test to have a dup index which exhibits the bug.


adizaimi:
Thiss will go into R7 only, in PR #2004. Masterbranch will get proper fix https://github.com/bloomberg/comdb2/pull/1999.

PR #1999, opened 2019-12-28
Fix: Get lock on node about to be split to avoid insert in wrong order

adizaimi:
Bug is that when insertion into an index is contended, sometimes we get values inserted in the wrong location (out of order). 
The reason for this bug is because we removed the cget() call in a previous checkin, however that cget() call write locks the page we are going to write. That write lock ensures that if we need to split the page, no other thread will sneak past and perform the split. Without that write lock, another thread can do the split of this node and we are left with a pointer to the left subtree--that is the wrong place to start search to perform the insert, thus we were getting wrong order in some of the keys inserted.
One solution is to perform bam_search from the root node rather than from the result of a split. 
However a better solution is to write lock the page before splitting it, so we are ensured to be the ones performing the split.
Including reproducer for the bug.


adizaimi:
For the record, I include here the active locks before and after releasing the lock stack, and it will help explain why this PR fixes the issue:
```
BEFORE: printing locks held fname=XXX.t_0000000000150000.index own=0 stack=1
2020/01/07 20:22:23 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
2020/01/07 20:22:23 Locks grouped by object
2020/01/07 20:22:23 TID            MRLocker Locker   Mode      Count Status  ----------------- Object ---------------
2020/01/07 20:22:23 0x7f3b2ec88700        0 80031a71 READ          1 HELD     tablelock t
2020/01/07 20:22:23 0x7f3b310f3700        0    7a13b READ          1 HELD     tablelock t
2020/01/07 20:22:23 0x7f3b2f089700        0 80031a73 READ          1 HELD     tablelock t
2020/01/07 20:22:23 0x7f3b2ec88700 80031a71 80031a72 WRITE         2 HELD    XXX.t_0000000000150000.index page     23835
2020/01/07 20:22:23 0x7f3b2f089700 80031a73 80031a74 WRITE         1 WAIT    XXX.t_0000000000150000.index page     23835
2020/01/07 20:22:23 0x7f3b2ec88700 80031a71 80031a72 WRITE         3 HELD    XXX.t_0000000000140000.datas3 page       264
2020/01/07 20:22:23 0x7f3b2f089700 80031a73 80031a74 WRITE         3 HELD    XXX.t_0000000000140000.datas4 page       264
AFTER: printing locks held own=0 stack=1
2020/01/07 20:22:23 =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
2020/01/07 20:22:23 Locks grouped by object
2020/01/07 20:22:23 TID            MRLocker Locker   Mode      Count Status  ----------------- Object ---------------
2020/01/07 20:22:23 0x7f3b2ec88700        0 80031a71 READ          1 HELD     tablelock t
2020/01/07 20:22:23 0x7f3b2f089700        0 80031a73 READ          1 HELD     tablelock t
2020/01/07 20:22:23 0x7f3b2ec88700 80031a71 80031a72 WRITE         1 HELD    XXX.t_0000000000150000.index page     23835
2020/01/07 20:22:23 0x7f3b2f089700 80031a73 80031a74 WRITE         1 WAIT    XXX.t_0000000000150000.index page     23835
2020/01/07 20:22:23 0x7f3b2ec88700 80031a71 80031a72 WRITE         3 HELD    XXX.t_0000000000140000.datas3 page       264
2020/01/07 20:22:23 0x7f3b3261e700 80031a75 80031a76 WRITE         3 HELD    XXX.t_0000000000140000.datas5 page       264
2020/01/07 20:22:23 0x7f3b2f089700 80031a73 80031a74 WRITE         3 HELD    XXX.t_0000000000140000.datas4 page       264
```
Notice the lock count for XXX.t_0000000000150000.index page     23835, which goes from 2 to 1, thus we call bam_split with this page held in write lock mode. Without the fix in this PR, the lock count goes to zero.

adizaimi:
Also for completeness, I tested the changes in this PR, but removing the line getting the lock in: https://github.com/bloomberg/comdb2/pull/2007 and as you can see the test fails because of verify error.

adizaimi:
Wrong button, ment to comment only.

PR #2000, opened 2019-12-30
Fix omission of 'drop_view' from the 'get_ddl_type_str' schema change function.

mistachkin:
The function modified by this pull request appears to be used only from the schema change status (reporting) virtual table.  When a view is being dropped, it should now be reported with a type column value of "view".


mistachkin:
Both of the remaining test failures are confirmed false alarms.

PR #2001, opened 2019-12-30
cdb2_verify: Detect inter-page out of order btree key entries in verify

adizaimi:
Start the search at root node (pg 1) and do a depth first search to check for out-of-order keys between pages. We will now detect issues with pages containing an out of order key being inserted at the end of a page:
```/dev/shm/master/build/db/cdb2_verify /dev/shm/test_12546/ixentrymissingreproducer12546/t_*index  
2019/12/30 15:03:00 cdb2_verify: Page 3599: out-of-order key at entry 0 *p1=0000004b3fc90005 *p2=0000004b3d1b0001
2019/12/30 15:03:00 cdb2_verify: DB->verify: /dev/shm/test_12546/ixentrymissingreproducer12546/t_000000493c9b0000.index: DB_VERIFY_BAD: Database verification failed
```

### DIFF
--- a/berkdb/btree/bt_verify.c
+++ b/berkdb/btree/bt_verify.c
@@ -374,9 +374,7 @@ __bam_vrfy(dbp, vdp, h, pgno, flags)
 		EPRINT((dbenv,
 		    "Page %lu: item order check unsafe: skipping",
 		    (u_long)pgno));
-	}
-#if 0
-	else if (!LF_ISSET(DB_NOORDERCHK) && (ret =
+	} else if (!LF_ISSET(DB_NOORDERCHK) && (ret =
 	    __bam_vrfy_itemorder(dbp, vdp, h, pgno, 0, 0, 0, flags)) != 0) {
 		/*
 		 * We know that the elements of inp are reasonable.
@@ -388,7 +386,6 @@ __bam_vrfy(dbp, vdp, h, pgno, flags)
 		else
 			goto err;
 	}
-#endif
 
 err:	if ((t_ret = __db_vrfy_putpageinfo(dbenv, vdp, pip)) != 0 && ret == 0)
 		ret = t_ret;
@@ -955,7 +952,7 @@ __bam_vrfy_itemorder(dbp, vdp, h, pgno, nentries, ovflok, hasdups, flags)
 	p1 = &dbta;
 	p2 = &dbtb;
 
-	if (last_key_from_prev_pg.data && (TYPE(h) == P_LBTREE)) {
+	if (last_key_from_prev_pg.data && (TYPE(h) == P_LBTREE) && LF_ISSET(DB_IN_ORDER_CHECK)) {
 		/*
         long int genid = *(long int*)(last_key_from_prev_pg.data+5);
 		printf("setting pgno=%lu from last_key_from_prev_pg, sz=%d, data=%p *data=%016lx\n", (u_long)pgno, last_key_from_prev_pg.size, last_key_from_prev_pg.data, flibc_ntohll(genid));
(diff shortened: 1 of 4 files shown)


berkdb/btree/bt_verify.c
```diff
@@ -1072,9 +1085,11 @@ overflow:		if (!ovflok) {
 			/* comparison succeeded */
 			if (cmp > 0) {
 				isbad = 1;
+                long int genid1 = *(long int*)(p1->data+5);
```
adizaimi: convert these to hex properly with util_tohex().

berkdb/db/db_vrfy.c
```diff
@@ -584,6 +596,59 @@ __db_vrfy_pagezero(dbp, vdp, fhp, flags)
...
+        VRFY_PAGEINFO *pip;
+        if ((ret = __db_vrfy_getpageinfo(vdp, pgno, &pip)) != 0)
+            return (ret);
+        u_int32_t nentries = pip->entries;
+        db_indx_t i;
+        for (i = 1; i < nentries; i+= O_INDX) {
+            BINTERNAL *bi;
+			bi = GET_BINTERNAL(dbp, h, i);
+
+            // printf("from pg %d to pg %d\n", pgno, bi->pgno);
```
adizaimi: Should verify item order for this page too.

berkdb/db/db_vrfy.c
```diff
@@ -584,6 +596,59 @@ __db_vrfy_pagezero(dbp, vdp, fhp, flags)
...
+            ret = __db_vrfy_depthfirst(dbp, vdp, handle, callback, flags, (u_long)bi->pgno);
+            if (ret) 
+                isbad = 1;
+        }
+    } else if (TYPE(h) == P_LBTREE) {
+        ret = __bam_vrfy_itemorder(dbp, vdp, h, pgno, 0, 0, 0, flags);
```
adizaimi: set flag LF_SET(DB_IN_ORDER_CHECK); here instead

berkdb/btree/bt_verify.c
```diff
@@ -980,7 +991,7 @@ __bam_vrfy_itemorder(dbp, vdp, h, pgno, nentries, ovflok, hasdups, flags)
 			bi = GET_BINTERNAL(dbp, h, i);
 			if (B_TYPE(bi) == B_OVERFLOW) {
 				bo = (BOVERFLOW *)(bi->data);
-				goto overflow;
+				abort(); //goto overflow;
```
akshatsikarwar: ?

akshatsikarwar:
What do you think about doing this during `bdb verify` instead. In logical verify, we can check every `next` produces a larger key. That will check all keys are in order, not just key from last page.

berkdb/btree/bt_verify.c
adizaimi: just for my testing -- will remove.

adizaimi:
> What do you think about doing this during `bdb verify` instead. In logical verify, we can check every `next` produces a larger key. That will check all keys are in order, not just key from last page.

Sure, that can be added as well -- issue here is that cdb2_verify checks keys inside page, but not between pages, thus this change. Also, what you propose may still not be enough because next  order may be correct, but tree structure may not.

akshatsikarwar:
> Also, what you propose may still not be enough because next order may be correct, but tree structure may not.

Should be enough - If every next is larger key, then all keys are in correct order.

adizaimi:
> > Also, what you propose may still not be enough because next order may be correct, but tree structure may not.
> 
> Should be enough - If every next is larger key, then all keys are in correct order.

Well, next uses the next pointer in leaf nodes to go to next node, however the actual internal structure of the tree may be such that internal nodes may have bad ordering (that case is NOT what prompted this, but still).

adizaimi:
Is there any harm to having both?

akshatsikarwar:
If corruption is in the middle of a leaf page, then this code will miss it.

adizaimi:
No it will not miss, this is a copy of the existing code which checks internal order—my modification merely prepopulates p2 so the check will include last item from prev page.

akshatsikarwar:
> No it will not miss, this is a copy of the existing code which checks internal order—my modification merely prepopulates p2 so the check will include last item from prev page.

Got it!

adizaimi:
Doing the bdb verify as you suggested in https://github.com/bloomberg/comdb2/pull/2005.

PR #2004, opened 2020-01-02
R7: disable nocget optimization

adizaimi:
Disable skip_cget_in_db_put which was added as an optimization to not crack the btree twice upon write (and also helped reduce deadlocks).
Adding new ixentry_order.test case that Mike initially developed to catch this bug.


adizaimi:
Aproved in https://github.com/bloomberg/comdb2/pull/1998.

PR #2007, opened 2020-01-08
Test that ixentry_order.test catches bad order of keys in btree

adizaimi:
Test that ixentry_order.test catches bad order of keys in btree


adizaimi:
This was ment to test PR https://github.com/bloomberg/comdb2/pull/1999 without the fix. Cosing.

PR #2008, opened 2020-01-08
make osql hshlck a mutex for better performance than a rd/rwlock

adizaimi:
make osql hshlck a mutex for better performance than a rd/rwlock


db/osqlrepository.c
```diff
@@ -36,11 +36,11 @@ struct osql_repository {
 
     hash_t *rqs; /* hash of outstanding requests */
     hash_t *rqsuuid;
-    pthread_rwlock_t hshlck; /* protect the hash */
-
+    pthread_mutex_t hshlck; /* protect the hash */
+    pthread_mutex_t cancelall_mtx; /* cancelall mutex */
```
riverszhang89: Where is this being used?

adizaimi: Merge error, used to be here--Dorin removed, good catch. Thanks

adizaimi:
I saw a very small 2-3s/260s improvement in insert_lots.test from this.

adizaimi:
Further testing of this shows that even in the case of heavy load, read-write lock has no measurable benefit than a mutex and several factors contribute to this: 1) the read count is comparable with the write count (about 2 times higher), 2) the amount of work being done inside the read lock is very small--reading a hash tbl. So mutex may have a small advantage to be used in this case.

PR #2012, opened 2020-01-10
Prevent access to freed memory after trans_commit / trans_abort

mistachkin:
Since the trans_commit() pipeline (eventually) frees its 'tran' argument, make sure to NULL it out so it will not be passed to trans_abort(), trans_commit(), etc again.

### DIFF
--- a/bdb/tran.c
+++ b/bdb/tran.c
@@ -261,6 +261,12 @@ void *bdb_get_physical_tran(tran_type *ltran)
     return ltran->physical_tran;
 }
 
+void bdb_reset_physical_tran(tran_type *ltran)
+{
+    assert(ltran->tranclass == TRANCLASS_LOGICAL);
+    ltran->physical_tran = NULL;
+}
+
 void *bdb_get_sc_parent_tran(tran_type *ltran)
 {
     assert(ltran->tranclass == TRANCLASS_LOGICAL);
@@ -1629,7 +1635,8 @@ int bdb_tran_commit_with_seqnum_int(bdb_state_type *bdb_state, tran_type *tran,
                         "%s:update_shadows_beforecommit nonblocking rc %d\n",
                         __func__, rc);
                 *bdberr = rc;
-                return -1;
+                outrc = -1;
+                goto cleanup;
             }
 
             if (!isabort && tran->committed_child &&
--- a/db/localrep.c
+++ b/db/localrep.c
@@ -854,11 +854,11 @@ again:
         goto done;
     }
     rc = trans_commit(&iq, trans, gbl_myhostname);
+    trans = NULL;
     if (rc) {
         printf("toclear: commit rc %d\n", rc);
         goto done;
     }
-    trans = NULL;
 
 done:
     if (trans) {
--- a/db/sqlanalyze.c
+++ b/db/sqlanalyze.c
@@ -592,11 +592,11 @@ again:
         goto done;
     }
     rc = trans_commit(&iq, trans, gbl_myhostname);
+    trans = NULL;
     if (rc) {
         logmsg(LOGMSG_ERROR, "analyze: commit rc %d\n", rc);
         goto done;
     }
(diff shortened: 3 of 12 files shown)


bdb/tran.c
```diff
@@ -1629,7 +1629,8 @@ int bdb_tran_commit_with_seqnum_int(bdb_state_type *bdb_state, tran_type *tran,
                         "%s:update_shadows_beforecommit nonblocking rc %d\n",
                         __func__, rc);
                 *bdberr = rc;
-                return -1;
+                outrc = -1;
+                goto cleanup;
             }
```
markhannum, line 1634: We leaked the memory here for "tran" .. good catch ..

db/localrep.c
```diff
@@ -854,11 +854,11 @@ int local_replicant_write_clear(struct ireq *in_iq, void *in_trans,
         goto done;
     }
     rc = trans_commit(&iq, trans, gbl_myhostname);
+    trans = NULL;
     if (rc) {
```
markhannum, line 858: Wow .. we access free'd memory if commit fails ..

db/sqlanalyze.c
```diff
@@ -591,11 +591,11 @@ static int local_replicate_write_analyze(char *table)
         goto done;
     }
     rc = trans_commit(&iq, trans, gbl_myhostname);
+    trans = NULL;
     if (rc) {
```
markhannum, line 595: wow .. again .. holy moly ..

db/toblock.c
```diff
@@ -2123,12 +2123,14 @@ static int osql_destroy_transaction(struct ireq *iq, tran_type **parent_trans,
     int rc = 0;
 
     rc = trans_abort(iq, *trans);
+    *trans = NULL;
     if (rc) {
```
markhannum, line 2127: This is harmless actually - its done later in the function ..

db/toblock.c
```diff
@@ -2418,14 +2420,15 @@ static void backout_and_abort_tranddl(struct ireq *iq, tran_type *parent,
 {
     int rc = 0;
     if (rowlocks) {
+        tran_type *new_parent = bdb_get_sc_parent_tran(parent); // TODO: Verify this is correct.
         assert(parent);
```
markhannum, line 2424: This is not necessary- the parent won't be touched - but (based on our conversation during the review) you could NULL the parent's physical transaction after this abort to guarantee a segfault should any later code attempt to access it.

db/toblock.c
```diff
@@ -5624,6 +5627,7 @@ static int toblock_main_int(struct javasp_trans_state *javasp_trans_handle,
                     assert(outrc || iq->sc_running == 0);
                     iq->sc_logical_tran = NULL;
                 } else {
+                    // TODO: Is parent_trans still valid after this?  Does it matter?
```
markhannum, line 5630: Yup- parent_trans will be invalid after this point- you can guarantee that we will segfault if you set it to NULL here

schemachange/sc_alter_table.c
```diff
@@ -1032,6 +1032,7 @@ int finalize_upgrade_table(struct schema_change_type *s)
         if (rc != 0) continue;
 
         rc = trans_commit(&iq, tran, gbl_myhostname);
+        tran = NULL;
     }
```
markhannum, line 1036: Yup - after you call commit, you don't want to call abort on free'd memory

schemachange/sc_queues.c
```diff
@@ -702,11 +702,11 @@ static int perform_trigger_update_int(struct schema_change_type *sc)
 
     if (!same_tran) {
         rc = trans_commit(&iq, tran, gbl_myhostname);
+        tran = NULL;
         if (rc) {
```
markhannum, line 706: wow .. holy moly .. again ..

markhannum:
These are good finds - thanks Joe - good to merge

Issue #2011, opened 2020-01-10
socksqlmasterswings test crashes possibly bad accounting with sess->clients or need_clean


adizaimi:
from master tests last night at http://comdb2.s3-website-us-east-1.amazonaws.com/tests/d9e884b9/detail.txt
Core was generated by `/comdb2/build/db/comdb2 socksqlmasterswings97153 --no-global-lrl --lrl /dedicat'.
Program terminated with signal SIGABRT, Aborted.
#0  0x00007f90692cf428 in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/unix/sysv/linux/raise.c:54
54	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
[Current thread is 1 (Thread 0x7f905e44f700 (LWP 1813))]
(gdb) #0  0x00007f90692cf428 in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007f90692d102a in __GI_abort () at abort.c:89
#2  0x00000000004f6c0a in _destroy_session (phase=0, prq=0x1ba7cd0)
    at /comdb2/db/osqlsession.c:138
#3  osql_close_session (psess=psess@entry=0x1ba7cd0, 
    is_linked=is_linked@entry=0, 
    func=func@entry=0x8e0f20 <__func__.45533> "osql_sess_set_terminate", 
    callfunc=callfunc@entry=0x0, line=line@entry=707)
    at /comdb2/db/osqlsession.c:110
#4  0x00000000004e2c9c in osql_bplog_free (iq=0x1a8be88, 
    are_sessions_linked=are_sessions_linked@entry=0, 
    func=func@entry=0x8e0f20 <__func__.45533> "osql_sess_set_terminate", 
    callfunc=callfunc@entry=0x0, line=line@entry=707)
    at /comdb2/db/osqlblockproc.c:507
#5  0x00000000004f7ef1 in osql_sess_set_terminate (sess=0x16c9728)
    at /comdb2/db/osqlsession.c:707
#6  osql_sess_try_terminate (sess=sess@entry=0x16c9728)
    at /comdb2/db/osqlsession.c:730
#7  0x00000000004f5ecd in osql_repository_add (sess=0x16b6e70, 
    replaced=replaced@entry=0x7f905e44e724) at /comdb2/db/osqlrepository.c:168
#8  0x00000000004ee6cd in sorese_rcvreq (
[...]

PR #1868, opened 2019-10-17
Enable master to return effects

nirbhayc:
.

PR #2013, opened 2020-01-10
Unlock mtxs before calling osql_repository_rem to avoid crash

adizaimi:
Unlock mtx before calling osql_bplog_free() to avoid crash. Before this PR, we destroy the session with the mutexes held via callstack: osql_bplog_free() -> osql_close_session() -> _destroy_session().
fixes PR #2011

### DIFF
--- a/db/osqlcomm.c
+++ b/db/osqlcomm.c
@@ -7265,8 +7265,11 @@ int osql_process_packet(struct ireq *iq, unsigned long long rqid, uuid_t uuid,
 static int sorese_rcvreq(char *fromhost, void *dtap, int dtalen, int type,
                          int nettype)
 {
-    osql_sess_t *sess = NULL;
     uint8_t *malcd = malloc(OSQL_BP_MAXLEN);
+    if (!malcd)
+        goto done;
+
+    osql_sess_t *sess = NULL;
     struct ireq *iq = NULL;
     osql_req_t req;
     bool is_reorder_on = false;
@@ -7282,15 +7285,11 @@ static int sorese_rcvreq(char *fromhost, void *dtap, int dtalen, int type,
     uuid_t uuid;
     int replaced = 0;
 
-    if (!malcd)
-        goto done;
-
     /* grab the request */
     if (osql_nettype_is_uuid(nettype)) {
         osql_uuid_req_t uuid_req;
         sql = (char *)osqlcomm_req_uuid_type_get(&uuid_req, p_req_buf,
                                                  p_req_buf_end);
-
         req.type = uuid_req.type;
         req.rqlen = uuid_req.rqlen;
         req.sqlqlen = uuid_req.sqlqlen;
@@ -7387,15 +7386,11 @@ static int sorese_rcvreq(char *fromhost, void *dtap, int dtalen, int type,
         debug = 1;
     }
 
-    /*
-       Blockproc does not create a copy of the request,
+    /* Blockproc does not create a copy of the request,
        but creates a different thread to work on it
-       Let THAT thread free it...
-    free(p_buf);
-    */
+       Let THAT thread free p_buf (malcd)...  */
 
-    /* for socksql, is this a retry that need to be checked for self-deadlock?
-     */
+    /* for socksql, is it a retry that needs to be checked for self-deadlock? */
     if ((type == OSQL_SOCK_REQ || type == OSQL_SOCK_REQ_COST) &&
         (req.flags & OSQL_FLAGS_CHECK_SELFLOCK)) {
         /* just make sure we are above the threshold */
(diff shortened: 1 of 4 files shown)


db/osqlcomm.c
```diff
@@ -5121,106 +5115,86 @@ int osql_comm_signal_sqlthr_rc(osql_sess_t *sorese, struct errstat *xerr,
...
+            rpl_xerr.dt = *xerr;
 
-                logmsg(LOGMSG_DEBUG,
-                       "%s line %d master signaling %s uuid %s with rc=%d "
-                       "xerr=%d\n",
-                       __func__, __LINE__, sorese->host,
-                       comdb2uuidstr(sorese->uuid, uuid), rc, xerr->errval);
+            osqlcomm_done_xerr_uuid_type_put(&(rpl_xerr), p_buf, p_buf_end);
 
-                msglen = OSQLCOMM_DONE_RPL_LEN;
```
adizaimi: This was a bug probably, should have been OSQLCOMM_DONE_UUID_RPL_LEN.

db/osqlcomm.c
```diff
@@ -5121,106 +5115,86 @@ int osql_comm_signal_sqlthr_rc(osql_sess_t *sorese, struct errstat *xerr,
...
 #endif
-        /* lazy again, works just because node!=0 */
-        irc = offload_net_send(sorese->host, type, buf, msglen, 1);
-        if (irc) {
-            irc = -1;
-            logmsg(LOGMSG_ERROR, "%s: error sending done to %s!\n", __func__,
-                    sorese->host);
-        }
-
-    } else {
```
adizaimi: this was moved up

db/osqlsession.c
```diff
@@ -693,8 +697,11 @@ static int osql_sess_set_terminate(osql_sess_t *sess)
 {
     int rc = 0;
     sess->terminate = OSQL_TERMINATE;
-    rc = osql_repository_rem(sess, 0, __func__, NULL,
-                             __LINE__); /* already have exclusive lock */
+    Pthread_mutex_unlock(&sess->completed_lock);
+    Pthread_mutex_unlock(&sess->mtx);
+
```
adizaimi: This is the real bug, these two should be unlocked before calling osql_repository_rem.

db/osqlcomm.c
```diff
@@ -5111,7 +5105,7 @@ int osql_comm_signal_sqlthr_rc(osql_sess_t *sorese, struct errstat *xerr,
         max = OSQLCOMM_DONE_XERR_RPL_LEN;
     if (OSQLCOMM_DONE_RPL_LEN > max)
         max = OSQLCOMM_DONE_RPL_LEN;
-    buf = alloca(max);
```
akshatsikarwar: I know not introduced in this PR and noted in prior code-comment about being kludgy - runtime decision for buf size. Perhaps not even a performance issue. However, while we are here can use a union to clean this up:
```
union {
   char a[OSQLCOMM_DONE_XERR_UUID_RPL_LEN];
   char b[OSQLCOMM_DONE_UUID_RPL_LEN];
   char c[OSQLCOMM_DONE_XERR_RPL_LEN];
   char d[OSQLCOMM_DONE_RPL_LEN];
} u;
uint8_t *buf = (uint8_t *) &u;
```

db/osqlcomm.c
```diff
@@ -7297,14 +7271,17 @@ int osql_process_packet(struct ireq *iq, unsigned long long rqid, uuid_t uuid,
 static int sorese_rcvreq(char *fromhost, void *dtap, int dtalen, int type,
                          int nettype)
 {
-    osql_sess_t *sess = NULL;
-    uint8_t *malcd = malloc(OSQL_BP_MAXLEN);
+    int rc = 0;
     struct ireq *iq = NULL;
+    uint8_t *malcd = malloc(OSQL_BP_MAXLEN);
```
akshatsikarwar: 32k - we can probably just allocate on stack `uint8_t malc[OSQL_BP_MAXLEN]` instead of malloc, free.

db/osqlcomm.c
```diff
@@ -7419,89 +7392,76 @@ static int sorese_rcvreq(char *fromhost, void *dtap, int dtalen, int type,
...
     }
 
-    /*
-       Blockproc does not create a copy of the request,
+    /* Blockproc does not create a copy of the request,
        but creates a different thread to work on it
-       Let THAT thread free it...
-    free(p_buf);
-    */
+       Let THAT thread free p_buf (malcd)...  */
```
adizaimi: to answer the above comment, malcd needs to survive...see this line.

db/osqlcomm.c
adizaimi: I like this idea.

adizaimi:
Thank you Akshat.

db/osqlcomm.c
akshatsikarwar: thanks - we can try passing it in so don't have to get from heap

akshatsikarwar: also simplifies cleanup on error

adizaimi: good idea, will try in separate PR.

db/osqlcomm.c
nirbhayc: It most certainly was. I tried to fix it as well (https://github.com/bloomberg/comdb2/pull/1868).

PR #2016, opened 2020-01-14
Adjust format of 'sql dump' - no newline, no cnonce, display origin

mponomar:
Removes some things added earlier (cnonce, accidental newline).  Cnonce doesn't seem useful to display here (anyone who feels differently, please tell me).  Newline was breaking expected output for people.  Rivers also discovered (thanks Rivers) that we somehow lost pid/argv0 of the requester in the output - that's now added back.

### DIFF
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -9012,17 +9012,6 @@ void cancel_sql_statement_with_cnonce(const char *cnonce)
         logmsg(LOGMSG_USER, "Query with cnonce %s not found (finished?)\n", cnonce);
 }
 
-/* log binary cnonce in hex format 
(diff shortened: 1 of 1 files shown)


db/sqlglue.c
```diff
@@ -9012,17 +9012,6 @@ void cancel_sql_statement_with_cnonce(const char *cnonce)
         logmsg(LOGMSG_USER, "Query with cnonce %s not found (finished?)\n", cnonce);
 }
 
-/* log binary cnonce in hex format
```
adizaimi: Good riddance, there probably is a dup for this (eventlog) and possibly elsewhere.

PR #2023, opened 2020-01-15
Only malloc necessary amount of memory in sorese_rcvreq

adizaimi:
We blindly allocate 32kb always, but most of the requests are
much smaller than that, so we should only malloc as much memory
 as is needed for the request.


dorinhogea:
This code is gone in the next osql iteration.

adizaimi:
Thanks Dorin. Closing.

PR #2026, opened 2020-01-21
fixed bug in test incrementalcopy

aakash10292:
This pull request is for a bug fix in the test incrementalcopy
$dbnm is not defined in the script, and as a result getmaster() function returns "" for master. This subsequently fails(silently, does not stop the test from proceeding) cdb2sql queries against master. 
Changed $dbnm to $DBNAME


adizaimi:
Since it is related and you are at it, ixentry_order has the same issue, and there are a couple of tests that dont use $dbnm even though they set it.

PR #2027, opened 2020-01-21
check bdb lock desired in dohsql master

dorinhogea:
Master thread waits for children sql thread to provides additional rows; the children sql threads can be blocked waiting on a bdb_lock_desired() flag.  Make sure master releases the bdb read lock similarly,

### DIFF
--- a/db/dohsql.c
+++ b/db/dohsql.c
@@ -1753,6 +1753,17 @@ retry_row:
     }
     if (rc != SQLITE_OK)
         return rc;
+    /* we look at all contributing children, and need to wait;
+       since we have the bdb read lock here, check if we need
+       to run recovery_deadlock */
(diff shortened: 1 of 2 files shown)


db/dohsql.c
```diff
@@ -759,7 +759,14 @@ static int dohsql_dist_next_row(struct sqlclntstate *clnt, sqlite3_stmt *stmt)
         }
     }
     if (!empty) {
-        poll(NULL, 0, 10);
+        if (bdb_lock_desired(thedb->bdb_env)) {
```
riverszhang89: Maybe we could wrap the logic into a function so we don't repeat ourselves?

dorinhogea: It is just an if, I prefer that to a another wrapper on recover_deadlock

dorinhogea:
Thank you Mike

PR #2029, opened 2020-01-22
Make tablename pointer to interned string

adizaimi:
make tablename pointer to interned string


adizaimi:
Resubmitted as https://github.com/bloomberg/comdb2/pull/2603

PR #2033, opened 2020-01-22
fix error path

dorinhogea:
There is a bug in the changes to osql done in https://github.com/bloomberg/comdb2/pull/1984, when a session fails immediately after creation.


dorinhogea:
thank you Adi

PR #2038, opened 2020-01-27
TLSv1.3 post-handshake authentication, ASLR, etc.

### LINKED ISSUE #2028: TLS 1.3 post-handshake client authentication breaks our SSL connection logic
In TLS 1.2, [client authentication happens before handshake finishes](https://tools.ietf.org/html/rfc5246#section-7.3), which means that a client is good to read or write application data after a successful handshake.

However, In TLS 1.3, [client authentication happens after handshake](https://tools.ietf.org/html/rfc8446#section-2). A bad client (a revoked or expired cert) may see a successful handshake but encounter an error when exchanging application data with the server.

This protocol change breaks our connection logic. Because the error happens when a bad client tries to read from or write to the server, it's treated as a network error by the current logic. This then results in client retries, which are doomed to fail again.

Note that this is not a security issue. Server would still be correctly rejecting retries from a bad client, when compiled with openssl 1.1.1 or above which implements TLS 1.3.

riverszhang89:
Fixes #2028.

### DIFF
--- a/tests/mem_tracker.test/runit
+++ b/tests/mem_tracker.test/runit
@@ -7,8 +7,10 @@ dbnm=$1
 host=`cdb2sql ${CDB2_OPTIONS} -s --tabs $dbnm default 'SELECT comdb2_host()'`
 
 # Get the base address of comdb2 binary in the address space of the db process
-dbpid=`pgrep -a comdb2 | grep $dbnm | awk '{print $1}'`
-baseaddr=`pmap $dbpid | grep comdb2$ | head -1 | awk '{printf "0x%s", $1}'`
+dbpid=`ssh $host 'pgrep -a comdb2' | grep $dbnm | awk '{print $1}'`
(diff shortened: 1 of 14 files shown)


tests/mem_tracker.test/runit
```diff
@@ -6,6 +6,10 @@ set -e
 dbnm=$1
 host=`cdb2sql ${CDB2_OPTIONS} -s --tabs $dbnm default 'SELECT comdb2_host()'`
 
+# Get the base address of comdb2 binary in the address space of the db process
+dbpid=`pgrep -a comdb2 | grep $dbnm | awk '{print $1}'`
```
adizaimi: in clustered mode you don't have comdb2 executable running

riverszhang89: Right. Thank you!

riverszhang89:

PR #2046, opened 2020-01-30
Simplify code that computes clntstats hash entry

adizaimi:
Simplify code that computes clntstats hash entry


adizaimi:
I had put it to 0 for testing, that's why the tunable test had the diff. Will run it again for completeness.
Thanks

nirbhayc:
> I had put it to 0 for testing, that's why the tunable test had the diff. Will run it again for completeness.
> Thanks

Really, didn't notice that in the patch.

adizaimi:
> > I had put it to 0 for testing, that's why the tunable test had the diff. Will run it again for completeness.
> > Thanks
> 
> Really, didn't notice that in the patch.

Probably I pushed the reversal just before you opened the PR for viewing. Sorry, and thanks.

PR #2047, opened 2020-01-30
Member rqid is moved so use hash init with offset

adizaimi:
Moving rqid in structure osql_sqlthr so need to use hash init with offset hash_init_o().


adizaimi:
Thank you Akshat!

PR #2048, opened 2020-01-30
Sc lockorder apply

markhannum:
Lots of changes in this branch which address various schemachange issues.  I'm preparing more formal documentation about what was changed, and why.  I'm submitting this as a PR now so that it will run under RR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,6 +83,48 @@ add_definitions(
   -D_FILE_OFFSET_BITS=64
 )
 
+option(DEBUG_STACK_TAG_SCHEMA "Stack at add or remove a tag-schema" OFF)
+mark_as_advanced(DEBUG_STACK_TAG_SCHEMA)
+if(DEBUG_STACK_TAG_SCHEMA)
+    add_definitions(-DDEBUG_STACK_TAG_SCHEMA)
+endif()
+
+option(DEBUG_STACK_AT_TXN_LOG "Stack when writing a commit record" OFF)
+mark_as_advanced(DEBUG_STACK_AT_TXN_LOG)
+if(DEBUG_STACK_AT_TXN_LOG)
+    add_definitions(-DDEBUG_STACK_AT_TXN_LOG)
+endif()
+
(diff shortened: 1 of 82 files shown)


markhannum:
This is ready for review- I will be attaching some support documentation / description of some of these changes.

markhannum:
See text-files (stil in progress) in sc_review directory for info

markhannum:
Did a review with Nirbhay and Mike (thanks guys)!  Am going to list a few follow-up items, and I think folks are reviewing the code more carefully:

1. In backout_and_abort_tranddl, why do we call "trans_commit" if we've actually closed files rather than trans_abort?
Yes- the function "backout_schema_changes", called later in this function calls reload_db_tran, which calls load_new_ondisk.  This re-opens the old version of the tables under the same transaction.

2. In db/osqlcomm.c we get the schema-lock while holding db-level pagelocks .. this breaks the normal rule of always acquiring the pagelocks first, doesn't it?
Yes- we'd only be holding database level locks at this point if the transaction is running mixed ddl and dml.  The sc_truncate testcase has never tested this specifically.  This is an area of future work, and potentially a deep rabbit hole.  We've discussed the idea of either acquiring the schema-lock earlier in toblock, or forbidding mixed dml/ddl transactions.

3. Can we generalize the "assert-lk" functions so that they could be used with any rw-mutex, or any pthread_mutex?
Yes- great idea .. we would also like the ability to assert that a thread holds database-level locks.  I.e., assert that at this point in the code the schema-change thread holds a writelock on the table being schema-changed.  This is also an area of future work.

4. osql_postcommit_handle references iq->sc->db .. this db may no longer be there
Uhhh .. yes .. the call to bdb_lock_tablename_read should be moved up.  I will change that tonight.

5. Do we pass the correct bdb_state into wait_for_seqnum_from_all_int?
Yes- trans_commit_int was changed to pass in the parent table rather than the child table.

I will address also the things you guys have commented directly on.  Thanks for the help guys!

markhannum:
> For #4, the correct fix for this is to add the file to the del-files list before the commit record has been written (while the tablelock is still held by the transaction).

PR #2057, opened 2020-02-01
sqlitex plugin

nirbhayc:
(for testing only)


cdb2api/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ set(src
 
 list(APPEND src ${PROJECT_SOURCE_DIR}/util/walkback.c)
 set_source_files_properties(${PROJECT_SOURCE_DIR}/util/walkback.c PROPERTIES COMPILE_FLAGS -DUSE_UNWIND)
+set_source_files_properties(cdb2api.c PROPERTIES COMPILE_FLAGS "-Wno-stringop-overflow")
```
akshatsikarwar: https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html
Lets do it for just the offending section - not entire file.

PR #2058, opened 2020-02-03
Add fingerprint column to active connections virtual table (backport)

mistachkin:


db/db_fingerprint.c
```diff
@@ -53,26 +53,35 @@ int clear_fingerprints(void) {
...
+    MD5Context ctx = {0};
+
+    assert(zNormSql);
+    assert(pnNormSql);
+
+    *pnNormSql = strlen(zNormSql);
+
+    MD5Init(&ctx);
+    MD5Update(&ctx, (unsigned char *)zNormSql, *pnNormSql);
+    memset(fingerprint, 0, FINGERPRINTSZ);
```
mponomar: Don't think this needs a memset?  We're about to overwrite it.

mponomar: I realize existing code did that as well.

PR #2062, opened 2020-02-03
Port from r6 1097

dorinhogea:
Currently releasing locks during cursor moves can leave sqlite engine in a non-recoverable state.  Since it is safe to release the lock during row emitting phase, we can postpone lock release until the engine sends a row to the client.
The only case when this is a problem is when a join runs a heavy inner loop, taking a long time to send a row back.  We provide a configurable time delay default to 5 seconds to prevent this from becoming a problem.
In the targeted non-recoverable state, sqlite engine prints "Dta lookup failed" error, and client gets a 203 error code.


dorinhogea:
NOTE: there are two more cases that require further thought:
- releasing the locks during the shadow update in old snapshot/serial mode (current patch introduces an alert to detect when this happens)
- fetching blobs 
These cases will be the subject of future PR-s.

PR #2066, opened 2020-02-04
sqlite: register date functions correctly

nirbhayc:
No need to check thd to register date functions.


nirbhayc:
Cherry-picked to 7.0.

nirbhayc:
@mistachkin Thanks for having a look at the patch!

PR #2067, opened 2020-02-05
fix off by one when string for stack or taskname is empty

adizaimi:
fix off by one when string for stack or taskname is empty


adizaimi:
Thanks Joe.

PR #2068, opened 2020-02-05
Remove unused atomic include file

adizaimi:
Remove unused atomic include file


adizaimi:
Cherry picked into 7.0.
Thanks Akshat.

PR #2069, opened 2020-02-05
Properly cleanup nodestat entries

adizaimi:
Properly cleanup nodestat entries.


db/reqlog.c
```diff
@@ -2172,75 +2168,109 @@ static nodestats_t *add_clientstats(unsigned checksum,
...
+{
+    nodestats_t *entry = get_new_nodestats(checksum, task_and_stack, task_len,
+                                           stack_len, node, fd);
+    if (!entry) return NULL;
+    nodestats_t *entry_chk = NULL;
     Pthread_rwlock_wrlock(&clientstats_lk);
     {
         entry_chk = hash_find(clientstats, entry);
         if (entry_chk) {
-            free(entry);
```
adizaimi: The thing that would leak here is entry->rawtotals.lk from line 2175.

PR #2070, opened 2020-02-06
Fix smoketest

nirbhayc:
.

### DIFF
--- a/tests/smoketest/Dockerfile.nossl.build
+++ b/tests/smoketest/Dockerfile.nossl.build
@@ -16,6 +16,6 @@ COPY . /comdb2/
 
 WORKDIR /comdb2/smoketestbuild/
 
-RUN cmake3 .. -DWITH_SSL=0 && make && make install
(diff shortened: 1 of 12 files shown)


tests/smoketest/Dockerfile.nossl.build
```diff
@@ -16,6 +16,6 @@ COPY . /comdb2/
 
 WORKDIR /comdb2/smoketestbuild/
 
-RUN cmake3 .. -DWITH_SSL=0 && make && make install
+RUN cmake3 .. -DWITH_SSL=0 && make -j install
```
mponomar: -j is going to try to build everything at once.  That's going to fail on some places (eg: raspberry pi port).  Maybe limit to -j16?  Otherwise looks good.

akshatsikarwar: How much do we care about smoke test on raspberry pi?

mponomar: Oh, this is smoke test, sorry.  Ignore me, carry on.

PR #2072, opened 2020-02-07
Dont create file for in memory repdb 

adizaimi:
Introducing a new global to control behavior of inmemory repdb.
Currently we misuse another global for that purpose.
Also, don't open repdb file if this new global is set.


adizaimi:
Many thanks Mark.

PR #2078, opened 2020-02-14
Fix a crash with PRAGMA

nirbhayc:
Bug found and fixed suggested by @mistachkin. Thanks!


nirbhayc:
@mistachkin Thanks again!

PR #2080, opened 2020-02-17
Fix copycomdb2 for cases when lz4 is not available

adizaimi:
For cases when lz4 is not available, copycomdb2 fails with 
`bash: line 7: nolz4: command not found` example from test truncatesc.
This is because indeed the $lz4_rmt is set to `nolz4` which is an undefined function in the remote ssh session. This checkin sets the ar_compress and ar_compress_rmt variables to `cat` explicitly rather than use a `nolz4` as a bash function. 


adizaimi:
Thank you Mike

PR #2082, opened 2020-02-20
osql: fix an assertion failure

nirbhayc:
Problem: sess_impl.dispatched flag doesn't get set even after DONE message i received.


dorinhogea:
Please check https://github.com/bloomberg/comdb2/pull/2084 for a more complete version.  Session cannot call osql_repository_put ignoring is_msg_done flag, since early session termination (osql_sess_try_terminate()) could free the session before it gets dispatched.

PR #2083, opened 2020-02-20
Views: fix a crash when db has both user views and time partitions

nirbhayc:
Problem: Time partition implementation wrongly tried to delete triggers formed on user defined view names.


nirbhayc:
Thanks @dorinhogea !

PR #2084, opened 2020-02-20
fix assert

dorinhogea:
Make sure session is marked dispatched before releasing the clients counter, to prevent termination race.


dorinhogea:
thank you for review

PR #2085, opened 2020-02-20
Delay a deadlock retry based on toblock() run time

riverszhang89:
On a deadlock the block processor polls a random amount of time between 1 ms and 25 ms before retrying, resulting in an average delay of 12.5 ms which is too high for modern hardware.

The patch uses average run time of `toblock()` calls to determine the deadlock retry delay, which matches the workload more closely.

### DIFF
--- a/db/sltdbt.c
+++ b/db/sltdbt.c
@@ -16,6 +16,7 @@
 
 /* comdb index front end */
 
+#include <unistd.h> /* For usleep() */
 #include <ctype.h>
 #include <epochlib.h>
 #include "comdb2.h"
@@ -123,8 +124,8 @@ static void adjust_maxwthreadpenalty(int *totpen_p,
         penaltyinc = 1;
     }
 
-    if (penaltyinc + gbl_maxwthreadpenalty > gbl_maxthreads)
-        penaltyinc = gbl_maxthreads - gbl_maxwthreadpenalty;
+    if (penaltyinc + gbl_maxwthreadpenalty > gbl_maxwthreads)
+        penaltyinc = gbl_maxwthreads - gbl_maxwthreadpenalty;
 
     gbl_maxwthreadpenalty += penaltyinc;
     *totpen_p += penaltyinc;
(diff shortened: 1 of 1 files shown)


adizaimi:
A couple of thoughts:
* This seems like a great idea to play with the amount of time we delay, since it appears to increase performance. Ideally we would not have any deadlocks and then this would not be an issue--we should indeed look for any side effects, like increased number of deadlocks. 
* What is the observed value of avg_toblock_us for different workloads? What happens if we set it to just 1ms or just to retry immediately? 
* Maybe we should consider to make delay proportional to the number of active writer threads rather than gbl_maxwthreads.

riverszhang89:
@adizaimi Thank you for the comment.

Deadlocks are inevitable. I don't think you could eliminate all the deadlocks from all the workloads.

avg_toblock_us is around 200 microseocnds for oltp_write_only on a btda machine. You would want to pace deadlock retries accordingly to increase their chances of success. 1ms seems very random. Retrying immediately will likely end up with more deadlocks.

I tried using number of active writers. It did not make a difference.

db/sltdbt.c
```diff
@@ -169,7 +176,16 @@ static int handle_op_block(struct ireq *iq)
...
     }
 
-    if (rc == RC_INTERNAL_RETRY) {
+    if (rc == 0) {
+        /* Calculate average runtime of last 16 toblock calls, and cap it
+           at 25 milliseconds. The value is used to determine how much time we
+           sleep before retrying on a deadlock. It does not have to be 100%
+           accurate so we do not hold a lock here. We could make the number of
+           calls and the cap tunables but for now leave them hardcoded. */
+        avg_toblock_us += ((stopus - startus - avg_toblock_us) >> 4);
```
adizaimi: Since avg_toblock_us is global, maybe we should get a local copy here, perform your average and checks, and then update global.

db/sltdbt.c
```diff
@@ -169,7 +176,16 @@ static int handle_op_block(struct ireq *iq)
         rc = ERR_NOT_DURABLE;
     }
 
-    if (rc == RC_INTERNAL_RETRY) {
+    if (rc == 0) {
+        /* Calculate average runtime of last 16 toblock calls, and cap it
```
adizaimi: Also is the comment `average runtime of last 16 toblock calls` correct? It seems this is a more like an exponential moving average?

db/sltdbt.c
riverszhang89: It'd be too much a micro-optimization without an actual gain.

riverszhang89:
@adizaimi Thank you so much for the review!

PR #2088, opened 2020-02-24
Cherrypick of SQLite query optimizer enhancement for STAT4

mistachkin:

If STAT4 determines that a WHERE clause term that is not used by an index has very high probability of being true, then do not use that term to reduce the estimated output row count.

This is a cherrypick of [40739c793b] from upstream SQLite trunk.

/plugin-branch 7.0


mistachkin:
> 
> If upstream SQLite introduced any tests for this case, we should consider pulling them in as well.
> 

The new test has now been added, along with several enhancements to the YAST test infrastructure to allow it to pass.

mistachkin:
RoboRivers is fully passing and all RoboMark failures are confirmed false alarms.

PR #2089, opened 2020-02-25
get rid of dbenv_from_ireq

dorinhogea:

Simple change; there is really only one struct dbenv (thedb), and we don't really need this extra indirection


dorinhogea:
Thanks Mike!

PR #2090, opened 2020-02-25
Fixes in user authentication

nirbhayc:
* Fix username buffer size to 16 (limits to 15 characters)
* Update error message to not append password string when it exceeds the limit
* Add tests
* Update docs to show username/password limits


nirbhayc:
Cherry-picked to 7.0.

PR #2092, opened 2020-02-26
switch osqllog instrumentation  from files to output

dorinhogea:

I used this feature in the early days of writing osql.  It hasn't seen usage in a decade to my knowledge.   This PR changes the code  to save instrumentation lines to output instead of files.


dorinhogea:
Remove the mechanism for creating files and switch instrumentation to output trace LOGMSG_DEBUG, keeping the activation parameter enable_osql_logging

dorinhogea:
Thank you Mike

PR #2093, opened 2020-02-26
get rid of use_handle

dorinhogea:

Simpler code, remove indirection of iq.use_handle


dorinhogea:
thank you Adi!

PR #2097, opened 2020-02-28
remove ireq from early osql session structure

dorinhogea:

Delay postponing allocating a struct ireq until the bplog is complete.  The dispatch and failure is simpler, since we keep less state; the memory footprint per bplog is reduced.

### DIFF
--- a/db/osqlcomm.c
+++ b/db/osqlcomm.c
@@ -77,8 +77,11 @@ extern int gbl_goslow;
 extern int gbl_partial_indexes;
 
 extern int db_is_stopped();
-static int osql_net_type_to_net_uuid_type(int type);
 int gbl_toblock_random_deadlock_trans;
+
+static int osql_net_type_to_net_uuid_type(int type);
+static void osql_extract_snap_info(osql_sess_t *sess, void *rpl, int rpllen,
+                                   int is_uuid);
 
 typedef struct osql_blknds {
     char *nds[MAX_CLUSTER];        /* list of nodes to blackout in offloading */
@@ -3285,7 +3288,7 @@ int gbl_disable_cnonce_blkseq;
  *
  */
 int osql_comm_is_done(osql_sess_t *sess, int type, char *rpl, int rpllen,
-                      int hasuuid, struct errstat **xerr)
+                      int is_uuid, struct errstat **xerr)
 {
     int rc = 0;
     switch (type) {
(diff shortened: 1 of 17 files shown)


db/osqlcomm.c
```diff
@@ -7096,84 +7058,49 @@ void signal_replicant_error(const char *host, unsigned long long rqid,
...
+        flags = req.flags;
+        tzname = req.tzname;
+        sqllen = req.sqlqlen;
     }
 
     /* create the request */
-    sess = osql_sess_create(sqlret, sqllenret, req.tzname, type, req.rqid, uuid,
-                            fromhost, malcd, is_reorder_on);
+    sess = osql_sess_create(sql, sqllen, tzname, type, rqid, uuid, fromhost,
+                            flags & OSQL_FLAGS_REORDER_ON);
```
adizaimi: flags & OSQL_FLAGS_REORDER_ON should be set only in the case of uuid net type -- in case that flag is accidentally on for rqid type, we should simply ignore it. I think this caused issues in my testing for rqid messages -- if you see any issues in robomark, we might want to leave it the way the parameter was passed before.

db/osqlcomm.c
```diff
@@ -7096,84 +7058,49 @@ void signal_replicant_error(const char *host, unsigned long long rqid,
...
-        is_reorder_on = ((uuid_req.flags & OSQL_FLAGS_REORDER_ON) != 0);
+        osql_uuid_req_t ureq;
+        sql =
+            (char *)osqlcomm_req_uuid_type_get(&ureq, p_req_buf, p_req_buf_end);
+        rqid = OSQL_RQID_USE_UUID;
+        comdb2uuidcpy(uuid, ureq.uuid);
+        flags = ureq.flags;
+        tzname = ureq.tzname;
+        sqllen = ureq.sqlqlen;
     } else {
```
adizaimi: I understand this is out of the scope of this project, but we should consider to remove rqid code totally--once everyone is on r7, we should turn on uuid mode for all and rqid mode can not exist in r7.1. 
Thoughts?

dorinhogea: lets have this discussion in a proper ticket/issue

db/osqlcomm.c
dorinhogea: it is set on replicant; bplog has no knowledge of the protocol replicant use when creating the bplog.  ordering the bplog indexes using different criteria's is still on ordering, and as long as it is global, it is serving the same purpose of serializing the writers with the purpose of avoiding deadlocks.

dorinhogea:
Thank you Adi!

PR #2098, opened 2020-02-28
Max blob field size

riverszhang89:
Max blob field size should be 268435455 (2^28 - 1).

(DRQS 153686908)


riverszhang89:
@mponomar reviewed it internally. Thanks!

PR #2100, opened 2020-03-03
Backport the comdb2_user() SQL function from master.

mistachkin:

This function is needed by the checked-in version of the 'userschema' test on this branch.


mistachkin:
> 
> There are some related tests under auth.test and userschema.test. Could you backport them as well?
> 

It appears the 7.0 branch already has the related 'userschema' tests?  I'm backporting the related 'auth' tests now.

nirbhayc:
@mistachkin Could you add DCO and merge it please?

PR #2101, opened 2020-03-03
Fix remote policy logic to be consistent with that in older versions

nirbhayc:

/plugin-branch comdb2-rmtpol-fix


nirbhayc:
https://github.com/bloomberg/comdb2/pull/2105 was up-merged to master.

PR #2104, opened 2020-03-04
Use DEBUG level for rqid/uuid trace.

riverszhang89:
The write slowdown we've seen with last commit is probably caused by the incorrect log level. Let's see whether sysbench agrees.


adizaimi:
Good catch!

PR #2105, opened 2020-03-04
Fix remote policy logic to be consistent with that in older versions (7.0)

nirbhayc:
/plugin-branch comdb2-rmtpol-fix-7.0


dorinhogea:
what is the difference between machine_class and machine_my_class?

nirbhayc:
> what is the difference between machine_class and machine_my_class?

Latter returns class of the current node.

PR #2108, opened 2020-03-04
fdb/sqlitex: leaking connection (7.0)

nirbhayc:
Due to struct mismatch (sqlite3 vs sqlitex) fdb_master_is_local was returning incorrect result after reading invalid memory. Fixed by caching sqlitex struct in BtCursor (alongside sqlite3)
and adding accessor methods to query_preparer plugin to return correct values out of the cached object.

/plugin-branch comdb2-sqlitex-conn-leak-7.0

### DIFF
--- a/bbinc/comdb2_query_preparer.h
+++ b/bbinc/comdb2_query_preparer.h
@@ -1,5 +1,5 @@
 /*
-   Copyright 2019 Bloomberg Finance L.P.
+   Copyright 2019, 2020 Bloomberg Finance L.P.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


dorinhogea:
After merging sqlitex plugin, we do duplicate remote schema collection.  This is inline, and expensive.  Since all the schema is available local, the remote access in second prepare should be removed; instead, the schema should be read from local cache, already populated by the initial prepare.  
Please check fdb_master_is_local(),  sqlite.init.busy should be 1 during schema preparation, in the secondary sqlite engine it is not set (forcing a remote access).

nirbhayc:
@dorinhogea : Thanks for the hint. I have rebased and pushed a commit that I believe fixes the issue.

nirbhayc:
@dorinhogea Thanks again!

PR #2109, opened 2020-03-05
Fix local replication for vutf8 and blob2 fields

mohitkhullar:
Fix local replication for vutf8 and blob2 fields


nirbhayc:
What was the issue? CONVERT_IGNORE_BLOBS flag is to ignore blob/vutf8 conversion failures?

mohitkhullar:
We don't pass in blob/vutf8 data to convert so its never converted, we pass NULL instead. The issue is was with local replication which stopped working for tables with vutf8/blobs in it.

mponomar:
Ah, ok.  Otherwise the value comes from here:

https://github.com/bloomberg/comdb2/pull/2109/files#diff-20b8ebf66c0ff339c2b20563d1f0d548L530

That looks correct.

PR #2110, opened 2020-03-05
Stop just request threads for old-style queue changes.

mponomar:
This adds a less severe stop_threads that only stops request threads, not all database threads.

### DIFF
--- a/bdb/fstdump.c
+++ b/bdb/fstdump.c
@@ -392,12 +392,12 @@ static void *fstdump_thread_inner(fstdump_per_thread_t *fstdump, void *sendrec,
 
         if (db_is_stopped()) {
             logmsg(LOGMSG_ERROR, "fstdump_thread: "
-                            "aborting due to stop_threads\n");
+                            "aborting due to stop_all_threads\n");
             Pthread_mutex_lock(&common->lock);
             {
                 common->bdberr = BDBERR_DEADLOCK;
                 snprintf0(common->errmsg, sizeof(common->errmsg),
-                          "aborted because database stop_threads");
+                          "aborted because database stop_all_threads");
             }
             Pthread_mutex_unlock(&common->lock);
             dbcp->c_close(dbcp);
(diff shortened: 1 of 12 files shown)


db/thrman.c
```diff
@@ -51,6 +51,8 @@
 #include "reqlog.h"
 #include "str0.h"
 
+#include <unistd.h>
```
adizaimi: leftover?

db/handle_buf.c
```diff
@@ -915,7 +915,7 @@ int handle_buf_main2(struct dbenv *dbenv, SBUF2 *sb, const uint8_t *p_buf,
     int numwriterthreads;
     struct dbq_entry_t *newent = NULL;
 
-    if (db_is_stopped()) {
+    if (db_requests_are_stopped()) {
```
adizaimi: if (db_is_stopped() || db_requests_are_stopped()) ?

db/thrman.c
mponomar: Yes, I had a sleep in there earlier to try to induce a race.

db/handle_buf.c
mponomar: Sure.  req_thds should all be gone at this point by virtue of calling stop_threads, but I can add that regardless.

bdb/fstdump.c
```diff
@@ -392,12 +392,12 @@ static void *fstdump_thread_inner(fstdump_per_thread_t *fstdump, void *sendrec,
 
         if (db_is_stopped()) {
             logmsg(LOGMSG_ERROR, "fstdump_thread: "
-                            "aborting due to stop_threads\n");
+                            "aborting due to stop_all_threads\n");
```
nirbhayc: "aborting due to database shutting down" instead?

nirbhayc: Ditto for all other similar changes.

db/comdb2.c
```diff
@@ -1102,8 +1107,7 @@ static void *purge_old_blkseq_thread(void *arg)
     loop = 0;
     sleep(1);
 
-    while (!db_is_stopped()) {
-
+    while (!db_is_stopped() && !db_requests_are_stopped()) {
```
nirbhayc: Setting requests_stopped would now cause purge_old_blkseq thread to exit forever. I don't think that intended.

db/thrman.c
```diff
@@ -471,7 +471,7 @@ static int thrman_check_threads_stopped_ll(void *context)
         all_gone = 1;
 
     /* if we're exiting then we don't want a schema change thread running */
-    if (db_is_stopped() && 0 != thr_type_counts[THRTYPE_SCHEMACHANGE])
```
nirbhayc: why?

db/comdb2.c
mponomar: It exits, but it's restarted in resume_threads.  Same as purge_old_files_thread.  In fact, that needs the same check.

db/thrman.c
mponomar: This logic didn't change as part of this PR, but I also question it.

mponomar:
#2252 is a better version of this.

PR #2112, opened 2020-03-06
uprec business logic relocation

dorinhogea:

A bit of osqlcomm spring cleanup incoming; relocate uprec logic in its own unit.


dorinhogea:
Thanks Adi!

PR #2111, opened 2020-03-05
remove duplicated code

dorinhogea:
Signed-off-by: Dorin Hogea <dhogea@bloomberg.net>

Remove duplicate offload_net_send code.

PR #2113, opened 2020-03-09
removed duplicated code, take 2

dorinhogea:

Repeat https://github.com/bloomberg/comdb2/pull/2111, with a fix and less code.


dorinhogea:
thanks Mike

PR #2114, opened 2020-03-10
Check a sequence's committed gen only if elect_highest_committed_gen …

markhannum:
This PR fixes an issue with the enable_seqnum_generations tunable: got_new_seqnum_from_node refuses to change a node's coherency if the most recent committed generation is not equal to the master's current generation.  The bug is that the most recent committed generation is set only if the berkdb attribute, "elect_highest_committed_gen" is enabled.  Straightforward fix: check a sequence number's committed_gen only if elect_highest_committed_gen is enabled.


markhannum:
> I'm working on master-swings in a separate branch

PR #2117, opened 2020-03-11
Soften asserts, be more permissive in distributed commit

markhannum:
Simple changes that soften the schema-lk asserts- there are cases where we acquire the schema lock in read mode recursively.


markhannum:
sc_truncate failure isn't related to this pr (I'm running this in a loop now & will put in a different pr to address).

PR #2120, opened 2020-03-11
fix test inconsistencies: dbnm not set and reorder_inserts threshold

adizaimi:
fix test inconsistencies: dbnm not set and reorder_inserts threshold


adizaimi:
Thank you Mike!

PR #2121, opened 2020-03-12
Provide queues with odh for fast depth calculation

markhannum:
Provide queues the ability to use and utilize ondisk-headers.  This gives us compression and version numbers automatically (as part of the odh structure).  I've added a 'sequence' to the queue object which will be monotonically increasing, and allows us to instantly calculate a queue's depth by retrieving only the first and final record rather than by walking the entire queue.  I'm putting in this PR early to get eyes on this.


plugins/dbqueuedb/dbqueuedb.c
```diff
@@ -501,6 +501,25 @@ static void admin(struct dbenv *dbenv, int type)
...
+{
+    struct consumer_stat *stats = userptr;
+
+    if (consumern < 0 || consumern >= MAXCONSUMERS) {
+        logmsg(LOGMSG_USER, "%s: consumern=%d length=%u epoch=%u\n",
+                __func__, consumern, (unsigned)length, epoch);
+    } else {
+        assert(!stats[consumern].has_stuff);
+        stats[consumern].has_stuff = 1;
+        stats[consumern].first_item_length = length;
```
mponomar: This needs `if (!stats[consumern].has_stuff)` - it's meant to latch the first value and not override it again.  Then again, we don't need to walk anymore with this code, we can just pull out the first and last sequence numbers in get_stats.

markhannum: No- it doesn't walk anymore- it pulls out the first and the last values, filling the original structure with the information.  See how this is called in stat_thread_int->dbq_odh_stats->bdb_queuedb_stats (line 397)

mponomar: Agreed - misread the code.

bdb/queuedb.c
```diff
@@ -97,32 +97,127 @@ int bdb_queuedb_best_pagesize(int avg_item_sz)
...
+     * serialized on the last page.  Grabbing this lock before getting a genid
+     * prevents genids from being added out-of-order.
+     *
+     * TODO: rather than grabbing inline, minimize the time we hold this lock by
+     * deferring queue-writes until after everything else in toblock is done.
+     * Also force all transactions to acquire these locks in the same order so
+     * that we can avoid deadlocks from two transactions which have both made
+     * it to that point. */
+
+    rc = bdb_lock_table_write(bdb_state, tran);
```
akshatsikarwar: Use different lock so we don't block consumer

markhannum: Use a different lock so we don't block out readers

bdb/queuedb.c
```diff
@@ -97,32 +97,127 @@ int bdb_queuedb_best_pagesize(int avg_item_sz)
...
+        dbt_key.ulen = QUEUEDB_KEY_LEN;
+        dbt_key.flags = DB_DBT_USERMEM;
+
+        dbt_data.data = qfnd_seq_fnd;
+        dbt_data.flags = DB_DBT_MALLOC;
+
+        /* Probe last sequence lock page */
+        rc = bdb_cget_unpack(bdb_state, dbcp, &dbt_key, &dbt_data, &ver,
+                DB_LAST|DB_RMW);
+        qfnd_seq.seq = 0;
```
markhannum: Get genid here

bdb/queuedb.c
```diff
@@ -203,6 +425,7 @@ int bdb_queuedb_walk(bdb_state_type *bdb_state, int flags, void *lastitem,
 {
```
markhannum: Revert this- if odh is enabled then you shouldn't get in here

schemachange/sc_queues.c
```diff
@@ -328,6 +377,27 @@ static int perform_trigger_update_int(struct schema_change_type *sc)
...
+
+    rc = bdb_lock_tablename_write(thedb->bdb_env, sc->tablename, tran);
+    if (rc) {
+        sbuf2printf(sb, "!Error %d getting tablelock for %s.\n",
+                rc, sc->tablename);
+        sbuf2printf(sb, "FAILED\n");
+        goto done;
+    }
+
     db = get_dbtable_by_name(sc->tablename);
```
markhannum: These should use the transaction

markhannum: >> just a hash table

schemachange/sc_queues.c
```diff
@@ -165,19 +166,35 @@ int add_queue_to_environment(char *table, int avgitemsz, int pagesize)
...
+    tran = bdb_tran_begin(thedb->bdb_env, NULL, &bdberr);
+    if (tran == NULL) {
+        logmsg(LOGMSG_ERROR, "%s:%d can't begin transaction rc %d\n", __FILE__,
+                __LINE__, bdberr);
+        rc = bdberr;
+        goto done;
+    }
+
+    bdb_get_tran_lockerid(tran, &lid);
+    bdb_set_tran_lockerid(tran, gbl_rep_lockid);
```
markhannum: Make sure this works correctly if the feature is disabled

bdb/queuedb.c
markhannum: > Changed my mind about this- it's harmless to keep it this way, and there may be a scenario in the future where we'd like to walk an odh queue ..

PR #2122, opened 2020-03-12
dbq_get: Obtain genid from cursor, not payload

akshatsikarwar:


akshatsikarwar:
The genid in payload is of questionable endianess. Use genid from key and pass untouched to master for consume. Master (7.1, 8.0?) will stop trying to paper over this during consume and will expect to receive proper genid.

PR #2123, opened 2020-03-12
Some scripts to make development easier

mponomar:
Provides a 'cluster' script to create a cluster of comdb2 containers.

### DIFF
--- a/contrib/dev-util/Makefile
+++ b/contrib/dev-util/Makefile
@@ -1,0 +1,13 @@
+OBJS=testrunner.o #plhash.o pool.o
+BINARY=testrunner
+
+CFLAGS=-pthread -Wall -g -Wextra -Wno-unused-parameter
+
+$(BINARY): $(OBJS)
+	gcc -o $(BINARY) $(OBJS) $(LDFLAGS)
+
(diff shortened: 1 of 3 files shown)


akshatsikarwar:
init should also build?

akshatsikarwar:
Error for missing `proxy.conf` should use `${dir}`

`${dir}` should be `comdb2` or `comdb2-util`?

akshatsikarwar:
`cluster build` failed with:
Step 33/35 : COPY cluster    /home/asikarw1/bin/cluster
COPY failed: stat /var/lib/docker/tmp/docker-builder615846446/cluster: no such file or directory

mponomar:
> Error message for missing `proxy.conf` should use `${dir}`
> 
> `${dir}` should be `comdb2` or `comdb2-util`?

`${dir}` is `comdb2-util`.  Script manages `comdb2-util`, which is transient.  `comdb2` is user settings, managed by user, and is persistent (we never write to it).

mponomar:
> `init` should also `build` - no need for `build` anymore really.

Good point! I'll change that.

PR #2124, opened 2020-03-12
Fix local replication for vutf8 and blob fields

mohitkhullar:
Fix local replication for vutf8 and blob fields


mohitkhullar:
approved here https://github.com/bloomberg/comdb2/pull/2109#issuecomment-595296593

PR #2125, opened 2020-03-12
Use the static table rather than dbs[0] in process_message

markhannum:
Patch up several places in process_message which dereference dbs[0], but instead use "dbenv->static_table".


nirbhayc:
This one's a subset of https://github.com/bloomberg/comdb2/pull/2128 ?

markhannum:
yes - withdrawing

PR #2129, opened 2020-03-13
Fix static table logic in R7

markhannum:
We allow databases to be initialized with no tables in R7, utilizing instead a static_table.  This PR addresses a few cases which "slipped through".  I noticed these issues while building with Bloomberg specific hooks, and then attempting to run some of our open-source tests.


markhannum:
Reviewed under https://github.com/bloomberg/comdb2/pull/2125

PR #2131, opened 2020-03-13
consolidate polling on bdb_lock_desired

dorinhogea:

Consolidated C&P code that pause sql during recover_deadlock, if bdb lock is desired.


dorinhogea:
Thank you Adi!

PR #2142, opened 2020-03-18
Stop tests after n failures

adizaimi:
Stop tests after n failures -- will simply not run the tests from testrunner.

### DIFF
--- a/tests/runtestcase
+++ b/tests/runtestcase
@@ -247,6 +247,16 @@ verify_db_at_finish() {
     done
 }
 
+stop_if_more_than_n_failed () {
+    N=10   # if more that N failures in test.log stop running
+    F=`grep failed ${TESTDIR}/test.log | wc -l`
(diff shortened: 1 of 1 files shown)


akshatsikarwar:
Still too many. 5 or 10 max I think.

mponomar:
My only objection is that this makes it difficult to tell between the cases that failed and the cases that refused to run because they hit the limit.  As long as we have some way to tell from the log, this is good to go.  Also agree with Akshat about making that 10.

adizaimi:
I think it is easy to identify the failed tests because `grep failed test.log` will not include the tests that were not run.
Note that I was trying to find a better way, but make does not allow me to stop after N failures (we could run with Ninja which supports -k N, or a scripting language).

adizaimi:
Thanks for looking.
We can investigate ways to better stop this via ninja or cmake in future.

PR #2143, opened 2020-03-19
Provide persistent queue sequence numbers, expose sequence and epoch

markhannum:
Provide a tunable (enabled by default, disabled by legacy defaults) which preserves and exposes sequence-numbers of queue items (which acts as a global counter of the number of items that have been queued), and exposes the epoch field to lua consumers.


markhannum:
Trying to reproduce the insert_lots and logfill failures against both this branch and master .. neither of these use queues, so I suspect that these failures are unrelated to this pr

markhannum:
Code review changes: make this opt-in at schema-change time, and provide a column in comdb2_triggers which shows whether this is enabled or disabled.  Also, cleanup the variable names in bdb_queuedb_add.

akshatsikarwar:
Decide during DDL time to get feature or not.

akshatsikarwar:
We discussed offline - Mark noted already above.

db/glue.c
```diff
@@ -4477,8 +4529,19 @@ get_put_db(queue_odh, META_QUEUE_ODH)
 // get_db_queue_compress, get_db_queue_compress_tran, put_db_queue_compress
 get_put_db(queue_compress, META_QUEUE_COMPRESS)
 
+<<<<<<< 325f92132da1e50f0f492eca2c713e8d3bf14fdc
```
adizaimi: some bad merge lines here

adizaimi:
The diff has a ton of doc gif files -- it seems this PR only changed one of the gif files, can we remove all the other gifs from the PR please?

markhannum:
Tx Adi - unneeded gifs removed, bad merge fixed.

markhannum:
I think all review items have been addressed now

PR #2144, opened 2020-03-19
Quick-fix for sequences

markhannum:
Set an initial sequence to 0 rather than whatever happened to be on the stack.  I didn't notice this before because the stat command would always return the correct answer (subsequent sequence numbers just look at the previous and add one).  PR https://github.com/bloomberg/comdb2/pull/2143 providing persistent sequence numbers- contains the same fix.

### DIFF
--- a/bdb/queuedb.c
+++ b/bdb/queuedb.c
@@ -181,7 +181,7 @@ int bdb_queuedb_add(bdb_state_type *bdb_state, tran_type *tran, const void *dta,
             qfnd_seq.epoch = tran->trigger_epoch = comdb2_time_epoch();
         }
 
-        qfnd_seq.seq = 0;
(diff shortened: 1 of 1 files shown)


bdb/queuedb.c
```diff
@@ -181,7 +181,7 @@ int bdb_queuedb_add(bdb_state_type *bdb_state, tran_type *tran, const void *dta,
             qfnd_seq.epoch = tran->trigger_epoch = comdb2_time_epoch();
         }
 
-        qfnd_seq.seq = 0;
+        qfnd_fnd.seq = 0;
```
adizaimi: This looks correct, without this, qfnd_fnd.seq is garbage and line 198 assignment will set qfnd_seq.seq to garbag.

PR #2145, opened 2020-03-19
isolate osql prefetch in its own unit

dorinhogea:

Osql prefetch thread pool separation.


dorinhogea:
Thank you!  I don't wanna create at this point a internal header file, especially if we might remove this code altogether following a performance evaluation.

PR #2148, opened 2020-03-21
commit

markhannum:
Not a real PR- submitting so that we can run ROBO-NIRBHAY with reordering disabled


adizaimi:
FTR, reordering idx operations deferred project in https://github.com/bloomberg/comdb2/pull/1810 showed that the only slowdown in roboperf was sysbench/oltp_read_write test and upon investigation we found that it was due to increased number of deadlocks with sql threads which get idx locks first then dta file locks (exact opposite with master lock schedule). 
Mark is trying to invert the masterlock schedule in https://github.com/bloomberg/comdb2/pull/2149.

PR #2149, opened 2020-03-21
Emit index locks before datafile locks

markhannum:
If the log_index_locks_first tunable is enabled, do a "first-pass" of a transaction's held locks, collecting locks against index btrees.


markhannum:
This is ready for review

berkdb/build/db.h
```diff
@@ -241,6 +241,7 @@ struct __db_trigger_subscription;
 #define	DB_REP_CREATE	      0x0000001	/* Open of an internal rep database. */
 #define	DB_XA_CREATE	      0x0000002	/* Open in an XA environment. */
 #define DB_INDEX_CREATE       0x0000004 /* Create with index priority. */
+#define DB_INDEX_PRIORITY     0x0000008 /* Create with index priority. */
```
mponomar: Same comment as above flag?

berkdb/lock/lock.c
```diff
@@ -2862,8 +2865,13 @@ expired:			obj_ndx = sh_obj->index;
 
 	if (is_pagelock(sh_obj))
 		sh_locker->npagelocks++;
-	if (IS_WRITELOCK(newl->mode))
+	if (IS_WRITELOCK(newl->mode)) {
```
mponomar: Should we always set priority to 0 for non-write locks?

PR #2150, opened 2020-03-23
Fix pmux copying to replicant nodes correctly

adizaimi:
Fix pmux copying to replicant nodes correctly.
Directory was not created so in my containers pmux would not copy at all resulting in failed cluntered tests.


adizaimi:
Thank you Akshat.

PR #2151, opened 2020-03-24
do not send propagate error on rollback

dorinhogea:

Regression on rollback in single mode (the optimization to skip all bplog processing when OSQL_XERR got merged with common error path).


dorinhogea:
Thanks Mike!
Failure is captured by smoke test.  We might wanna consider running the regressions also in single node (currently only runs in clustered).

PR #2152, opened 2020-03-24
Add rollback test, correctly return rc

adizaimi:
- Add rollback checking output to basic.test
    - in osql_sess_rcvop() perr can be uninitialized so we should not derefetence
      it if is_msg_done is false, thus on failure only return rc
    - add ability to run test in single node even when CLUSTER is specified
      by just having a file singlenode.testopts in test directory.

### DIFF
--- a/db/comdb2.h
+++ b/db/comdb2.h
@@ -385,21 +385,20 @@ enum RCODES {
     ERR_NO_SUCH_TABLE = 310,  /* operation tried to use non-existant table */
     ERR_CALLBACK = 311,       /* operation failed due to errors in callback */
     ERR_TRAN_FAILED = 312,    /* could not start of finish transaction */
-    ERR_CONSTR =
-        313, /* could not complete the operation becouse of constraints in the
-                table */
-    ERR_SC_COMMIT =
-        314, /* schema change in its final stages; proxy should retry */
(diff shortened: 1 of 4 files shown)


adizaimi:
> To reproduce the issue, the server needs to run single-node. The test does not reproduce the issue
Dorin, the test that added here did reproduce the issue by virtue that I tested locally with master at the time. What you are proposing is to run basic.test in single node as well as clustered, and that can be added here.

adizaimi:
> I am pretty sure that osql_sess_rcvop needs to return rc, instead of 0.

I was fairly certain too -- I will fix.

dorinhogea:
> > To reproduce the issue, the server needs to run single-node. The test does not reproduce the issue
> 
> Dorin, the test that added here did reproduce the issue by virtue that I tested locally with master at the time. What you are proposing is to run basic.test in single node as well as clustered, and that can be added here.

The rollback is already tested in quite a few places.  The only thing needed was the ability to run regression tests single node as well.

adizaimi:
> I still strongly suggest that we should run most of regression tests in single node as well as clustered.

Oh, I agree. That's a bit more involved in deciding the details of how we should do that. 
For now this contains method to run any test as a testopt in single node.

PR #2153, opened 2020-03-24
Put SEQUENCE on the fallback list, fix comdb2sys test

markhannum:
The "SEQUENCE" token was recently introduced into the SQLite parser- it is used when creating a lua trigger to specify that the underlying queue should maintain persistent sequence numbers.  This change adds SEQUENCE to the fallback list- meaning that it no longer a reserved keyword: the SQLite parser will assume it's an "ID" if it's used in a different context.


markhannum:
Thanks Nirbhay!

PR #2156, opened 2020-03-25
Revert the reorder-index-locks code to see if this caused robo-nirbha…

markhannum:
We think this is the root-cause of Robo-Nirbhay failures.  Revert this pr.


markhannum:
Gives reasonable sysbench numbers again ..

PR #2162, opened 2020-03-25
use sc_set_running function definition from .h file

adizaimi:
use sc_set_running function definition from .h file
using this pr to fix internal bulk_import test too.

/plugin-branch fix_sc_set_running


adizaimi:
Thank you.

PR #2167, opened 2020-03-27
Keep all of a transactions queue operations in a temptable

markhannum:
Enqueuing an item to a queuedb type queue will block all other transactions attempting to add to the same queue.  This pull requests attempts to minimize the amount of time that a transaction will maintain a lock on the queue by deferring all of the queue processing to the end of the transaction.


markhannum:
Approved by Akshat and Joe ..

markhannum:
This was merged accidentally and reverted - i will create another pr ..

PR #2168, opened 2020-03-27
Don't leak the schema-lk from alter table

markhannum:
Fix bug in a recent checkin where an alter-table can leak the schema lock.  This bug was causing the yast test to fail.


markhannum:
Approved by Akshat and Joe

PR #2173, opened 2020-03-30
Use Canonical Name everywhere

riverszhang89:
If the master node has a Canonical Name different from its local hostname, and its Canonical Name is written in `hostname.lrl`, it'll think that the master is down and then call for an election, repeatedly.

This is because `gbl_myhostname`, which is the local hostname, is used in the `net/event` layer to check for localhost, whereas `gbl_node`, which is the host's Canonical Name, is used to create a `net`. Hence if `gbl_myhostname` and `gbl_node` are different, the `net/event` layer will fail to filter out net messages from localhost to localhost.

The patch gets rid of `gbl_myhostname`, and uses `gbl_mynode` everywhere.

### DIFF
--- a/bbinc/bb_oscompat.h
+++ b/bbinc/bb_oscompat.h
@@ -25,12 +25,13 @@ extern "C" {
 
 struct in_addr;
 
-typedef int(hostbyname)(char **, struct in_addr *);
+typedef int(hostbyname)(char **, struct in_addr *, char **);
 
 hostbyname *get_os_hostbyname();
 void set_hostbyname(hostbyname *);
 
-hostbyname comdb2_gethostbyname;
+int comdb2_gethostbyname(char **name, struct in_addr *addr);
+char *comdb2_getcanonicalname(char *name);
 void comdb2_getservbyname(const char *, const char *, short *);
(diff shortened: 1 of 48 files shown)


riverszhang89:
The PR also fixes the smoketest failure: 
`Master node downgrading - new master will resume schemachange`

It's caused by the unnecessary master downgrade too.

akshatsikarwar:
I think idea was `gbl_mynode` would be the node number (as string) for legacy uses (not sure where it is used though). If it is not used, this is fine.
If it still has some utility, we can use gbl_myhostname everywhere and leave gbl_mynode only for legacy uses.

akshatsikarwar:
Do test 'send bdb cluster', 'sanc' etc in internal build. DBA scripts depend on those being node numbers.

riverszhang89:
[`gbl_mynodeid`](https://github.com/bloomberg/comdb2/blob/master/db/comdb2.c#L304) stores the node number. I've verified the output of `bdb cluster`, `bdb sanc`, etc. I also changed `gbl_mynode` to `gbl_myhostname` per your suggestion.

riverszhang89:
@akshatsikarwar Thank you so much for the code review!

PR #2174, opened 2020-03-30
Fix crashes in solaris because atomic inc needs a 32 bit int

adizaimi:
Fix crashes in solaris because atomic inc needs a 32 bit int


akshatsikarwar:
Did sun compiler give warning before fix?

adizaimi:
Heh, if one would have looked, there would be warnings to see: 
"/home/azaimi/mbig2099/comdb2/bulkimport/bdb/bdb_verify.c", line 214: warning: argument #1 is incompatible with prototype:
Thanks for looking at this.

PR #2175, opened 2020-03-30
Fix upsert to work with DATACOPY

nirbhayc:


nirbhayc:
Review by @mponomar

PR #2179, opened 2020-03-31
handle new OP_IfNoHope opcode, and more redundant bug protections

dorinhogea:

Interesting enough, this sqlite tidbit was blocking replication when a schema change occurred!
New opcode, OP_IfNoHope, not handled, error ignored, both remote and source end up waiting for each other, hogging the bdb write lock.  Simple fix, added redundant checks for future sqlite upgrades (send error when we break sqlite generation, also stop fast when new opcodes show up).


dorinhogea:
Thank you Joe!

PR #2180, opened 2020-04-01
Ufid log nolimbo

markhannum:
Combines the ufid_log changes with the nolimbo changes, and enables the ufid_log and omit_dbreg tunables.  The ufid_log changes simplifies the log-stream format by emitting records which contain a btree's full 20-byte file-id rather than a 4 byte dbreg.  We instead keep track of all btree ufid's in a local hash that's populated from the database's working directory at startup.


markhannum:
This hung robo-rivers "reproducibly" at the end of a test, cdb2sql was blocked on a database task .. both cdb2sql and the task showed the connection as being "ESTABLISHED" .. unfortunately, I upset the state before getting a core of the database.  I'm resubmitting in hopes of getting this to occur again.

PR #2178, opened 2020-03-31
to not use sqlitex if remsql backend

dorinhogea:
Signed-off-by: Dorin Hogea <dhogea@bloomberg.net>

Remove double prepare for backend sql, it is not facing the client

PR #2181, opened 2020-04-01
to not use sqlitex if remsql backend

dorinhogea:

Port bf from master


dorinhogea:
https://github.com/bloomberg/comdb2/pull/2178

PR #2182, opened 2020-04-02
sqlitex: fix memory leaks

nirbhayc:
.


nirbhayc:
Merged to master branch.

PR #2187, opened 2020-04-05
Show running SC seed when sending stat to the db

adizaimi:
While there is a schema change running, `@send stat` no longer
displays a seed (accidentally removed). This patch will enable
to show running SC seed when sending stat to the db, and also adds
a check in testcase that comdb2_sc_status table has the same seed.


adizaimi:
Thank you for looking.

PR #2195, opened 2020-04-10
Enable master to return effects (7.0)

nirbhayc:


nirbhayc:
@markhannum Thanks for the review! I'll have also have them stored in blkseq.

PR #2196, opened 2020-04-13
Fix for live vutf8 schema change (7.0)

riverszhang89:
This is an edge case when extending inline size of a vutf8 field:

Let P and Q denote the inline size in the old and new schemas, respectively. Stripe S finishes converting. A new record now lands on stripe S. It must be written to both old and new tables as the schema change hasn't completed yet. Let X denote the length of the vutf8 field in the new record.

If `P < X <= Q`, when `osql_odh_blob` is on, the insert would fail because `utf8_validate()` would be examining wrong data and the schema change would subsequently abort; when it is off, the database would crash in an invalid `free`.

The bug was introduced by blob preprocessing (#1399). The fix is simple: If the old or new type of a field is vutf8, we unodhfy the payload of any new records during the schema change.

(DRQS 155195897)

### DIFF
--- a/db/db_tunables.h
+++ b/db/db_tunables.h
@@ -1826,5 +1826,4 @@ REGISTER_TUNABLE("debug_queuedb",
                  "(Default: off)",
                  TUNABLE_BOOLEAN, &gbl_debug_queuedb, EXPERIMENTAL, NULL, NULL,
                  NULL, NULL);
-
 #endif /* _DB_TUNABLES_H */
--- a/schemachange/schemachange.c
+++ b/schemachange/schemachange.c
@@ -671,25 +671,55 @@ int live_sc_post_delete(struct ireq *iq, void *trans, unsigned long long genid,
     return rc;
 }
 
+/* If the schema change is to 1) remove ODH, 2) change the compression
+   algorithm, 3) or alter a field from or to vutf8, unpack the blobs. */
 static int unodhfy_if_necessary(struct ireq *iq, blob_buffer_t *blobs,
                                 int maxblobs)
 {
-    int i, rc, oldodh, newodh, reccompr, oldcompr, newcompr;
+    int i, rc, oldodh, newodh, reccompr, oldcompr, newcompr, fromidx, blobidx;
+    struct schema *from, *to;
 
-    for (i = 0, rc = 0; i != maxblobs; ++i) {
-        bdb_get_compr_flags(iq->usedb->sc_from->handle, &oldodh, &reccompr,
-                            &oldcompr);
-        bdb_get_compr_flags(iq->usedb->sc_to->handle, &newodh, &reccompr,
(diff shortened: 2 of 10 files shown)


schemachange/schemachange.c
```diff
@@ -671,27 +671,57 @@ int live_sc_post_delete(struct ireq *iq, void *trans, unsigned long long genid,
...
-           unpack the blobs. */
-        if ((oldodh && !newodh) || oldcompr != newcompr) {
+    int i, rc, oldodh, newodh, reccompr, oldcompr, newcompr, fromidx, blobidx;
+    struct schema *from, *to;
+
+    bdb_get_compr_flags(iq->usedb->sc_from->handle, &oldodh, &reccompr,
+                        &oldcompr);
+    bdb_get_compr_flags(iq->usedb->sc_to->handle, &newodh, &reccompr,
+                        &newcompr);
+    (void)reccompr;
```
nirbhayc, line 686: Not your change, but, what's this for?

db/db_tunables.h
```diff
@@ -1821,5 +1821,10 @@ REGISTER_TUNABLE("debug_queuedb",
                  "(Default: off)",
                  TUNABLE_BOOLEAN, &gbl_debug_queuedb, EXPERIMENTAL, NULL, NULL,
                  NULL, NULL);
-
+REGISTER_TUNABLE("scconvert_finish_delay",
+                 "Delay returning from convert_record when a stripe finishes. "
+                 "This would create a scenario where scgenids are on the right "
+                 "of any new genids to reproduce a vutf8 schema change bug. ",
+                 TUNABLE_BOOLEAN, &gbl_scconvert_finish_delay,
+                 EXPERIMENTAL | INTERNAL, NULL, NULL, NULL, NULL);
```
nirbhayc, line 1829: [Optional] I'd prefix this with 'debug_' to make it explicit that this tunable is intended for debugging/testing.  Also, suggest you to make it of INTEGER type, so that the sleep duration in the test can later be adjusted.

schemachange/schemachange.c
riverszhang89: The old logic basically says, "for each blob in the blob list, if the schema change is to change the compression algorithm, or to remove ODH, unpack the blob." Note that the same check is done over and over again inside the for-loop. I've placed the check before the for-loop. Hence the new logic now says," if the schema change is to change the compression algorithm or to remove ODH, unpack each blob in the blob list."

db/db_tunables.h
riverszhang89: Thanks for the advice. I've made it a debug switch. I'm not quite in favor of making it an integer as sleep duration is unlikely to change. A hardcoded value would suffice.

riverszhang89:
@nirbhayc Thank you for the review! I will merge the pull request once RR and RM approve.

PR #2203, opened 2020-04-17
do not use client cnonce when transaction is chunked

dorinhogea:

If "SET TRANSACTION CHUNK n" is used, we need to disable using client cnonce as blkseq.  We are generating multiple transactions, and each needs its own blkseq.


dorinhogea:
note: there is a test already, but it disables sending cnonce using lrl

dorinhogea:
Thank you Mike!

PR #2206, opened 2020-04-22
Fix TXN accounting in comdb2_clientstats

embiv22:
bug fix

When you run a read committed txn, the snapiso and serialze counters are incremented by 1 as well. Only recom counter should be incremented

set transaction read committed
begin
insert into t1 values (1)
commit

select * from comdb2_clientstats --> serialize, and snapiso will be incremented by 1 as well as recom.


nirbhayc:
http://comdb2.s3-website-us-east-1.amazonaws.com/tests/d45cc8cb/detail.txt

nirbhayc:
Backported to 7.0.

nirbhayc:
Thanks Ed!

PR #2207, opened 2020-04-22
Fix sicountbug crash

markhannum:
Fix for sicountbug crashes that occur periodically on robo-rivers.  A snapshot session attempts to traverse the fileid queues locklessly when possible.  There was a bug in this logic- the fix updates pagelogs under the queue-lk until reaching the first record which has a commit_lsn which is greater than the transaction's birth lsn.

### DIFF
--- a/bdb/cursor.c
+++ b/bdb/cursor.c
@@ -7922,6 +7922,7 @@ static int update_pglogs_from_global_queues_int(
     bdb_cursor_impl_t *cur, struct pglogs_queue_cursor *qcur, int *bdberr)
 {
     struct pglogs_queue_key *current, *prev, *last;
+    int update_current_pglogs = 0;
 
 #ifdef NEWSI_STAT
     struct timeval before, after, diff;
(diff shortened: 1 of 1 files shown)


bdb/cursor.c
```diff
@@ -7956,6 +7956,15 @@ static int update_pglogs_from_global_queues_int(
 
         if (!found_greater)
             current = NULL;
+
+        while (current && current != last) {
+            current = current->lnk.next;
+            update_pglogs_from_queue(cur->shadow_tran, qcur->fileid, current);
```
adizaimi, line 7962: Is there a chance current can be NULL here, should we check?

markhannum: It shouldn't happen- it would have to go past "last" to be NULL, which was set under the same readlock that we're currently holding.  The thread which cleans these up holds the write-lock.

markhannum: If it's NULL I want to see the core that this will produce

adizaimi: Ok, we first go to prev elements, then we walk next and we stop when we hit current == last. 
This is fine.

PR #2209, opened 2020-04-23
DDL: Retrieve NONE rec/blob compression option properly (7.0)

nirbhayc:
* Also document the NONE option


riverszhang89:
Would you also consider adding NONE to [`bdb_compr2algo()`](https://github.com/bloomberg/comdb2/blob/master/bdb/odh.c#L159-L170), just to complete the logic?

PR #2210, opened 2020-04-23
Quote control characters in table_to_json

mponomar:

`db:table_to_json` was incorrectly not quoting control characters (`0x00-0x1F`). 
The [JSON standard](https://tools.ietf.org/html/rfc7159#section-7) requires it.    Add it as a tunable, on by default, off for existing databases for compatibility.


akshatsikarwar:
IMO, on by default with a way to opt out if breaks something (i.e. don't force this in legacy_defaults). Also perhaps 7.0 only  - for master we should consider moving to json provided by SQLite so we have just the one implementation in the database.

mponomar:
SQLite's implementation handles this correctly, unsurprisingly.  I'm good with making this <= 7.0 only.  I'd still rather add this to legacy_defaults so we don't slow down 7.0 deployment due to behavior changes.  Using SQLite's JSON won't be a clean fit, but I agree, it'd be nice to have a single JSON implementation in the database...

akshatsikarwar:
Not really - this is a bugfix to make output standards-compliant. We can assume that JSON parsers that receive this will handle it correctly.

But I don't want to delay 7.0 either - So how about making it off with legacy_defaults for 7.0 only. I can work on gutting cson and using SQLite in master.

I'm not in favor of exposing this as argument to `db:table_to_json` - then it is application code change to manage it. LRL option only please.

akshatsikarwar:
@mponomar 
> it'd be nice to have a single JSON implementation in the database...

This work was done: https://github.com/bloomberg/comdb2/pull/2235

PR #2212, opened 2020-04-23
Don't force C++11

mponomar:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,6 @@ cmake_minimum_required (VERSION 3.1)
 project (Comdb2)
 
 set(CMAKE_C_STANDARD 99)
-set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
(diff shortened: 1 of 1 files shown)


akshatsikarwar:
Fails to build with old compilers (smoke test fails to build):
```
/bb/comdb2a/roborivers/branches/master/tools/comdb2ar/appsock.h:29:5: error: 'unique_ptr' in namespace 'std' does not name a type     std::unique_ptr<Appsock_impl> impl;
```

akshatsikarwar:
We need c++11 only for tools -- so to fix sun build (where we don't compile tools) we can move this flag here: https://github.com/bloomberg/comdb2/blob/master/CMakeLists.txt#L223

akshatsikarwar:
Or better yet, we can move this into dedicated option and builddb, dpgk can call cmake with option to skip setting c++11/17
Alternatively, cmake option to add c++11/17 and smoketest can build with it.

akshatsikarwar:
Fixed: https://github.com/bloomberg/comdb2/commit/c9216cc7a123e6c2af7d16fc2be11c3fb6331822

PR #2220, opened 2020-04-29
Fix adding old-style-queues by msgtrap, and initcomdb2 with the queue directive

markhannum:
R7 fails initcomdb2 for lrl files which contain the queue directive.  This also contains fixes for adding queue via a message trap.

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -1866,7 +1866,6 @@ size_t gbl_lkr_parts = 23;
 size_t gbl_lk_hash = 32;
 size_t gbl_lkr_hash = 16;
 
-char **qdbs = NULL;
 char **sfuncs = NULL;
 char **afuncs = NULL;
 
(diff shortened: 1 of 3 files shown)


db/thrman.c
```diff
@@ -587,6 +588,7 @@ void resume_threads(struct dbenv *dbenv)
     if (!dbenv->purge_old_blkseq_is_running ||
         !dbenv->purge_old_files_is_running)
         create_old_blkseq_thread(dbenv);
+    create_watcher_thread(dbenv->bdb_env);
```
riverszhang89, line 591: I think the watcher thread needs to be registered in thrman (using `thrman_register()`) and checked in [`thrman_check_threads_stopped_ll()`](https://github.com/bloomberg/comdb2/blob/master/db/thrman.c#L462-L470). There'd otherwise be a slim chance that we reach here before the watcher thread's[ stop check](https://github.com/bloomberg/comdb2/blob/master/bdb/rep.c#L5275) (since `stop_threads()` doesn't wait for the thread to exit), and wind up having 2 watcher threads.

markhannum: YES .. thank you ..

markhannum: > Mike has a different (better) solution in https://github.com/bloomberg/comdb2/pull/2110 - I'm removing that part.

PR #2224, opened 2020-04-30
Log in statreq write count for every table 

adizaimi:
* Use statthd thread to update saved_write_count and aa_new_count 
* Show count diffs in statreq file
* Cleanup that logic from autoanalyze thread, update test
* Use atomic ops to update write_count for more accurate accounting
* Refactor function to print diff to statreq and update prev value, dont print DB number
* Get schema_lk in statthd when walking table list
* Refactor printing of OSQL type via array of strings instead of case stmt

The new write counts will appear as such:
```
04/30 13:20:57:   DIFF REQUEST STATS FOR TABLE 't1'
04/30 13:20:57:       inserted rows          32
04/30 13:20:57:       updated rows           32
04/30 13:20:57:       deleted rows           32
```

### DIFF
--- a/db/sqloffload.c
+++ b/db/sqloffload.c
@@ -189,63 +189,43 @@ void osql_cleanup(void)
     bdb_osql_destroy(&bdberr);
 }
 
-char *osql_breq2a(int op)
-{
+const char *osql_tp_str[] = {
+    "OSQL_RPLINV",
+    "OSQL_DONE",
+    "OSQL_USEDB",
+    "OSQL_DELREC",
+    "OSQL_INSREC",
+    "OSQL_CLRTBL",
+    "OSQL_QBLOB",
+    "OSQL_UPDREC",
+    "OSQL_XERR",
+    "OSQL_UPDCOLS",
+    "OSQL_DONE_STATS",
+    "OSQL_DBGLOG",
+    "OSQL_RECGENID",
+    "OSQL_UPDSTAT",
+    "OSQL_EXISTS",
+    "OSQL_SERIAL",
+    "OSQL_SELECTV",
+    "OSQL_DONE_SNAP",
+    "OSQL_SCHEMACHANGE",
+    "OSQL_BPFUNC",
+    "OSQL_DBQ_CONSUME",
+    "OSQL_DELETE",
+    "OSQL_INSERT",
+    "OSQL_UPDATE",
+    "OSQL_DELIDX",
+    "OSQL_INSIDX",
+    "OSQL_STARTGEN",
+    "OSQL_DONE_WITH_EFFECTS",
+};
 
-    switch (op) {
-    case OSQL_RPLINV:
-        return "INVALID";
-    case OSQL_DONE:
-        return "OSQL_DONE";
-    case OSQL_DONE_STATS:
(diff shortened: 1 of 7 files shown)


db/sqloffload.c
```diff
@@ -189,63 +189,43 @@ void osql_cleanup(void)
     bdb_osql_destroy(&bdberr);
 }
 
-char *osql_breq2a(int op)
+const char *osql_tp_str[] = {
```
akshatsikarwar, line 192: What do you think about doing it like: `CLIENT_TYPE_TO_STR`?

akshatsikarwar: But simpler still, so you just quote the field using preprocessor

akshatsikarwar:
Are the stats accurate if `include_updates` is disabled? Should we even continue to keep this option?

adizaimi:
> Are the stats accurate if `include_updates` is disabled? Should we even continue to keep this option?

Yeah, I think you are right -- updates won't be counted ... I may be able to get around it to fix the logic so they are accurate. But you have a valid point, is this option needed?

akshatsikarwar:
> But you have a valid point, is this option needed?

I think my original idea was to tune workloads which do a lot of updates but to non-indexed fields. I don't know if this was ever used.

adizaimi:
Not all updates are the same, for instance some may not change indices at all, some may change only one or two. So it is really hard to do smart accounting on updates -- I think we should remove it because there is no downside to perform an analyze, even when one is not really useful.

adizaimi:
As an aside, we might want to add logic for autoanalyze threads to run at a lower priority, or otherwise backoff if db/system is under heavy load.

db/sqloffload.c
adizaimi: I'll fix this in a subsequent PR. Good suggestion.

adizaimi:
> > Are the stats accurate if `include_updates` is disabled? Should we even continue to keep this option?
> 
> Yeah, I think you are right -- updates won't be counted ... I may be able to get around it to fix the logic so they are accurate. But you have a valid point, is this option needed?

I am trying to understand what you mean and it seems I did previously, and now can't figure it out anymore. 
aa_include_updates is infact disabled by default, and whether it is on or off, the only thing that it affects is tbl->aa_saved_counter:
     `4602: ATOMIC_ADD32(tbl->aa_saved_counter, (curr - prev));` 
because curr and prev will be affected by that tunable.
The `inserted_rows`, `udated_rows`, and `deleted_rows` do not depend on that tunable. Please take another look because I am a bit confused at this point.

adizaimi:
A slight problem might be aa_saved_counter() where we read the values of aa_saved_counter[] without lock, thus we could be reading in the middle of them being updated (from statthd()), and so they have a minuscule chance of being slightly incorrect. The workaround for that is to do what we do in auto_analyze_main(), which is 
`435:             get_saved_counter_epoch(tbl->tablename, &llmeta_aa_saved_counter, NULL);`
ie. reads from llmeta, however that is a bit too heavy handed, and unnecessary for stats--do you agree?

adizaimi:
Thanks Akshat.

PR #2229, opened 2020-05-06
timeout remote writes if we lose some connectivit

dorinhogea:

This is the simplified timeout for remote writes from R6.   The plan going forward is to replace the fdb custom wire protocol with cdb2api.


dorinhogea:
Approved in R6.

PR #2233, opened 2020-05-08
PORTED R7: Fix crashes because atomic inc needs a 32 bit int

adizaimi:
Fix crashes in solaris because atomic inc needs a 32 bit int
PR to masterbranch #2174 


adizaimi:
Thank you for the review.

PR #2235, opened 2020-05-12
cson: Replace cson with SQLite's JSON implementation

akshatsikarwar:
/silent


akshatsikarwar:
I will be adding a standalone binary to test just the library.

PR #2237, opened 2020-05-15
Simple fixes for dohast on upgrades and downgrades

markhannum:
Fix dohast child and master logic for upgrades and downgrades: dohsql_end_distribute and dohsql_wait_for_master block indefinitely waiting for state changes initiated by the other thread.  Make sure that these release their reference to the bdb-readlock if this machine should want to upgrade, downgrade, or do a verify-match.

### DIFF
--- a/db/dohsql.c
+++ b/db/dohsql.c
@@ -318,6 +318,7 @@ static void _track_que_free(dohsql_connector_t *conn)
 static void _que_limiter(dohsql_connector_t *conn, sqlite3_stmt *stmt,
                          int row_size)
 {
+    int rc;
     if (gbl_dohsql_max_queued_kb_highwm) {
         conn->queue_size += row_size;
     }
@@ -335,6 +336,16 @@ cleanup:
                 conn->status != DOH_MASTER_DONE) {
                 Pthread_mutex_unlock(&conn->mtx);
                 poll(NULL, 0, gbl_dohsql_full_queue_poll_msec);
+                if (bdb_lock_desired(thedb->bdb_env)) {
+                    rc = recover_deadlock_simple(thedb->bdb_env);
+                    if (rc) {
+                        Pthread_mutex_lock(&conn->mtx);
+                        logmsg(LOGMSG_ERROR,
+                               "%s: failed recover_deadlock rc=%d\n", __func__,
+                               rc);
+                        return;
+                    }
+                }
                 Pthread_mutex_lock(&conn->mtx);
                 goto cleanup;
             }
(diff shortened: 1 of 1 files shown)


markhannum:
.. the smoke testing passes when I run on my laptop .. in the failures above I see this trace:

cdb2_open rc -1 cdb2_get_dbhosts: can't do dbinfo query on db1 hosts.
cdb2_open rc -1 cdb2_get_dbhosts: can't do dbinfo query on db1 hosts.
cdb2_open rc -1 cdb2_get_dbhosts: can't do dbinfo query on db1 hosts.
cdb2_open rc -1 cdb2_get_dbhosts: can't do dbinfo query on db1 hosts.
Timed out!

> I will try to run directly on RR

riverszhang89:
> .. the smoke testing passes when I run on my laptop .. in the failures above I see this trace:
> 
> cdb2_open rc -1 cdb2_get_dbhosts: can't do dbinfo query on db1 hosts.
> cdb2_open rc -1 cdb2_get_dbhosts: can't do dbinfo query on db1 hosts.
> cdb2_open rc -1 cdb2_get_dbhosts: can't do dbinfo query on db1 hosts.
> cdb2_open rc -1 cdb2_get_dbhosts: can't do dbinfo query on db1 hosts.
> Timed out!
> 
> > I will try to run directly on RR

Master fails it the same way intermittently. I don't think it's related to your changes.

markhannum:
Gotcha - thanks Rivers!

db/dohsql.c
```diff
@@ -1285,6 +1296,14 @@ int dohsql_end_distribute(struct sqlclntstate *clnt, struct reqlogger *logger)
         while (conns->conns[i].status != DOH_CLIENT_DONE) {
             Pthread_mutex_unlock(&conns->conns[i].mtx);
             poll(NULL, 0, 10);
+            if (bdb_lock_desired(thedb->bdb_env)) {
```
dorinhogea, line 1299: NOTE: this should be safe imho since this is called in the parallel worker thread when calling write_response(RESPONSE_ROW).   We already call recover deadlock in this place for other reasons.

db/dohsql.c
```diff
@@ -1361,6 +1381,16 @@ void dohsql_wait_for_master(sqlite3_stmt *stmt, struct sqlclntstate *clnt)
         while (conn->status == DOH_RUNNING && queue_count(conn->que) > 0) {
             Pthread_mutex_unlock(&conn->mtx);
             poll(NULL, 0, 10);
+            if (bdb_lock_desired(thedb->bdb_env)) {
```
dorinhogea, line 1384: NOTE: this should be safe imho since we are calling this from parallel worker thread when its work is done; it waits for master to consume its row from que before freeing the resources allocated and terminating.

PR #2238, opened 2020-05-15
Dohsql recover deadlock 7

markhannum:
This is the 7.0 port of https://github.com/bloomberg/comdb2/pulls/2237


markhannum:
Reviewed and approved in https://github.com/bloomberg/comdb2/pull/2237

PR #2243, opened 2020-05-16
Support for bind datetimeus in cdb2sql

adizaimi:
Add support to bind datetimeus in cdb2sql

### DIFF
--- a/tests/cdb2bind.test/out.expected
+++ b/tests/cdb2bind.test/out.expected
@@ -1,30 +1,30 @@
 Usage: @bind <type> <paramname> <value>, with type one of the following:
-CDB2_{INTEGER,REAL,CSTRING,BLOB,DATETIME,INTERVALYM,INTERVALDS}
+CDB2_{INTEGER,REAL,CSTRING,BLOB,DATETIME[,US]}
 [@bind] rc -1
 Usage: @bind <type> <paramname> <value>, with type one of the following:
-CDB2_{INTEGER,REAL,CSTRING,BLOB,DATETIME,INTERVALYM,INTERVALDS}
+CDB2_{INTEGER,REAL,CSTRING,BLOB,DATETIME[,US]}
 [@bind] rc -1
 Usage: @bind <type> <paramname> <value>, with type one of the following:
-CDB2_{INTEGER,REAL,CSTRING,BLOB,DATETIME,INTERVALYM,INTERVALDS}
+CDB2_{INTEGER,REAL,CSTRING,BLOB,DATETIME[,US]}
 [@bind  CDB2_STUFF] rc -1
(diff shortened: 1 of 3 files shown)


cdb2api/cdb2api.c
```diff
@@ -4746,30 +4746,26 @@ static int cdb2_run_statement_typed_int(cdb2_hndl_tp *hndl, const char *sql,
...
-    case CDB2_INTERVALYM:
-        return "CDB2_INTERVALYM";
-    case CDB2_INTERVALDS:
-        return "CDB2_INTERVALDS";
-    case CDB2_DATETIMEUS:
-        return "CDB2_DATETIMEUS";
-    case CDB2_INTERVALDSUS:
-        return "CDB2_INTERVALDSUS";
-    default:
+const char *cdb2_tp_str[] = {
```
nirbhayc, line 4749: static

mponomar, line 4749: I'd really rather not expose more cdb2api internals if it can be helped.  The types are exposed in cdb2api.h, maybe have a similar function/array there?

adizaimi: Not sure why i can't see Nirbhay's suggestion anymore--to make the array static:
static const char *cdb2_tp_str[] = { ... }
would that be ok with you? 
I'm not sure what moving this function or the array to the .h accomplishes.

mponomar: Sure.  I didn't want to expose more cdb2api innards - that was my only concern.

cdb2api/cdb2api.c
```diff
@@ -4746,30 +4746,26 @@ static int cdb2_run_statement_typed_int(cdb2_hndl_tp *hndl, const char *sql,
...
+    "CDB2_BLOB",
+    "???",
+    "CDB2_DATETIME",
+    "CDB2_INTERVALYM",
+    "CDB2_INTERVALDS",
+    "CDB2_DATETIMEUS",
+    "CDB2_INTERVALDSUS"
+};
+
+inline const char *cdb2_type_str(int type)
```
mponomar, line 4763: Maybe use #2242 to get this instead of exposing another API call?

cdb2api/cdb2api.c
adizaimi: K, i understand we don't want to add random functions to the api. This can be implemented separately by client program. Thanks.

cdb2api/cdb2api.c
adizaimi: Its ok. im leaving cdb2api out of this one, as discussed.

adizaimi:
Thank you for the review Rivers and Mike.

PR #2248, opened 2020-05-17
cdb2api: bind parameter list

adizaimi:
Bind parameter list by calling new cdb2api function:
```
int cdb2_bind_list(cdb2_hndl_tp *hndl, const char *varname, int count,
                   int type[], const void *varaddr[], int length[]);
int cdb2_bind_array(cdb2_hndl_tp *hndl, const char *varname, unsigned int count,
                    int type, const void *varaddr[], int typelen, int length[]);
```

Example usage:
```
    int type[2] = {CDB2_INTEGER, CDB2_INTEGER};
    int a = 11, b = 33;
    const void *vals[2] = {&a, &b};
    int lengths[2] = { sizeof(a), sizeof(b)};

    cdb2_bind_list(hndl, "list", 2, type, vals, lengths);
    cdb2_run_statement("select * from t2 where a in (@list[])");
```

Will replace the `@list[]` part with `@list_0,@list_1` and bind each of
of the `count==2` elements passed in.
Please see testcase in pr for more usage examples.

### DIFF
--- a/cdb2api/cdb2api.c
+++ b/cdb2api/cdb2api.c
@@ -1035,8 +1035,10 @@ struct cdb2_hndl {
     char stack[MAX_STACK];
     int send_stack;
     void *user_arg;
-    char *bind_list_name;
+    char *bind_list_string; /* string to replace with a list */
     char *bind_list_repl;
+    char *bind_array_name;
+    char *bind_array_repl;
     int gbl_event_version; /* Cached global event version */
     int api_call_timeout;
     int connect_timeout;
@@ -4806,20 +4808,20 @@ int cdb2_run_statement_typed(cdb2_hndl_tp *hndl, const char *sql, int ntypes,
     sql = cdb2_skipws(sql);
     char *new_sql = NULL;
     const char *actual_sql = sql;
-    if (hndl->bind_list_name && hndl->bind_list_repl) {
-        char *p = strstr(sql, hndl->bind_list_name);
+    if (hndl->bind_list_string && hndl->bind_list_repl) {
+        char *p = strstr(sql, hndl->bind_list_string);
         if (!p) {
-            fprintf(stderr, "List bind parameter '%s' not found in sql '%s'\n", hndl->bind_list_name, sql);
+            fprintf(stderr, "List bind parameter '%s' not found in sql '%s'\n", hndl->bind_list_string, sql);
             return -1;
         }
         int p1len = p - sql; // part 1
 
         int list_len = strlen(hndl->bind_list_repl);
-        new_sql = malloc(strlen(sql) + list_len);
+        new_sql = malloc(strlen(sql) + list_len + 1);
         strncpy(new_sql, sql, p1len);
         strncpy(&new_sql[p1len], hndl->bind_list_repl, list_len);
         strcat(&new_sql[p1len + list_len],
-               &sql[p1len + strlen(hndl->bind_list_name)]);
+               &sql[p1len + strlen(hndl->bind_list_string)]);
         actual_sql = new_sql;
         /* replace ',' with \0 for the bind parameter name */
         for(int i = 0; i < list_len; i++) {
@@ -4833,8 +4835,8 @@ int cdb2_run_statement_typed(cdb2_hndl_tp *hndl, const char *sql, int ntypes,
     if (rc)
         debugprint("rc = %d\n", rc);
 
-    if (hndl->bind_list_name && hndl->bind_list_repl) {
-        free(hndl->bind_list_name); hndl->bind_list_name = NULL;
+    if (hndl->bind_list_string && hndl->bind_list_repl) {
+        free(hndl->bind_list_string); hndl->bind_list_string = NULL;
         free(hndl->bind_list_repl); hndl->bind_list_repl = NULL;
         free(new_sql); new_sql = NULL;
     }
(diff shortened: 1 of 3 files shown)


mponomar:
This is great, I haven't considered doing this on the client side.   There's  a couple of things:

- It breaks for other bind names that are a prefix, eg: `select a, @abc from t where a in (@a)`
- It breaks for tricky edge case like `select a, '@a' from t where a in (@a)`.  We can easily exclude comments and strings I guess, but it involves some additional client-side parsing (and with the above, you need to teach it token boundaries)
- There's two type of developers who want to use this, and I don't think we can please both with one call.
  - I have a list of arbitrary types and values in my python list and I don't want to bind values individually.  This is great for them.
  - I have a C++ vector and I want to bind the values as efficiently as possible.   This is cumbersome for this type of user, since they still need to go over each value - they'd want something like `cdb2_bind_array(cdb2_hndl_tp *, void *values, int count, int type, int *lengths)` (with lengths being optional for everything except blob).

tests/tools/cdb2bind.c
```diff
@@ -262,13 +262,102 @@ void test_03()
...
+        "SELECT COUNT(*) FROM t1 WHERE c1 = x'' AND c2 = ''";
+    const char *select_null_cmd =
+        "SELECT COUNT(*) FROM t1 WHERE c1 IS NULL AND c2 IS NULL";
+        */
+
+    test_open(&hndl, db);
+
+    {
+        int type[2] = {CDB2_INTEGER, CDB2_INTEGER};
+        int a = 11, b = 33;
```
godlygeek, line 290: Would it be interesting to do:
```c
int a = 11;
short b = 33;
```
instead?  The case of binding two different integers of different sizes seems like it could be worth covering.

Also, does the case where `count=0` warrant a test?

cdb2api/cdb2api.c
```diff
@@ -5081,9 +5115,53 @@ int cdb2_clearbindings(cdb2_hndl_tp *hndl)
...
+/* bind @list with array of type/variable */
+int cdb2_bind_list(cdb2_hndl_tp *hndl, const char *varname, int count,
+                   int type[], const void *varaddr[], int length[])
+{
+    free(hndl->bind_list_name);
+    free(hndl->bind_list_repl);
+    int len = strlen(varname);
+    hndl->bind_list_name = malloc(len + 1);
+    hndl->bind_list_name[0] = '@';
+    strcpy(&hndl->bind_list_name[1], varname); // to search
```
godlygeek, line 5134: Unless I'm misreading, there's an off by one here and this needs to `malloc(len + 2)`.  I believe this overflows the buffer - given `varname="x"`, I think this allocates 2 bytes but writes `{'@', 'x', '\0'}`

cdb2api/cdb2api.c
```diff
@@ -5081,9 +5115,53 @@ int cdb2_clearbindings(cdb2_hndl_tp *hndl)
...
+    hndl->bind_list_repl = malloc((len + 8) * count); // to replace
+    hndl->bind_list_repl[0] = '\0';
+
+    char loc_varname[len + 6];
```
godlygeek, line 5138: Why `+ 6`?  32-bit ints are up to 10 digits long, so for this to be correct it seems like you need to check if `0 <= count <= 9999` here.

cdb2api/cdb2api.c
```diff
@@ -5081,9 +5115,53 @@ int cdb2_clearbindings(cdb2_hndl_tp *hndl)
...
+                   int type[], const void *varaddr[], int length[])
+{
+    free(hndl->bind_list_name);
+    free(hndl->bind_list_repl);
+    int len = strlen(varname);
+    hndl->bind_list_name = malloc(len + 1);
+    hndl->bind_list_name[0] = '@';
+    strcpy(&hndl->bind_list_name[1], varname); // to search
+    hndl->bind_list_repl = malloc((len + 8) * count); // to replace
+    hndl->bind_list_repl[0] = '\0';
```
godlygeek, line 5136: Do you usually guard against bad user inputs?  If a negative count is provided, this will crash.

cdb2api/cdb2api.c
```diff
@@ -4809,10 +4812,41 @@ int cdb2_run_statement_typed(cdb2_hndl_tp *hndl, const char *sql, int ntypes,
     }
 
     sql = cdb2_skipws(sql);
-    rc = cdb2_run_statement_typed_int(hndl, sql, ntypes, types, __LINE__);
+    char *new_sql = NULL;
+    const char *actual_sql = sql;
+    if (hndl->bind_list_name && hndl->bind_list_repl) {
```
godlygeek, line 4817: Is this just for proof of concept purposes, or is it intended to be part of the contract that the SQL string may only contain one list, and binding another is an error?  If there may be only one list, you could consider using `@0`, `@1`, etc as the placeholders (comdb2 allows strings starting with a digit as the placeholder) - but you should also make a second call to `cdb2_bind_list` without an intervening call to `cdb2_clearbindings` or `cdb2_execute` return an error.

Using placeholders that start with a digit would also ameliorate some of the concerns about colliding with a placeholder that the user put into the string, since users are relatively unlikely to bind parameters beginning with digits.

adizaimi:
@godlygeek thanks for the comments; we are currently investigating what is appropriate wrt to offering this feature--so as it is, this PR is a proof of concept and will probably be modified.

godlygeek:
> * I have a C++ vector and I want to bind the values as efficiently as possible.   This is cumbersome for this type of user, since they still need to go over each value - they'd want something like `cdb2_bind_array(cdb2_hndl_tp *, void *values, int count, int type, int *lengths)` (with lengths being optional for everything except blob).

It occurs to me - `lengths` being optional is probably not what type of developer would want, either.  `cdb2_bind_param` allows binding a `short` or `int` or `long long` as `CDB2_INTEGER`, so users who want to bind every item of a `vector` to a list would probably actually want the ability to specify one length that specifies the size of every `vector` element, rather than an array of lengths that collectively specify the size of each element.  Most users aren't holding `bsl::vector<long long>` I suspect, but rather `bsl::vector<int>`.

mponomar:
Sure.  `cdb2_bind_array(cdb2_hndl_tp *, void *values, int count, int type, int typelen, int *lengths)` then, or something similar

adizaimi:
This was a fun experiment, but a better way of doing this is from the server side by sending array in the message and binding on the server directly. That work will be done in https://github.com/bloomberg/comdb2/pull/2599, so this PR can be closed.

Issue #2247, opened 2020-05-17
Failed testcase isn't flagged correctly


akshatsikarwar:
Run testcase: `make eventlog`:
```
...
check all .*sql entries for fingerprint
Failed output is different from expected
exit -1 
check the sql parameter now
...
```
Looks like this only exits the sub-shell, while test keeps going. Also, I think the test itself is buggy as expected output is incorrect.

akshatsikarwar:
I have fixed this particular instance (https://github.com/bloomberg/comdb2/pull/2235/commits/a80b52359d70bee04d8c7cf2f4b56beddb3567c0) but perhaps we need a larger fix so to catch where ever this construct is used.
Proposal: `failexit` can touch a trap file and test-harness can check for its existence so we don't silently pass any failing tests.

akshatsikarwar:
@adizaimi

adizaimi:
Yes, possibly more tests have same issue -- your suggestion is a good one. Few tests do that already, we can generalize it that way.

akshatsikarwar:
This is fixed. Thank you Adi.

PR #2249, opened 2020-05-18
Tests: check for failexit to correctly identify failed tests

adizaimi:
From failexit() function touch ${DBNAME}.failexit file and
from runtestcase check for such file to always correctly
identify failed tests (which were not being caught because of
subshell).


akshatsikarwar:
Same for `assertres`?

adizaimi:
As expected eventlog is failing. Assertres calls failexit so does not need to be changed.

adizaimi:
Fixes issue  #2247.

PR #2253, opened 2020-05-19
Use thread local variable in get_area()

adizaimi:
End result is same underlying behavior but without the penalty
of pthread_getspecific(), which allows some savings of about 1%
in a simple select test case (as reported by valgrind).

The usage here is a bit peculiar in having redundant pointers
to the same zone because there is no way to cleanup __thread
variables in c. We use the pthread_key variable to perform
cleanup on thread exit, and for checking whether zone is null
(which is done hundreds of thousand of times) we use the
__thread variable.

### DIFF
--- a/mem/mem.c
+++ b/mem/mem.c
@@ -270,6 +270,7 @@ static int comdb2ma_destroy_int(comdb2ma cm);
 
 #ifdef PER_THREAD_MALLOC
 __thread const char *thread_type_key;
+static __thread comdb2ma *t_zone;
 static comdb2ma get_area(int indx);
 static void destroy_zone(void *);
 #else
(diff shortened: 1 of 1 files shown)


mem/mem.c
```diff
@@ -256,6 +256,10 @@ static struct {
           .mmap_thresh = 0,
           .nice = 0};
 
+#ifdef PER_THREAD_MALLOC
+static __thread comdb2ma *t_zone = NULL;
+#endif
+
```
riverszhang89, line 262: There's already a `#ifdef PER_THREAD_MALLOC` block just a few lines down here. Can this declaration be moved there?

mem/mem.c
```diff
@@ -1406,6 +1411,7 @@ static void destroy_zone(void *arg)
             }
 
         mspace_free(root.m, arg);
+        t_zone = NULL;
```
riverszhang89, line 1414: This is needless as the thread is already exiting.

mem/mem.c
adizaimi: yes, thanks.

mem/mem.c
adizaimi: yep, so was the initial = NULL. Thanks.

adizaimi:
Thank you Rivers!

PR #2256, opened 2020-05-20
Custom delimiter string

mohitkhullar:
Custom delimiter string.


tests/cdb2sql_options.test/runit
```diff
@@ -0,0 +1,31 @@
...
+        echo "failed $testname"
+        echo "see diffs here: $HOSTNAME"
+        echo "> diff -u ${PWD}/{$testname.expected,$testname.output}"
+        echo
+        exit 1
+    fi
+done
+
+echo
+exit 0
```
nirbhayc, line 31: Use [compare_results.sh](https://github.com/bloomberg/comdb2/blob/master/tests/tools/compare_results.sh) instead?

mohitkhullar: will need to change the script to start accepting additional options, the script already uses -d for database name.

nirbhayc: You can put it in $CDB2_OPTIONS.

tools/cdb2sql/cdb2sql.cpp
```diff
@@ -1699,12 +1735,13 @@ int main(int argc, char *argv[])
         {"cdb2cfg", required_argument, NULL, 'c'},
         {"file", required_argument, NULL, 'f'},
         {"gensql", required_argument, NULL, 'g'},
+        {"delim", required_argument, NULL, 'd'},
```
nirbhayc, line 1738: Also add it to [usage_test]()https://github.com/bloomberg/comdb2/blob/master/tools/cdb2sql/cdb2sql.cpp#L127

tools/cdb2sql/cdb2sql.cpp
```diff
@@ -1379,7 +1414,8 @@ static int run_statement(const char *sql, int ntypes, int *types,
         } else if (printmode & DISP_TABS) {
             fprintf(out, "\n");
         } else if (printmode & DISP_GENSQL) {
-            fprintf(out, ");\n");
+            fprintf(out, ");");
```
nirbhayc, line 1417: We should not need `;` anymore.
`fprintf(out, ")");`

tools/cdb2sql/cdb2sql.cpp
```diff
@@ -472,6 +477,20 @@ static bool skip_history(const char *line)
     return true;
 }
 
+int has_delimiter(char *line, int len, char *delimiter, int dlen)
+{
+    if (dlen > len)
+        return -1;
```
nirbhayc, line 483: [optional] use bool instead - return false here.

tools/cdb2sql/cdb2sql.cpp
mohitkhullar: This is to keep the current behavior, we use newline as delimiter but we output ';\n' in gensql.

tools/cdb2sql/cdb2sql.cpp
mohitkhullar: done

tools/cdb2sql/cdb2sql.cpp
mohitkhullar: done

PR #2258, opened 2020-05-20
make sure we generate proper target table in dohast

dorinhogea:

Currently dohast code doesn't generate proper table names for remote tables.  This patch appends the class and database name for remote tables.  The union all now supports remote parallel execution as well.


dorinhogea:
Thank you Nirbhay!

PR #2259, opened 2020-05-20
make sure we generate proper target table in dohast

dorinhogea:

7.0 port for https://github.com/bloomberg/comdb2/pull/2258


dorinhogea:
Thanks again!

PR #2266, opened 2020-05-21
simple fix for deadlock recovery for remote write backend

dorinhogea:

Recover deadlock uses sql thread thd to recover clnt state.  It is not set for remote writes backend, simple fix for it.


dorinhogea:
Thank you Nirbhay!

PR #2271, opened 2020-05-22
cson: Small string optimization

akshatsikarwar:
/silent

### DIFF
--- a/cson/cson.c
+++ b/cson/cson.c
@@ -521,10 +521,6 @@ int cson_array_append(cson_array *ar, cson_value *v)
     cson_value_free(v);
     return 0;
 }
-int cson_array_reserve(cson_array *ar, unsigned int size)
-{
-    return 0;
-}
 cson_value *cson_value_new_integer(cson_int_t v)
 {
     cson_value *val = calloc(1, sizeof(cson_value));
--- a/cson/cson.c
+++ b/cson/cson.c
@@ -127,11 +127,12 @@ struct cson_value {
     int sub_type;
     int value_bytes;
     char *value_text;
+    char value_buf[128];
+    int modified;
     struct {
         cson_value **slots;
         int capacity;
         int used;
-        int modified;
     } replace;
     JsonParse parse;
     union {
@@ -257,9 +258,11 @@ static void cson__reset(cson_value *val)
     val->replace.capacity = 0;
     val->replace.slots = NULL;
 
-    free(val->value_text);
-    val->value_text = NULL;
-    val->value_bytes = 0;
+    if (val->value_text != val->value_buf) {
+        free(val->value_text);
+        val->value_text = NULL;
+        val->value_bytes = 0;
+    }
 
     if (!val->sub_type) {
         return;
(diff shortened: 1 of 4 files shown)


cson/cson.c
```diff
@@ -518,18 +528,18 @@ int cson_array_append(cson_array *ar, cson_value *v)
...
     val->u.num = v;
-    char buf[128];
-    val->value_bytes = sprintf(buf, "%" PRId64, v);
-    val->value_text = strndup(buf, val->value_bytes);
+    val->value_bytes = snprintf(val->value_buf, sizeof(val->value_buf), "%" PRId64, v);
+    if (val->value_bytes < sizeof(val->value_buf)) {
+        val->value_text = val->value_buf;
+    } else {
+        val->value_text = malloc(val->value_bytes + 1);
+        sprintf(val->value_text, "%" PRId64, v);
```
adizaimi, line 541: not sure why you cant do here: 
  val->value_text = strndup(buf, val->value_buf);
but if you have to malloc, instead of sprintf you can do:
  strcpy(val->value_text, val->value_buf);

akshatsikarwar: I can't do that because `value_buf` is too small to hold stringy-fied integer.

akshatsikarwar: This and other such codepaths are taken when stringy-fied versions of value is larger than 48 bytes. So most small strings, double values, integer values etc never hit this codepath.

cson/cson.c
```diff
@@ -127,11 +127,13 @@ struct cson_value {
     int sub_type;
     int value_bytes;
     char *value_text;
+    char value_buf[48];
```
adizaimi, line 130: This is good to have a optional buffer to use like this.
i was wondering if we could stick value_text at the end of the obj so for instance when you create an int we would do:
```
cson_value *cson_value_new_integer(cson_int_t v)
 {
     char buf[128];
     val->value_bytes = sprintf(buf, "%" PRId64, v);

     cson_value *val = calloc(1, sizeof(cson_value)  + val->value_bytes );
     val->sql_type = SQLITE_INTEGER;
     val->u.num = v;
     val->value_text = val + sizeof(cson_value); 
     strncpy(val->value_text, buf, val->value_bytes);
     return val;
 }
```

More work needed for correctness, you would need to not free() value_text, etc.

cson/cson.c
adizaimi: I understand, maybe we could keep the original char buf[128]; version, int count = sprintf() into it, and if (count < 48) { strcpy(value_buf, buf) } else {value_buf = strdup(buf)};

cson/cson.c
akshatsikarwar: We can revisit it in subsequent PR. Having fixed size buf lets us increase without new alloc. Imagine an array with 2 small ints. Initial allocation of array needs 3 bytes only `[`,`]`,`0`. On appending, we need to realloc.

cson/cson.c
akshatsikarwar: I think it is better to optimize for common case where payload fits in buffer and no additional copies or allocations are done. Still, I've increased size of buf to 128 to catch even larger things (fingerprints, small queries, etc). Let is review this if it is actually a performance issue.

akshatsikarwar:
I can't get a clean test run. Master is failing as well. e.g.: http://comdb2.s3-website-us-east-1.amazonaws.com/tests/3cd7a871/detail.txt
Gonna merge this now to move forward.

PR #2273, opened 2020-05-26
Replication lag

riverszhang89:
Report how far an incoherent replicant is behind the master, in the `comdb2_repl_stats` system table.

### DIFF
--- a/tests/repl_stats.test/runit
+++ b/tests/repl_stats.test/runit
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+# Test replication stats
+#
+# Force replication delay on a replicant and then do a pushnext on the master
+# We should be seeing the replicant is behind the master in master's
+# comdb2_repl_stats table.
+
+bash -n "$0" | exit 1
+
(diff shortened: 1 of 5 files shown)


tests/repl_stats.test/runit
```diff
@@ -0,0 +1,26 @@
...
+
+bash -n "$0" | exit 1
+
+set -e
+
+dbnm=$1
+
+master=`cdb2sql --tabs ${CDB2_OPTIONS} $dbnm default 'select host from comdb2_cluster where is_master="Y"'`
+rep=`cdb2sql --tabs ${CDB2_OPTIONS} $dbnm default 'select host from comdb2_cluster where is_master="N" limit 1'`
+
```
nirbhayc, line 17: One quick note: the test would timeout if run against a standalone setup. I'd skip the test in that case.
`
if [[ -z "${rep}" ]]; then
    echo "skipping, it's a cluster test"
    exit 0;
fi
`

riverszhang89:
@nirbhayc Thank you for the review! I've made changes per your suggestions.

PR #2274, opened 2020-05-27
Avoid doing unnecessary work for eventlog when not enabled

adizaimi:
Avoid doing unnecessary work for eventlog:
* Dont create cson_value_xxx when arrary is null (when eventlog is not detailed)
* use strcpy instead of sprintf("%s") for better performance


adizaimi:
Thank you Akshat.

PR #2278, opened 2020-05-28
net: Set addr when accepting new host

akshatsikarwar:
/silent


akshatsikarwar:
`udp_send` uses this cached value to send data.

PR #2279, opened 2020-05-29
Rollover queuedbs between one and two files based on tunables.

mistachkin:

When there are two files for a queuedb, the first one is always being consumed from and the second one is alrways being added to. Furthermore, when the first file is empty, it will be automatically cleaned up and the second file will be swapped over to it, thus becoming the new first file.  New tunables are added to control this process.  In addition, several related concurrency enhancements have been made.


bdb/bdb_schemachange.h
```diff
@@ -24,32 +24,46 @@ extern volatile uint32_t gbl_analyze_gen;
...
-    lua_sfunc,
-    lua_afunc,
-    rename_table,
-    change_stripe,
-    user_view,
+    invalid = -1,            // -1
+    alter,                   //  0
+    fastinit,                //  1
+    add = fastinit,          //  2
+    drop,                    //  3
```
mistachkin, line 31: Typo in comment here.  Will fix in subsequent commit.

PR #2282, opened 2020-05-30
Dump sql commands that are blocking replication thread

nirbhayc:
* tunable: dump_sql_on_repwait_sec
* @send sql dump repblockers


berkdb/lock/lock.c
```diff
@@ -2605,6 +2607,14 @@ __lock_get_internal_int(lt, locker, in_locker, flags, obj, lock_mode, timeout,
 			++sh_obj->generation;
 			unlock_detector(region);
 		}
+
+                extern u_int32_t gbl_rep_lockid;
```
riverszhang89, line 2611: spaces -> tabs for `berkdb/`.

akshatsikarwar:
Can you pull in new master and check with libevent? Thanks!

berkdb/rep/rep_record.c
```diff
@@ -5192,7 +5193,7 @@ bad_resize:	;
 		goto err;
 #endif
 	}
-
+	gbl_rep_lockid = lockid;
```
markhannum, line 5196: YES .. parallel rep can have multiple outstanding lockers, but only one of them will be acquiring locks at any given time ..

berkdb/lock/lock.c
```diff
@@ -2031,6 +2035,51 @@ rep_return_deadlock(DB_ENV *dbenv, u_int32_t sz)
...
+	region = lt->reginfo.primary;
+
+	OBJECT_INDX(lt, region, &gbl_rep_lockobj.obj, ndx, partition);
+
+	lock_obj_partition(region, partition);
+
+	ret = __lock_getobj(lt, &gbl_rep_lockobj.obj, ndx, partition, 0, &obj);
+
+	if (ret != 0 || !obj) {
+		lock_obj_partition(region, partition);
```
markhannum, line 2069: > This should be unlock_obj_partition ..

berkdb/lock/lock.c
```diff
@@ -2605,6 +2654,17 @@ __lock_get_internal_int(lt, locker, in_locker, flags, obj, lock_mode, timeout,
 			++sh_obj->generation;
 			unlock_detector(region);
 		}
+
+		extern u_int32_t gbl_rep_lockid;
+		if (locker == gbl_rep_lockid) {
```
adizaimi, line 2659: Can we optimize and do this only if action is  HEAD,TAIL, or SECOND?

adizaimi: Also, can we just copy newl  and do the __lock_to_dbt() conversion when we need to dump the holder's list?

nirbhayc: > Can we optimize and do this only if action is HEAD,TAIL, or SECOND?

That covers all the cases.

adizaimi: Sorry, I failed to see that we were already inside the outer switch. Ignore the first comment.

nirbhayc:
@markhannum @adizaimi Thanks for the review & suggestions!

PR #2284, opened 2020-06-03
Hibernate Dialect for JDBC table creation Foreign key Fix

aibolit:
bug fix
Current behavior is crashing when creating tables with foreign keys
Have hibernate create 2 tables with a foreign key. It will crash. 
An easy test project that demonstrates this is https://github.com/aibolit/StandaloneGORM. Cloning and running the project will cause the crash. It will print out the bad table creation SQL. You will notice the FK sql is run right afterwards. 
The old behavior used to inline the foreign key in the table creation. This was done 3 years ago. Since then, the ddl changed to not only not require this but create a crash.


aibolit:
This PR also upgrades the version of the Gradle build tool and the hibernate core library.

riverszhang89:
@aibolit Thank you for the pull request!

PR #2285, opened 2020-06-03
Serialize a protobuf object to sbuf directly (7.0)

riverszhang89:
Utilize `protobuf_c_message_pack_to_buffer` to directly write a protobuf message to sbuf without serializing it to memory first.

(DRQS 159321603)


riverszhang89:
@mponomar Thank you for the reivew!

PR #2286, opened 2020-06-03
Fix error message when upsert is not supported

nirbhayc:
Comdb2 reported 'unknown error' for INSERT OR IGNORE when upsert wasn't enabled. Also update docs and tests for INSERT OR (IGNORE|REPLACE) syntax.


adizaimi:
We don't have a case in our testsuite where we would print 'UPSERT not enabled', correct? We always have that enabled.
This looks good.

nirbhayc:
@adizaimi added some tests, thanks for looking!

nirbhayc:
Cherry-picked to 7.0

nirbhayc:
Thank you @adizaimi !

PR #2287, opened 2020-06-03
sqlite: disable unsupported features in parser

nirbhayc:
The following sqlite extensions are not supported in comdb2:
* INSERT OR ABORT
* UPDATE OR REPLACE
* UPDATE OR IGNORE


adizaimi:
Looks good, but can we add tests that they are blocked, since we are at it?

nirbhayc:
@adizaimi added some tests.

PR #1532, opened 2019-04-26
CHECK constraint

nirbhayc:
.

PR #2291, opened 2020-06-05
Supporting multiple constraints sections in csc2 (7.0)

riverszhang89:
The bug is 7.0 only: the server crashes if a csc2 schema contains multiple `constraints` sections.

(DRQS 159823090)

### DIFF
--- a/csc2/macc_so.c
+++ b/csc2/macc_so.c
@@ -3215,15 +3215,13 @@ int dyns_get_table_field_count(char *tabletag)
 
 int dyns_get_constraint_count(void)
 {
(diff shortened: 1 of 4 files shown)


riverszhang89:
The issue was fixed in master by https://github.com/bloomberg/comdb2/pull/1532. However it appears that we don't have a test for it. I'm adding one now.

riverszhang89:
@nirbhayc Thank you for the review! I agree with you that it makes more sense to remove `end_constraint_list()` and increment the counter in `start_constraint_list()` instead. However I wanted to keep the change minimal so I kept the function there.

PR #2293, opened 2020-06-05
Fix double counting of num_accepts (7.0)

riverszhang89:
(DRQS 158891075)

### DIFF
--- a/net/net.c
+++ b/net/net.c
@@ -5676,8 +5676,6 @@ static void *accept_thread(void *arg)
             last_stat_dump_time = pollend;
         }
 
-        netinfo_ptr->num_accepts++;
-
         /* drop connection on poll error */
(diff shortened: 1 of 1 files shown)


net/net.c
```diff
@@ -5676,8 +5676,6 @@ static void *accept_thread(void *arg)
             last_stat_dump_time = pollend;
         }
 
-        netinfo_ptr->num_accepts++;
```
riverszhang89, line 5679: We've already incremented it just [a few lines above](https://github.com/bloomberg/comdb2/pull/2293/files#diff-158d650f8c314f5cf37ed472eed53b5bL5670).

nirbhayc:
I have resubmitted it to rr as it was apparently broken during the first run.

riverszhang89:

PR #2295, opened 2020-06-11
Test and fix locking for comdb2_queues

markhannum:
Protect the comdb2_queues system table with a lock on that table.  All schema-changes on queues must acquire this lock in write-mode.  I've written the queue_locking test which demonstrates the issue- current master fails this test.

### DIFF
--- a/db/dbqueuedb.c
+++ b/db/dbqueuedb.c
@@ -163,12 +163,11 @@ int dbqueuedb_get_stats(struct dbtable *db, struct consumer_stat *stats, uint32_
         bdb_get_tran_lockerid(trans, &savedlid);
         bdb_set_tran_lockerid(trans, lockid);
 
-        // TODO: re-enable when systable fixes are checked in
-        if (0 && gbl_debug_systable_locks) {
+        if (gbl_debug_systable_locks) {
             bdb_assert_tablename_locked(bdb_state, "_comdb2_systables", lockid, ASSERT_TABLENAME_LOCKED_READ);
         }
 
-        if (0 && gbl_assert_systable_locks) {
+        if (gbl_assert_systable_locks) {
             bdb_assert_tablename_locked(bdb_state, "comdb2_queues", lockid, ASSERT_TABLENAME_LOCKED_READ);
         }
     }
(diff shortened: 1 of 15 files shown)


markhannum:
The timeout for the maxtable test is too aggressive (it had finished successfully).

markhannum:
This is a moving target (still), but this pr has gotten large.  Akshat asked that I outline the current design for anyone who wants to review this.  This originally was intended to address locking in the comdb2_queues table (a customer had crashed the database while accessing this table).  The testcase has unearthed two design flaws in our systable implementation:

* System tables access global data structures which are protected by schemalk, however, because it's executing on behalf of an SQL transaction which can be holding any number of berkley locks, acquiring the schema lock here violates lock-ordering, and can deadlock the database.
* As almost all system tables create a berkley transaction to probe llmeta- this is wrong because the sql-curtran-transaction in the same thread can already be holding any number of berkley locks, which can again deadlock the database.

To address the first issue, we are introducing three additional berkley level locks ("comdb2_tables", "comdb2_queues", and "comdb2_views") which an SQL thread will acquire in READ mode if it accesses systables which display information about tables, queues, or views.  SQL threads will acquire one or more of these locks in the perpare phase.  SQLite calls the authorizer against every table in a query during this phase- we record any additional locks this table would require in a list hung off of the authorizer structure.  At the end of prepare, we copy this list into the SQLite pCur/vdbe object.  It stays inside of the pCur/vdbe for the same reason the list of tables stays- to accommodate prepared statements.  sqliteLockStmtTables now gathers tablelocks for each of these, and all relavent schema changes to acquire the appropriate lock in write mode.

To address the second issue, I've created a utility function, curtran_gettran, which returns a transaction object which utilizes the thread's curtran-lockid.  This transaction is used while the systable retrieves information from llmeta, and then destroyed with curtran_puttran.  Alot of this code in this PR deals with creating transactional versions of these llmeta functions- it looks like alot, it's extremely straightforward.

markhannum:
I'm chunking this into smaller PRs

markhannum:
https://github.com/bloomberg/comdb2/pull/2351 is the final checkin for this- this pr is no longer necessary

PR #2296, opened 2020-06-12
ANALYZE fails for OP user

nirbhayc:
When authentication is enabled, ANALYZE table commands are limited to OP users only. But, since the user credentials are not inherited by the analyze threads, they fail the authorization check.

/plugin-branch comdb2-auth-analyze

### DIFF
--- a/db/sql.h
+++ b/db/sql.h
@@ -576,6 +576,17 @@ struct sql_hist_cost {
     int64_t rows;
 };
 
+struct user {
+    char name[MAX_USERNAME_LEN];
+    char password[MAX_PASSWORD_LEN];
+
+    uint8_t have_name;
+    uint8_t have_password;
+    /* 1 if the user is retrieved from a client certificate */
+    uint8_t is_x509_user;
+};
+
+
 #define in_client_trans(clnt) ((clnt)->in_client_trans)
 
 /* Client specific sql state */
@@ -692,14 +703,7 @@ struct sqlclntstate {
     struct query_effects log_effects;
     int64_t nsteps;
 
-    int have_user;
-    char user[MAX_USERNAME_LEN];
-    int is_x509_user; /* True if the user is retrieved
-                         from a client certificate. */
-
-    int have_password;
-    char password[MAX_PASSWORD_LEN];
-
+    struct user current_user;
     int authgen;
 
     int no_transaction;
--- a/db/sqlinterfaces.c
+++ b/db/sqlinterfaces.c
@@ -2910,7 +2910,7 @@ int release_locks_on_emit_row(struct sqlthdstate *thd,
     return 0;
 }
 
-/* if userpassword does not match this function
+/* If user password does not match this function
  * will write error response and return a non 0 rc
  */
 static inline int check_user_password(struct sqlclntstate *clnt)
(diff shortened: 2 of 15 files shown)


riverszhang89:
```
/bb/bigstorn/systems/gitbuilds/comdb2/opencomdb2/cdb2test/pnj-comdb2btda1/R20200611.7/bb-plugins/sqlitex/comdb2build.c:1493:71: error: 'struct sqlclntstate' has no member named 'user'
          snprintf(tableName, MAXTABLELEN, "%s@%s", p->zName, thd->clnt->user);
                                                                       ^~
plugins/sqlitex/CMakeFiles/sqlitex.dir/build.make:187: recipe for target 'plugins/sqlitex/CMakeFiles/sqlitex.dir/comdb2build.c.o' failed
make[3]: *** [plugins/sqlitex/CMakeFiles/sqlitex.dir/comdb2build.c.o] Error 1
```

db/sql.h
```diff
@@ -576,6 +576,16 @@ struct sql_hist_cost {
     int64_t rows;
 };
 
+struct user {
+    int have_name;
```
adizaimi, line 580: we can turn this into uint8_t since you are at it...have_password too. Also reorder so char name[] is first, int field is second, and the the uint8_t and the char passwd.

db/sqlinterfaces.c
```diff
@@ -2944,12 +2944,42 @@ static inline int check_user_password(struct sqlclntstate *clnt)
...
+    if (clnt && !clnt->current_user.is_x509_user &&
+        clnt->current_user.have_name) {
+        return clnt->current_user.name;
     }
     return NULL;
 }
 
+void copy_user(struct user *dst, struct user *src)
+{
+    dst->have_name = src->have_name;
```
adizaimi, line 2956: why would simple assignment not work for copying this struct?
*dst = *src;

db/sqlinterfaces.c
```diff
@@ -2944,12 +2944,42 @@ static inline int check_user_password(struct sqlclntstate *clnt)
...
+        strcpy(dst->password, src->password);
+    }
+}
+
+static void reset_user(struct sqlclntstate *clnt)
+{
+    if (!clnt)
+        return;
+
+    /* reset the user */
```
adizaimi, line 2973: how about: 
memset(&clnt->current_user, 0, sizeof(struct user));

adizaimi: or maybe :
clnt->current_user.have_name = 0;
    clnt->current_user.is_x509_user = 0;
clnt->current_user.name[0] = '\0';
clnt->current_user.have_password = 0;
clnt->current_user.password[0] = '\0';
...
Not sure which would be faster. Certainly have function be inline for completenes.

riverszhang89: We could just bzero `current_user` all together.

plugins/newsql/newsql.c
```diff
@@ -1642,18 +1642,19 @@ static int process_set_commands(struct dbenv *dbenv, struct sqlclntstate *clnt,
...
-                    if (strlen(sqlstr) >= sizeof(clnt->user)) {
+                    if (strlen(sqlstr) >= sizeof(clnt->current_user.name)) {
                         snprintf(err, sizeof(err),
                                  "set user: '%s' exceeds %zu characters",
-                                 sqlstr, sizeof(clnt->user) - 1);
+                                 sqlstr, sizeof(clnt->current_user.name) - 1);
                         rc = ii + 1;
                     } else {
-                        clnt->have_user = 1;
+                        clnt->current_user.have_name = 2;
```
riverszhang89, line 1651: Why is it set to 2 here?

nirbhayc: Typo, I have tmux leader mapped to Ctrl+a, which also increments numbers in Vim.

sqlite/src/comdb2build.c
```diff
@@ -2244,7 +2246,7 @@ int resolveTableName(struct SrcList_item *p, const char *zDB, char *tableName,
    if ((zDB && (!strcasecmp(zDB, "main") || !strcasecmp(zDB, "temp"))))
    {
        snprintf(tableName, len, "%s", p->zName);
-   } else if (clnt && (clnt->user[0] != '\0') &&
+   } else if (clnt && (clnt->current_user.name[0] != '\0') &&
```
riverszhang89, line 2249: Actually, should we use `clnt->current_user.have_name` here to determine whether a username exists? If we assume a username can't be empty, we do not need `have_name`. If a username can be an empty string, `have_name` should be used here.

nirbhayc: 'empty' user name (and password) seems to be working fine except for when user schema is enabled, where I have found a problem. Let me handle that in a separate PR.

riverszhang89: Yes we use `have_user` and `user[0] != '\0'` interchangeably...

nirbhayc:
@riverszhang89 @adizaimi Thanks for the review!

nirbhayc:
Cherry-picked to 7.0

PR #2299, opened 2020-06-16
cdb2api: Fix incorrect rollback behavior

riverszhang89:
If a transaction bails out because a node goes down, the API retries its subsequent ROLLBACK statement which result in a "wrong sqlengine state" error. The ROLLBACK statement should be a no-op and yield a good return code.

(DRQS 159888006)


riverszhang89:
The change was reviewed internally.

Issue #2272, opened 2020-05-24
Two undefined behavior problems 


Jerry-aha:
Two undefined behavior problems are found when fuzzing with comdb2.

## Problem 1
The bug report is:
```sh
/root/workspace/comdb2/sources/db/comdb2.c:5422:33: runtime error: left shift of 3653224 by 16 places cannot be represented in type '__pid_t' (aka 'int')
    #0 0x2b94254 in main /root/workspace/comdb2/sources/db/comdb2.c:5422:33
    #1 0x7fedcd41a001 in __libc_start_main (/usr/lib/libc.so.6+0x27001)
    #2 0x711b5d in _start (/root/workspace/comdb2/bin_asan/comdb2.dfuzz-counter,dfuzz-asan-instrument.out+0x711b5d)
```
The code is 
```C
srand(time(NULL) ^ getpid() << 16);
```
The return value type of getpid() is int, and the bit operation of int might overflows, so the behavior is undefined. The compiler may optimize the operation arbitrarily, resulting in random number seed conflict.
According to ISO C99 6.5.7
> The result of E1 << E2 is E1 left-shifted E2 bit positions; vacated bits are filled with zeros. If E1 has an unsigned type, the value of the result is E1 * 2^E2, reduced modulo one more than the maximum value representable in the result type. If E1 has a signed type and nonnegative value, and E1 * 2^E2 is representable in the result type, then that is the resulting value; otherwise, the behavior is undefined.

## Problem 2
The bug report is:
```sh
/root/workspace/comdb2/sources/net/net.c:3023:37: runtime error: signed integer overflow: 3653224 * 65537 cannot be represented in type 'int'
    #0 0x2c513f9 in create_netinfo /root/workspace/comdb2/sources/net/net.c:3023:37
    #1 0x2be1630 in open_bdb_env /root/workspace/comdb2/sources/db/glue.c:3654:33
    #2 0x2b9e43c in init /root/workspace/comdb2/sources/db/comdb2.c:3796:9
    #3 0x2b93abe in main /root/workspace/comdb2/sources/db/comdb2.c:5466:9
    #4 0x7fedcd41a001 in __libc_start_main (/usr/lib/libc.so.6+0x27001)
    #5 0x711b5d in _start (/root/workspace/comdb2/bin_asan/comdb2.dfuzz-counter,dfuzz-asan-instrument.out+0x711b5d)
SUMMARY: UndefinedBehaviorSanitizer: signed-integer-overflow
```

[...]

PR #2300, opened 2020-06-18
Correct two instances of undefined behavior

adizaimi:
Casting to unsigned int to avoid undefined behavior
from left shifting a large int.
Addresses #2272

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -5418,7 +5418,7 @@ int main(int argc, char **argv)
 
     init_q_vars();
 
-    srand(time(NULL) ^ getpid() << 16);
+    srand(time(NULL) ^ (((unsigned int)getpid()) << 16));
     srandom(comdb2_time_epochus());
 
     if (debug_switch_verbose_deadlocks())
--- a/net/net.c
+++ b/net/net.c
@@ -3020,7 +3020,7 @@ netinfo_type *create_netinfo(char myhostname[], int myportnum, int myfd,
     netinfo_ptr->user_data_buf_size = 256 * 1024;
 
     netinfo_ptr->throttle_percent = 50;
-    netinfo_ptr->seqnum = (getpid() * 65537);
+    netinfo_ptr->seqnum = ((unsigned int)getpid()) * 65537;
     netinfo_ptr->myport = myportnum;
     netinfo_ptr->myhostname = intern(myhostname);
     netinfo_ptr->myhostname_len = strlen(netinfo_ptr->myhostname) + 1;


riverszhang89:
The warnings are integer overflows. I'm not sure casting an operand to unsigned is the right way.

adizaimi:
> The warnings are integer overflows. I'm not sure casting an operand to unsigned is the right way.

>>> If E1 has a signed type and nonnegative value, and E1 × 2E2 is representable in the result type, then that is the resulting value; otherwise, the behavior is undeﬁned.

Telling the compiler to treat the bit representation as unsigned, will allow it to perform correct shifting to the left (which is defined behavior).

riverszhang89:
Gotcha. Then maybe `getpid() << 16U` is simpler.

adizaimi:
> Gotcha. Then maybe `getpid() << 16U` is simpler.

I don't know that for this purpose 16U does what you want here. E1 type seems to not be affected by E2 (based from the wording from the standard).

riverszhang89:
Shouldn't the signed be converted to an unsigned if you make one of them unsigned?

riverszhang89:
The change looks good. I'm just trying to understand if there's a simpler way to do this.

adizaimi:
I don't think E1 is converted to unsigned by doing E1 << 16U.

riverszhang89:
`netinfo_ptr->seqnum = ((unsigned int)getpid()) * 65537;`

Actually, how could this avoid the integer overflow? seqnum is an `int`.

riverszhang89:
And you could certainly do `getpid() * 65537U` here if you just want unsigned operands.

adizaimi:
> And you could certainly do `getpid() * 65537U` here if you just want unsigned operands.



> `netinfo_ptr->seqnum = ((unsigned int)getpid()) * 65537;`
> 
> Actually, how could this avoid the integer overflow? seqnum is an `int`.

According to the standard(http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf), unsigned int overflow is well defined:
6.2.5 (pg 34): 9 The   range   of   nonnegative   values   of   a   signed   integer   type   is   a   subrange   of   thecorresponding  unsigned  integer  type,  and  the  representation  of  the  same  value  in  eachtype  is  the  same.31)Acomputation  involving  unsigned  operands  can  neveroverflow,because  a  result  that  cannot  be  represented  by  the  resulting  unsigned  integer  type  isreduced  modulo  the  number  that  is  one  greater  than  the  largest  value  that  can  berepresented by the resulting type.

However assigning to a signed int value that overflows is undefined:
J.3.5: (page 506):
—The  result  of,  or  the  signal  raised  by,converting  an  integer  to  a  signed  integer  typewhen the value cannot be represented in an object of that type (6.3.1.3).

So in this case we should be ok since seqnum is unsigned int:
301     unsigned int seqnum;

adizaimi:
> And you could certainly do `getpid() * 65537U` here if you just want unsigned operands.

I was not sure if the second operand determines the final un/signedness of the multiplication. Standard apparently says you are correct (6.3.1.8  Usualarithmetic conversions) but the description of the rule is a bit confusing to me.

It is cleaner just casting i think.

PR #2302, opened 2020-06-19
Don't hold usedb between appsock calls

mponomar:
This just removes usedb from the shared appsock struct, and replaces it with name.  There's further plugin changes that address locking usedb while it's being used.


markhannum:
This looks great - makes sense with the corresponding bb-plugins pr.  Good to merge!

PR #2304, opened 2020-06-22
String param to comdb2_die so abort reason in stack trace

adizaimi:
String param to comdb2_die so abort reason in stack trace

### DIFF
--- a/db/watchdog.c
+++ b/db/watchdog.c
@@ -435,7 +435,7 @@ void watchdog_enable(void)
 
 void lock_info_lockers(FILE *out, bdb_state_type *bdb_state);
 
-void comdb2_die(int aborat)
+void comdb2_die(int doabort, const char *str)
 {
     pid_t pid;
     char pstack_cmd[128];
@@ -463,7 +463,7 @@ void comdb2_die(int aborat)
         }
     }
 
-    if (aborat)
+    if (doabort)
         abort();
     else
         _exit(1);
@@ -489,7 +489,7 @@ static void *watchdog_watcher_thread(void *arg)
             */
             if (failed_once > 0) {
                 logmsg(LOGMSG_FATAL, "watchdog thread stuck, exiting\n");
-                comdb2_die(1);
+                comdb2_die(1, "watchdog thread stuck");
             }
             failed_once++;
             continue;
@@ -504,7 +504,7 @@ static void *watchdog_watcher_thread(void *arg)
            gets stuck every time there is berkdb lockdown */
         if (tmstmp - gbl_watcher_thread_ran > 60) {
             logmsg(LOGMSG_FATAL, "rep watcher thread stuck, exiting\n");
-            comdb2_die(1);
+            comdb2_die(1, "rep watcher thread stuck");
         }
     }
     return NULL;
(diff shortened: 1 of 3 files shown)


riverszhang89:
Isn't the entire call stack visible in gdb? Why do we need this?

adizaimi:
To instantaneously identify the reason from the stack trace from the crash without needing to open the source code of the caller.

riverszhang89:
The caller already tells you what has aborted the database. For instance, if watcher_thread calls comdb2_die, we know it's the watchdog.

riverszhang89:
Let's say the stack trace shows `TRANS_ABORT FAILED`. Since there're a handful occurrences of `TRANS_ABORT FAILED`, how am I able to identify where exactly abort was requested without a debugger?

adizaimi:
Well, I was looking at such a stack:
#0  0x00007fca518991f7 in raise () from /lib64/libc.so.6 
#1  0x00007fca5189a8e8 in abort () from /lib64/libc.so.6 
#2  0x000000000067f903 in comdb2_die (aborat=1) at watchdog.c:446 
#3  0x000000000067f965 in watchdog_watcher_thread (arg=0x4bb5c50) 
    at watchdog.c:462 
#4  0x00007fca52ce0e25 in start_thread () from /lib64/libpthread.so.0 
#5  0x00007fca5195c34d in clone () from /lib64/libc.so.6 

And it was not clear what's the reason from line 462. I'd need to get that version's source to know...and thought 'it would be surely nice to just have the reason in the stack.'

adizaimi:
> Let's say the stack trace shows `TRANS_ABORT FAILED`. Since there're a handful occurrences of `TRANS_ABORT FAILED`, how am I able to identify where exactly abort was requested without a debugger?

In this case it is less useful, surely.

db/watchdog.c
```diff
@@ -435,7 +435,7 @@ void watchdog_enable(void)
 
 void lock_info_lockers(FILE *out, bdb_state_type *bdb_state);
 
-void comdb2_die(int aborat)
+void comdb2_die(int doabort, const char *str)
```
riverszhang89, line 438: Since you're not using `str` here at all, maybe void it - some compilers do warn of it.

riverszhang89:
> Well, I was looking at such a stack:
> #0 0x00007fca518991f7 in raise () from /lib64/libc.so.6
> #1 0x00007fca5189a8e8 in abort () from /lib64/libc.so.6
> #2 0x000000000067f903 in comdb2_die (aborat=1) at watchdog.c:446
> #3 0x000000000067f965 in watchdog_watcher_thread (arg=0x4bb5c50)
> at watchdog.c:462
> #4 0x00007fca52ce0e25 in start_thread () from /lib64/libpthread.so.0
> #5 0x00007fca5195c34d in clone () from /lib64/libc.so.6
> 
> And it was not clear when i opened that file what's the reason from line 462 (because it's from an older version). I'd need to get that version's source to know...and thought 'it would be surely nice to just have the reason in the stack.'

Before the database deliberately aborts, it usually prints a line of message. That message is the reason. And I do not think you can debug a coredump without having read and understood the source code.

adizaimi:
This is not very useful, admittedly. Thanks Rivers.

PR #2305, opened 2020-06-22
Fix sc_rebuilds.test when run in singlenode

adizaimi:
- missing assigning master for singlenode case
- added check for correct response when trimming sc_history


markhannum:
Very straightforward- good to merge!

PR #2306, opened 2020-06-22
tz: Delay freeing tz on exit

akshatsikarwar:
`backend_close` correctly waits for threads which have bdb rdlock, so any point after it would be ok to free tz. I've moved it all the way to the end of `clean_exit` just in case something else may call into tz routines (event_log?).

/silent


riverszhang89:
Cherry-pick it back to 7.0?

PR #2307, opened 2020-06-23
Add millisecond/microsecond functions

mponomar:
We have functions for other datetime/datetimeus components
but did not have them for millisecond/microsecond.


sqlite/src/dttz.c
```diff
@@ -153,134 +157,130 @@ int convMem2ClientDatetimeStr(Mem *pMem, void *out, int outlen, int *outdtsz) {
...
+          assert(out != NULL);
+
+          if(convMem2ClientDatetime(argv[0], out) != SQLITE_OK) {
+              sqlite3_result_null(context);
+              return;
+          }
+
+          switch (pMem->du.dt.dttz_prec) {
+              case DTTZ_PREC_MSEC:
+                  sqlite3_result_double(context, (double) cdt.msec);
```
riverszhang89, line 232: I understand why you would want to return a double here. However it seems a bit inconsistent if the fraction field is returned as a double but the seconds field is returned as an integer. After all both "1.2 second" and "0.1 milliseconds" are valid. Maybe just round it down here. What do you think?

mponomar: I thought about that, and yes, it's inconsistent with the other fields.   The problem is that we truncate data - microseconds(....7us) -> 0, which also felt wrong.  It makes sense that the lowest resolution field supports decimals.

mponomar: Another option is to return the milliseconds/microseconds as if they were two separate fields.  Ie: 2020-01-01T01:01:01.002004 returns 2 for millisecond() and 4 for microseconds.  That's consistent with other fields.

riverszhang89: I rethought about it - you're right it makes more sense to return decimals for fractions.

adizaimi: It *is* a bit odd that for '2021-02-03T04:05:06.000007', millisecond() returns 0.007000. Not sure if any alternative is better.
For, microsecond(), there will never be digits beyond the decimal point, right? So microsecond(b)=7.000000 could in fact be an integer? (I'm not sure i understand Mike's first comment about 7us -> 0.)

PR #2309, opened 2020-06-23
Trigger epoch

akshatsikarwar:
Allow trigger clients to obtain event epoch. Also includes `dbq_get` cleanup. `epoch` argument is redundant.


akshatsikarwar:
Thank you!

PR #2310, opened 2020-06-24
Cache internal statements used to perform various checks

nirbhayc:
Comdb2 internally prepares and executes statements to perform verification for check constraints and function indices (during ALTER). Caching these statements has shown significant improvement in the overall execution times.

CHECK constraint
```sql
-- prepare
create table t1(i int, check (i > 0)) $$
-- test
insert into t1 select * from generate_series(1,10000);
```
Execution times - _before caching_: 1.007s -- _after caching_: 0.622s (**+61.90%**)

ALTER a table with partial index
```sql
-- prepare
create table t2(i int) $$
create index "idx1" on t2(i) where (i > 0);
insert into t2 select * from generate_series(1,10000);
-- test
alter table t2 alter column i type float $$
```
Execution times - _before caching_: 2.510s -- _after caching_: 1.623s (**+54.65%**)

This PR also reorgs/refactors the statement caching code, which has now been moved to a separate compilation unit under db/sql_stmt_cache.(h|c).


adizaimi:
This is nifty!
May I propose two things: 1) for more clarity, can you put info on what internal stmt is being cached in example in PR summary above, and 2) It may be easier to see the difference here if there were two checkins, one to split stmt caching into new file (in which case it would be mostly moving existing code), and the second with changes for caching internal sql (if it is too much trouble, obviously don't bother).

nirbhayc:
@adizaimi I'll break it into 2 PRs. Thank you for looking!

nirbhayc:
I'll open 2 separate PRs for this.

PR #2311, opened 2020-06-24
Fix locking in queues and queue systables

markhannum:
This is a bite-sized chunk of https://github.com/bloomberg/comdb2/pull/2295.  This fixes locking in queues and in queue-admin.  dbq_get has been changed to accept a lockid arguement.  The code will use this lockid in the transaction it creates to query the queue.  Additionally, this revises sp.c to always getqueuebyname while holding the schema-lk to accommodate the possibility that the queue was dropped during a recover_deadlock.

The systable_locking test has been de-fanged and only queries systables relating to queues: comdb2_queues and comdb2_triggers.

### DIFF
--- a/lua/sp.c
+++ b/lua/sp.c
@@ -163,6 +163,7 @@ struct dbconsumer_t {
     int register_timeoutms;
     int emit_timeoutms;
     time_t registration_time;
+    char name[MAXTABLELEN];
 
     /* signaling from libdb on qdb insert */
     pthread_mutex_t *lock;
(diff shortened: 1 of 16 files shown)


markhannum:
Actually the test I just added can fail because sql isn't acquiring the queue_lock (or any other lock) before querying the queue systable.  That's coming in a later chunk: I will disable this part of the test for now

lua/sp.c
```diff
@@ -197,6 +198,7 @@ static int setup_dbconsumer(dbconsumer_t *q, struct consumer *consumer,
 {
     init_fake_ireq(thedb, &q->iq);
     int spname_len = info->spname_len;
+    strncpy(q->name, qdb->tablename, MAXTABLELEN);
```
adizaimi, line 201: off by one?

markhannum:
Behaves like master under RoboMark

markhannum:
> Reviewed this with Mike earlier today

PR #2313, opened 2020-06-25
Change default for escaping control characters in table_to_json

mponomar:
This was meant to be enabled by default in 7.0

### DIFF
--- a/db/config.c
+++ b/db/config.c
@@ -368,7 +368,6 @@ static char *legacy_options[] = {
     "init_with_queue_ondisk_header off",
     "init_with_queue_compr off",
     "usenames",
-    "json_escape_control_characters off",
 };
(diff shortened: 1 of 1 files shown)


db/config.c
```diff
@@ -368,7 +368,7 @@ static char *legacy_options[] = {
     "init_with_queue_ondisk_header off",
     "init_with_queue_compr off",
     "usenames",
-    "json_escape_control_characters off",
+    "json_escape_control_characters on",
```
nirbhayc, line 371: This tunable is enabled by default anyway (https://github.com/bloomberg/comdb2/blob/7.0/lua/sp.c#L93), may be just remove it from this list altogether.

riverszhang89: Yeah good catch.

PR #2315, opened 2020-06-25
Acquire locks against certain comdb2 systables

markhannum:
This is another chunk of https://github.com/bloomberg/comdb2/pull/2295 - this detects the tables used for this sql statement via the authorizer code as an SQLITE_READ.  If the table is a systable which requires a lock, the name of that lock is added to a list, and stored with the prepared statement.  In addition to the normal table locks, schema change acquires writelocks against systables categories that can be affected.

### DIFF
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -61,6 +61,7 @@
 #include <bdb_api.h>
 #include <bdb_cursor.h>
 #include <bdb_fetch.h>
+#include <bdb_int.h>
 #include <time.h>
 
 #include "comdb2.h"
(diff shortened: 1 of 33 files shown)


db/sqlglue.c
```diff
@@ -7329,6 +7330,7 @@ static int rootpcompare(const void *p1, const void *p2)
     return strcmp(tp1->zName, tp2->zName);
 }
 
+int gbl_debug_systable_locks = 0;
```
nirbhayc, line 7333: Can we have this enabled for DEBUG builds? This will allow us to automatically test these changes without having to explicitly flip this switch.

db/sqlinterfaces.c
```diff
@@ -708,6 +708,49 @@ void sqlinit(void)
         abort();
 }
 
+static char *vtable_lockname(const char *table)
```
riverszhang89, line 711: I'm thinking that this part could be integrated into the system table creation routine. The system table lock type can be an additional argument of the creation routine. And then in this function, we could just look up the `table` in system table hashtable (just like how we do it in `select * from comdb2_systabs`), and return its lock type. This way it'd be easier to maintain.

markhannum: wow great idea .. actually ..

markhannum:
I actually have to disable acquiring the systable locks from sql threads until more of this is checked in.  This is because several systables acquire the schema-lock.  Getting a systable lock at prepare time and then later acquiring the schema-lock reverses the lock-order and causes deadlocks if run concurrently with schema change.  I will address this in the next chunk of https://github.com/bloomberg/comdb2/pull/2295

PR #2317, opened 2020-06-26
Use bplog to send to master operation to delete from sc_history

adizaimi:
- Use bplog commands to delete from sc_history
   - Remove remnants of stale osql_check_replicant_numops tunable
   - sc_rebuilds.test can send send trim command to any node, fix tail -n +3

### DIFF
--- a/db/bpfunc.c
+++ b/db/bpfunc.c
@@ -33,6 +33,7 @@ static int exec_rowlocks_enable(void *tran, bpfunc_t *func,
                                 struct errstat *err);
 static int exec_genid48_enable(void *tran, bpfunc_t *func, struct errstat *err);
 static int exec_set_skipscan(void *tran, bpfunc_t *func, struct errstat *err);
+static int exec_delete_from_sc_history(void *tran, bpfunc_t *func, struct errstat *err);
 /********************      UTILITIES     ***********************/
 
 static int empty(void *tran, bpfunc_t *func, struct errstat *err)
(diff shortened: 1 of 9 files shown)


sqlite/src/comdb2build.c
```diff
@@ -6602,3 +6632,43 @@ void comdb2_drop_view(Parse *pParse, SrcList *pName)
 out:
     free_schema_change_type(sc);
 }
+
+/* delete entry by seed from comdb2_sc_history 
+ * called from stored procedure function db_comdb_delete_sc_history()
+ */
+int comdb2DeleteFromScHistory(char *tablename, uint64_t seed)
+{
+    BpfuncArg *arg = (BpfuncArg*) malloc(sizeof(BpfuncArg));
```
riverszhang89, line 6641: A structure initialized on stack would suffice.

db/bpfunc.c
```diff
@@ -520,3 +526,12 @@ static int exec_rowlocks_enable(void *tran, bpfunc_t *func, struct errstat *err)
     return rc;
 }
 
+static int exec_delete_from_sc_history(void *tran, bpfunc_t *func,
+                                       struct errstat *err)
+{
+    BpfuncDeleteFromScHistory *tblseed = func->arg->tblseed;
+    int rc = bdb_del_schema_change_history(NULL, tblseed->tablename, tblseed->seed);
```
markhannum, line 533: Why aren't you using the transaction for bdb_del_schema_change_history here?

adizaimi: Good catch. Fixing.

sqlite/src/comdb2build.c
adizaimi: yes, i think on the stack works. thanks.

markhannum:
Looks good- okay to merge!

PR #2319, opened 2020-06-30
db:emit type-conv failure

akshatsikarwar:
Delay throwing runtime-error when type conversion fails during `emit`. Instead, we wait for a failed `write_response`, release `emit_mutex`, then throw error.


akshatsikarwar:
Working on test cases.

akshatsikarwar:
O HI DCO CHEX

PR #2321, opened 2020-07-02
Reproduce sc replication error

markhannum:
Simple fixes to paper over schema-change finalize commits which successfully write a commit record, but fail to replicate to all of the replicant nodes.  We should ignore this error code, and write an scdone record in this case.  I've modified the simple_timepart test to reproduce the issue while running the reptimeout.testopt version of this test.

### DIFF
--- a/bbinc/debug_switches.h
+++ b/bbinc/debug_switches.h
@@ -63,6 +63,7 @@ int debug_switch_verbose_cursor_deadlocks(void);         /* 0 */
 int debug_switch_check_multiple_lockers(void);           /* 1 */
 int debug_switch_dump_pool_on_full(void);                /* 1 */
 int debug_switch_scconvert_finish_delay(void);           /* 0 */
+int debug_switch_fake_sc_replication_timeout(void);      /* 0 */
 
 /* value switches */
 int debug_switch_net_delay(void); /* 0 */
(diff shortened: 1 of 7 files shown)


db/db_tunables.h
```diff
@@ -1977,6 +1977,10 @@ REGISTER_TUNABLE("client_abort_on_slow",
                  "Enable watchdog to abort if a \"slow\" client is detected."
                  "  (Default: off)", TUNABLE_BOOLEAN, &gbl_client_abort_on_slow,
                  EXPERIMENTAL | INTERNAL, NULL, NULL, NULL, NULL);
+REGISTER_TUNABLE("fake_sc_replication_timeout",
```
riverszhang89, line 1980: [`debug_switches.c`](https://github.com/bloomberg/comdb2/blob/master/util/debug_switches.c) seems a better place for the tunable since it's for debugging purpose only.

adizaimi:
NOTE that now we have two ways of setting debug ability for testing--the other one uses the "debug." notation:
`(name='debug.autoanalyze', description='debug autoanalyze operations', type='BOOLEAN', value='OFF', read_only='N')`
which is set via `cdb2sql --tabs ${CDB2_OPTIONS} --host $master $dbnm "put tunable 'debug.autoanalyze' 1"`

PR #2323, opened 2020-07-02
Fix a memory leak in datetime and interval arithmetic (7.0)

riverszhang89:
(DRQS 160938282)


riverszhang89:
@adizaimi Thank you for the review! That's a fantastic idea! I'll be working on it in a separate pull request.

PR #2327, opened 2020-07-08
Fix error with expert mode and datacopy index

adizaimi:
If there is a datacopy index we have issue parsing
the sql string from sqlite_master and throw error
`Support for 'COLLATE DATACOPY' syntax has been removed; use OPTION DATACOPY.`
This patch changes the sqlite_master sql for index creation
from `create index "I1" on "t1" ("i", "b" collate DATACOPY, "c");`
to `create index "I1" on "t1" ("i", "b" , "c") OPTION DATACOPY;`
thus allowing the expert mode to be able to parse the index creation sql.


markhannum:
Looks good - okay to merge!

akshatsikarwar:
This fails to build
```
[100%] Linking CXX executable comdb2
ld: 0711-317 ERROR: Undefined symbol: .strcasestr
```

adizaimi:
Thanks Akshat, fixed in 097ea5c1: fix missing strcasestr in ibm build, use strstr.

PR #2328, opened 2020-07-08
Discard locks at cursor close for readonly lockids which are wrapped in a transaction

markhannum:
I've ported Akshat's demonstration as a new test ("qget_curtran") to our testsuite.  This test demonstrates that simple queue operations can cause the database to deadlock when they are performed in single-node mode.  The issue is that Berkley leaves locks on the queue intact so long as the cursor is associated with a transaction.  In our queue code, we maintained api compatibility between the bdb-layer and the sql layer by creating a transaction, and substituting the sql's curtran (which is a readonly lockid) for the original transaction id.  This pull request changes cursor-close: it forces the release of the lock held by the cursor if the lockid is marked as READONLY.


markhannum:
OKAY .. dialing this back a bit ..

PR #2331, opened 2020-07-09
block tpt creation that violates constraints limits

dorinhogea:

Creating a time partition on a table Ta that has a foreign key constraint on another table Tb can fail shard creation.  Since each shard registers its own constraint with the Tb, the Tb might run out of reverse constraints space (which is currently hardcoded all over the code base to 32).  This patch check at time partition creation that this simple scenario does not happen.

### DIFF
--- a/db/views_serial.c
+++ b/db/views_serial.c
@@ -478,13 +478,37 @@ static int _views_do_partition_create(void *tran, timepart_views_t *views,
                  first_shard);
         goto error;
     }
+
+    /* constraints from time partition to other tables are only possible if the existing
+       reverse constraints for the other tables plus the total number of shards is less
+       then MAXCONSTRAINTS */
+    if (db->n_constraints > 0 ) {
+        struct dbtable *rev_db;
+        int i,j;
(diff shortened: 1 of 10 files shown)


dorinhogea:
Indeed, there is more than one scenario that can get an rollout error due to rev constraints violation.  
The syntax is too powerful, we have tables with multiple constraints and each constraint can have multiple rules, each rule targeting possibly a different table.
Implementation wise, constraints limits are hardcoded, and there are db pointers hard-locking structures to one another, which we might wanna change in the future.  Counting the rev constraints get tricky, and there is a simpler alternative at this point.  Since all these corner cases happen during fillout phase, it is easy to avoid the issues by creating a time partition in the past and have shards generated at the time of time partition creation.  This prevents a rollout issue in the future.
That being said, we will review options at a future PR (might be preferable to get rid of the hardcoded constraint limit instead of working around it).

dorinhogea:
Thanks for the review!

dorinhogea:
simple_timepart failure is a different issue, for which I will submit a separate PR (testing various periods is  racy).

PR #2334, opened 2020-07-09
Fix for Smartbeats

riverszhang89:
Do not send "hung" smartbeats to a LUA consumer if it isn't consuming events in a timely manner.

(DRQS 160019348)


riverszhang89:

PR #2336, opened 2020-07-10
make sure not to race shard rollout with testing

dorinhogea:

Testing various rollout periods in "simple_timepart" races against the testing itself.  60 seconds preamble was enough for test2min rollout, but not for daily/weekly/etc (changed it to one hour).  Also, a month is not 30 days all the time, so in some months we get one extra rollout, and one extra shard, which breaks the test.


dorinhogea:
Thank you Adi

PR #2339, opened 2020-07-13
Create transactional versions of several llmeta functions

markhannum:
This is another piece of https://github.com/bloomberg/comdb2/pull/2295.  This pr provides transactional versions of several llmeta accessors.  The next pr will utilize these accessors from within systable code.

### DIFF
--- a/bdb/bdb_api.h
+++ b/bdb/bdb_api.h
@@ -545,9 +545,11 @@ int bdb_handle_dbp_hash_stat_reset(bdb_state_type *bdb_state);
 int bdb_close_temp_state(bdb_state_type *bdb_state, int *bdberr);
 
 /* get file sizes for indexes and data files */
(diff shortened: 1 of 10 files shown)


bdb/bdb_api.h
```diff
@@ -546,8 +546,11 @@ int bdb_close_temp_state(bdb_state_type *bdb_state, int *bdberr);
 
 /* get file sizes for indexes and data files */
 uint64_t bdb_index_size(bdb_state_type *bdb_state, int ixnum);
```
riverszhang89, line 548: Maybe we could get rid of the non-tran version (i.e., `bdb_index_size`, `bdb_data_size` and `bdb_queue_size`). They're not used anywhere in your patch. Doing so also makes it clear that these functions should be invoked transactionally.

markhannum: YES .. lemme rework this a bit .. thanks Rivers!

PR #2340, opened 2020-07-13
Tick more frequently in luabb_trigger_register()

riverszhang89:
If a trigger register request to the master times out after 1 second, we end up sending a "hung" heartbeat to the client.

The change fixes it.

(DRQS 160019348 20)


riverszhang89:
@akshatsikarwar Thank you for the review!

PR #2343, opened 2020-07-14
Introduce curtran_gettran, and utilize for a few db-layer functions

markhannum:
This is another chunk of https://github.com/bloomberg/comdb2/pull/2295 which introduces a wrapper function, curtran_gettran.  curtran_gettran produces a bdb-level tran_type, but replaces it's lockid with the thread's curtran lockid.  This checkin further utilizes these routines for a few tasks in the db layer.

### DIFF
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -11321,6 +11321,49 @@ out:
     thread_memdestroy_and_restore(&thread_oldm);
 }
 
+void bdb_lock_stats_me(bdb_state_type *bdb_state, FILE *out);
+
+void curtran_assert_nolocks(void)
+{
+    struct sql_thread *thd = pthread_getspecific(query_info_key);
+    if (!thd)
+        return;
+    uint32_t lockid = bdb_get_lid_from_cursortran(thd->clnt->dbtran.cursor_tran);
+    if (!lockid)
+        return;
+
+    int nlocks = bdb_nlocks_for_locker(thedb->bdb_env, lockid);
+    if (nlocks > 0) {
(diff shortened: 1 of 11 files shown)


markhannum:
The systable_locking failures are temporary and will be fixed in the next checkin.  It's due to the triggers systable acquiring the schema-lock while holding Berkley lock references to pages in the llmeta table.

markhannum:
Okay - second thought - will port the minimum required to get systable_locking to pass.

db/sqlglue.c
```diff
@@ -11321,6 +11321,48 @@ void stat4dump(int more, char *table, int istrace)
...
+}
+
+tran_type *curtran_gettran(void)
+{
+    int bdberr;
+    tran_type *tran = NULL;
+    struct sql_thread *thd = pthread_getspecific(query_info_key);
+    if (!thd)
+        return NULL;
+    uint32_t lockid = bdb_get_lid_from_cursortran(thd->clnt->dbtran.cursor_tran);
```
riverszhang89, line 11350: Better to check for NULL `cursor_tran` here.

db/sqlglue.c
```diff
@@ -11321,6 +11321,48 @@ void stat4dump(int more, char *table, int istrace)
...
+    tran = bdb_tran_begin(thedb->bdb_env, NULL, &bdberr);
+    bdb_get_tran_lockerid(tran, &tran->original_lid);
```
riverszhang89, line 11352: Check for NULL `tran` here.

PR #2344, opened 2020-07-14
build: Fix build on Fedora 32, gcc 10.1.1

akshatsikarwar:
/silent


akshatsikarwar:
Not sure how we were able to build prior - these are flat out bugs and without `extern` this should have produced multiple definition errors.

riverszhang89:
Will update smoketest docker images ...

riverszhang89:
Ubuntu updated to 20.04; CentOS updated to 8.2.2004.

PR #2349, opened 2020-07-15
Updated limits for dohsql 

dorinhogea:
Default limit for per queue max mem size lowered to 10MB down from effectively 1.5GB.  Limit number of parallel engines. Make sure we force dispatch parallel load for order by statements.

limit dohsql queue max size to 10MB; better tracing

order by parallel statement need force_dispatch

configurable max threads for dohsql

testcase

format


dorinhogea:
Thank you Mike

PR #2351, opened 2020-07-16
Modify systables to use curtran_gettran

markhannum:
This is the final checkin for https://github.com/bloomberg/comdb2/pull/2295.  It revises our systables to utilize a curtran_gettran transaction- a bdb transaction which has its lockid replaced with the thread's sql lockid.  As we cannot grab the schema-lock in systable code (because the thread calling this code can have any number of Berkley locks), we remove those calls.  Each of these cases is now be protected by the associated table's systable-lock.  This checkin additionally re-enables the stubbed-out parts of the systable_locking test.  In short, we ask the database for a list of all system-tables, and we do a "SELECT * LIMIT 10" against each of these in a loop while doing other operations (writes and schema-changes) against the database.

This checkin additionally adds the extralock.testopts file to the systable_locking, schemalk, and sc_rebuilds test.  This sets a tunable which forces any SQL that accesses a systable to grab "_systable_lock", and forces every schema change to acquire "_systable_lock" in write mode.  Although we generally don't want to lock any more than necessary, this option has been invaluable in identifying various instances of incorrect locking in our codebase.

### DIFF
--- a/sqlite/ext/comdb2/schistory.c
+++ b/sqlite/ext/comdb2/schistory.c
@@ -48,8 +48,13 @@ static int get_status(void **data, int *npoints)
     int rc, bdberr, nkeys;
     sc_hist_row *hist = NULL;
     struct sc_hist_ent *sc_hist_ents = NULL;
+    tran_type *trans = curtran_gettran();
+    if (trans == NULL) {
+        logmsg(LOGMSG_ERROR, "%s: cannot create transaction object\n", __func__);
+        return SQLITE_INTERNAL;
+    }
 
-    rc = bdb_llmeta_get_sc_history(NULL, &hist, &nkeys, &bdberr, NULL);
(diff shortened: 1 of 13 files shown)


sqlite/ext/comdb2/schistory.c
```diff
@@ -48,8 +48,9 @@ static int get_status(void **data, int *npoints)
     int rc, bdberr, nkeys;
     sc_hist_row *hist = NULL;
     struct sc_hist_ent *sc_hist_ents = NULL;
+    tran_type *trans = curtran_gettran();
```
nirbhayc, line 51: if (!trans) - error

riverszhang89: yes - the error check is also needed [here](https://github.com/bloomberg/comdb2/pull/2351/files#diff-1e5f67b586fef5a1387551dd66024cf5R197).

riverszhang89:
Thank you for breaking down the original patch into small chunks and making it super easy to understand!

PR #2353, opened 2020-07-16
Provide custom location for event log file

adizaimi:
Provide custom location for event log file by means of reql command:
`@send reql events file /abc/def/myfile.log`


adizaimi:
I was thinking about adding that ("location events /path/to/dir") at first, then thought about specifying file (was quicker to code too), and then I was thinking to maybe we can have both.
Once you roll from the specified file (either manually or you reached the roll size) then it goes back to creating a name as usual.

PR #2354, opened 2020-07-17
Prevent buffer overflow

nirbhayc:
When the output is truncated, snprinf returns the number of characters that WOULD HAVE BEEN written to the buffer.
The logic incorrectly used this return value to calculate remaining buffer size.

Also, updated log message to indicate truncated column list.


nirbhayc:
> Maybe consider using strbuf for generating the output?

That's a great suggestion Mike!

PR #2355, opened 2020-07-20
Set analyze threads to run with low priority

adizaimi:
In order to minimize normal db activity, set analyze threads to run with low priority at the os level.


db/sqlanalyze.c
```diff
@@ -287,11 +288,29 @@ static int sample_index_int(index_descriptor_t *ix_des)
...
+        logmsg(LOGMSG_WARN, "%s:failed to set nice value:'%s'\n", __func__, strerror(errno));
+        return;
+    }
+    struct sched_param param;
+    int policy;
+    pthread_getschedparam(pthread_self(), &policy, &param);
+    if(param.sched_priority != 0)
+        logmsg(LOGMSG_WARN, "%s:param.sched_priority %d != 0\n", __func__, param.sched_priority);
+#endif
+}
```
riverszhang89, line 305: Did you miss a `pthread_setschedparam` here?

riverszhang89:
I have concerns about this pull request.

1. An Operation System, or a Time-sharing Operating System to be exact, is usually quite good at scheduling. After all task-scheduling is what an OS is for. I'm not sure how useful the manual tuning via pthread_setschedparam is. [This article](https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/KernelProgramming/scheduler/scheduler.html) on Darwin Kernel Programming points out that their scheduling API isn't useful in general cases:

> This API is frequently used in multimedia applications to obtain real-time priority.

> This relatively simple API is not particularly useful for most developers. However, it may be beneficial if you are developing a graphical user interface for Darwin.

In fact, searching `pthread_setschedparam` on github shows no meaningful use of this function.

2. There seems to be better approaches out there. What problem do you see with analyze threads here? If they're using excess CPU, we could delay each of them for a random amount of time to spread out the load. If it's disk IO, we could apply the same technique or just use larger temp-table cache.

adizaimi:
Thanks for looking Rivers:
analyze has been sometimes blamed for 'making queries a bit slower'. The biggest player here is autoanalyze which can trigger anytime, possibly in moments when the DB or the system is under heavy load. This patch intends to use Linux's ability to schedule appropriately and lower the effect analyze has on the normal operation of the machine and the databases residing there. So instead of using random sleeps and backoff methods, we can try to be nice to the other threads on the machine. How effective this is in practice is a matter of how it is implemented in the various Linux kernels.

I don't know why other projects don't use this in github, but I can presume that most software wants threads to be running at the highest priority thay can--there are probably few cases where a thread wants to run at lower priority than the rest of the process. When we run hundreds of databases in the same machine, I expect Linux to do the right thing and run lower priority threads at a lower priority : ) and that would be a positive thing to lessen the impact on the other db threads.

Regardless, I can't find a reason why to run analyze at lowest priority would harm anything

riverszhang89:
I just don't think changing the priority of analyze threads is the answer to "why does analyze make queries slower". Yes analyze threads can be spawned at any time even when the db is under heavy load, but so do SQL threads, checkpoint threads, memp trickle threads and dozens of others. If this is a scheduling problem, why isn't the same complaint made to those threads? It appears that we have not yet fully understood the problem. As you pointed out, the effectiveness of changing thread priority is up to the OS. So somehow it feels more wrong to apply a technique that we don't have control of to a problem that we do not fully understand.

As for your last point, I can think of a bad case: Assuming the scheduling API effectively lowers analyze thread priority to the lowest, what if I write a lot of data and then immediately execute a read load against the new data, expecting analyze to get updated statistics of the data change to give me the optimal query plans? Wouldn't your change delay the update of the statistics and cause sub-optimal query plans for my subsequent read load?

adizaimi:
I don't think analyze makes queries slower, however in a busy system it will compete with sql. So will mptrickle and other threads which arguably can also be set to lower priority.

The scenario that this is aimed at is analyze threads across *many* databases doing a lot of processing in a busy system. I think I do understand the problem, one DB needs to play nice to other DBs on the same machine. That's the point of different scheduling priorities, no?

It obviously is good to have high priority and low priority threads working in a busy system: analyze is a low importance thread--there is no need for it to compete at a scheduling level with sql and other threads (again, not necessarily within same DB). The example that you bring up is not really a problem because autoanalyze thresholds are picked arbitrarily (trigger analyze when x% of the the data changes) and if that is delayed a bit is not consequential. However, many analyze threads running across many databases can be a real problem (ex. data-upload backends across many DBs in same machine). While I don't think it is possible to lower the IO priority of a thread (ionice is per process I understand), I would have liked to be able to have analyze threads run with lower io priority as well.

I'm simply trying to use the OS's scheduling capabilities by *telling* it which threads are less important then others. If there is more threads we can put in low priority mode, we probably should.

db/sqlanalyze.c
adizaimi: Yes, first iteration used pthread_setschedparam, but incidentally all threads have sched_priority of 0, so it seems I would need to change the policy to SCHED_IDLE to achieve something similar to nice().

riverszhang89:
Because you said that analyze has been sometimes blamed for "making queries a bit slower", I guess the question is, does it make queries slower? If so, how?

I'm not disagreeing with your view on scheduling policies and priorities. But again, what you said about analyze is kind of universal: In a busy system, analyze competes with SQL. So do other threads. Analyze can be doing a lot of processing across many databases in a busy system. So do other threads. So why is analyze singled out if the same applies to other types of threads? That's why I feel that we have not fully understood the claim yet.

The example I gave was in fact a real word usage. We've had customers who'd turn on auto-analyze, do drastic data change and then query against the new data, expecting auto-analyze to give them immediate optimal results. If the delay in updating statistics caused by setting the priority of the analyze thread to the lowest isn't consequential (too short to notice, I assume), what makes me believe that analyze could effectively avoid competing with SQL, using the same technique? Admittedly their use case is a bit too harsh but it at least demonstrates some uncertainty of your change.

Often times, analyze, specifically auto-analyze, is just an effect, not the cause. As you pointed out, 

> However, many analyze threads running across many databases can be a real problem (ex. data-upload backends across many DBs in same machine).

How do we know having too many data-update backends isn't a problem, but having many analyze threads, which is apparently caused by the former, is?

PR #2360, opened 2020-07-22
Don't run recover deadlock from sql_tick in the prepare phase

markhannum:
Very simple change in sql_tick logic: do not call recover_deadlock if the sqlite engine is in the prepare phase.  This will alleviate some assert failures that we are seeing on beta machines.

### DIFF
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -625,22 +625,22 @@ static int sql_tick(struct sql_thread *thd)
     if ((rc = check_recover_deadlock(clnt)))
         return rc;
 
-    if ((gbl_epoch_time - clnt->last_sent_row_sec) >=
-        gbl_delay_sql_lock_release_sec) {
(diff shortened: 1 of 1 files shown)


db/sqlglue.c
```diff
@@ -625,22 +625,22 @@ static int sql_tick(struct sql_thread *thd)
     if ((rc = check_recover_deadlock(clnt)))
         return rc;
 
-    if ((gbl_epoch_time - clnt->last_sent_row_sec) >=
-        gbl_delay_sql_lock_release_sec) {
+    if (clnt->no_transaction == 0) {
```
riverszhang89, line 628: I don't mean to nitpick but `if (!clnt->no_transaction)` seems slightly better here since `no_transaction` is a boolean flag.

markhannum: oh .. yes .. okay .. i read this late (and already merged) but i can do this ..

PR #2362, opened 2020-07-22
Optionally verify record count

riverszhang89:
Make record count verification optional to temporarily suppress our recnums index corruption tickets.

(DRQS 144891579)


riverszhang89:
@mistachkin Thank you Joe!

PR #2363, opened 2020-07-22
longblockop: Revert back to old logic

nirbhayc:
Do not abort() in case sequence is not found in blkseq; return RC_TRAN_CLIENT_RETRY instead.

{DRQS 158155375}


nirbhayc:
@markhannum thank you for pointing out the issue!

nirbhayc:
Ported to master branch.

PR #2366, opened 2020-07-24
Send constraint errors to comdb2sc

adizaimi:
Currently constraint errors are not being reported by comdb2sc on alters (ex. when foreign table does not exist), and so this patch fixes that by sending the error back if s->sb is set.
Fix will make error look like this:
> ./comdb2sc.linux.tsk adidb alter t2 ~/b.csc2
Constraint error for table "t2" key "$KEY_A44A20B" -> <"nonexistent":"I1">: foreign table not found
Failed to process schema!
Schema change failed

### DIFF
--- a/db/constraints.c
+++ b/db/constraints.c
@@ -1976,12 +1976,20 @@ static void constraint_err(struct schema_change_type *s, struct dbtable *db,
                   "<\"%s\":\"%s\">: %s",
                   db->tablename, ct->lclkeyname, ct->table[rule],
                   ct->keynm[rule], err);
-    } else
+    } else { 
+        if (s && s->sb)
+          sc_errf(s,
+                  "Constraint error for table \"%s\" key \"%s\" -> "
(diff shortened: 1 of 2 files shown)


nirbhayc:
> The sc_rebuilds failures are probably due to the changed schema seed trace in this PR though.

Yea, apparently they are.

adizaimi:
Thank you for looking, yes fixing the silly printf. The comdb2sc change is in a separate PR. I will test it with -y only.

adizaimi:
with the comdb2sc changes, this is what the error from dryrun will look like:
./comdb2sc.linux.tsk -y -v adidb alter t2 ~/b.csc2
Connected to master node 7822
constraint error for table "t2" key "$KEY_A44A20B" -> <"nonexistent":"I1">: foreign table not found
Schema change failed

adizaimi:
Thank you for looking Rivers and Nirbhay!

PR #2367, opened 2020-07-26
cdb2api bug fixes: pmux, garbage config and the local tier

riverszhang89:
The change fixes 3 bugs in cdb2api.

1. Unnecessary pmux lookup in DIRECT-CPU mode
In DIRECT-CPU mode the API always looks up the database port. This costs an unnecessary round trip to pmux when there is already a cached connection in the sockpool. We should do so only when the API needs to make a connection on its own.

2. Garbage configuration file
The API attempts to process "(null).cfg". This is caused by passing a NULL database name to snprintf when formatting the database configuration file path.

3. The "local" tier does not mean localhost
The API performs dbinfo query and pulls all nodes in the cluster regardless of tier. However dbinfo should be skipped when "local" is specified as tier.

### DIFF
--- a/cdb2api/cdb2api.c
+++ b/cdb2api/cdb2api.c
@@ -1482,6 +1482,14 @@ static int cdb2_dbinfo_query(cdb2_hndl_tp *hndl, const char *type,
 static int get_config_file(const char *dbname, char *f, size_t s)
 {
     char *root = getenv("COMDB2_ROOT");
+
+    /* `dbname' is NULL if we're only reading defaults from comdb2db.cfg.
+       Formatting a NULL pointer with %s specifier is undefined behavior,
+       and whatever file path it produces (in glibc, it's "(null).cfg")
+       does not make sense. Return an error. */
+    if (dbname == NULL)
+        return -1;
+
     if (root == NULL)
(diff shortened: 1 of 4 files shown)


cdb2api/cdb2api.c
```diff
@@ -6165,7 +6142,10 @@ int cdb2_open(cdb2_hndl_tp **handle, const char *dbname, const char *type,
         strcpy(hndl->policy, "room");
     } else {
         hndl->flags |= CDB2_RANDOMROOM;
-        strcpy(hndl->policy, "random_room");
+        /* DIRECTCPU mode behaves like RANDOMROOM. But let's pick a shorter policy name
+           for it so we can fit longer hostnames in the sockpool type string which is
+           merely 48 chars. */
+        strcpy(hndl->policy, (hndl->flags & CDB2_DIRECT_CPU) ? "dc" : "random_room");
```
mohitkhullar, line 6148: can have 'hostname' instead of 'dc'  for direct cpu.. to store it in sockpool

mohitkhullar:
looks good

riverszhang89:
Thank you Mohit!

PR #2368, opened 2020-07-26
Provide 'nextsequence' as a default dbstore type

markhannum:
This is a new feature which allows programmers to specify a dbstore value of NEXTSEQUENCE.  If this value is not specified for an inserted record, then it will default to one larger than the largest value which the column has ever held.  The largest value is stored as a record in llmeta which is modified in-line with the transaction which is changing it.  This differs from non-transactional sequences in that there will be no gaps.  This pr contains a  'nextsequence' test which verifies that nextsequence defaults behave correctly under various circumstances.


markhannum:
The PR is a bit long- I may break this up into smaller pieces

riverszhang89:
Since you're adding a new serial type, could you update [the `sqlite3IsFixedLengthSerialType` macro](https://github.com/bloomberg/comdb2/blob/master/sqlite/src/vdbeInt.h#L799) and [the comment here](https://github.com/bloomberg/comdb2/blob/master/sqlite/src/vdbeaux.c#L3683)?

sqlite/src/vdbeInt.h
```diff
@@ -796,6 +798,6 @@ int sqlite3LockStmtTables(sqlite3_stmt *);
 
 Mem* sqlite3GetCachedResultRow(sqlite3_stmt *pStmt, int *nColumns);
 
-#define sqlite3IsFixedLengthSerialType(t) ( (t)<12 || ((unsigned int)t)==SQLITE_MAX_U32 || ((unsigned int)t)==(SQLITE_MAX_U32-1) )
+#define sqlite3IsFixedLengthSerialType(t) ( (t)<12 || ((unsigned int)t)==SQLITE_MAX_U32 || ((unsigned int)t)==(SQLITE_MAX_U32-1) || ((unsigned int)t)==(SQLITE_MAX_U32-2) )
```
riverszhang89, line 801: I wonder maybe we should just change it to `( (t)<12 || ((unsigned int)t) >= (SQLITE_MAX_U32-2) )`.

sqlite/ext/comdb2/columns.c
```diff
@@ -70,7 +71,8 @@ static int systblColumnsConnect(
                                 "varinlinesize,"
                                 "defaultvalue,"
                                 "dbload,"
-                                "isnullable)");
+                                "isnullable,"
+                                "nextsequence)");
```
nirbhayc, line 75: Wondering if we should name it ‘lastsequence’ instead?

docs/src/sqlitegen/bubble-generator-data.tcl
```diff
@@ -770,6 +773,7 @@ stack
                               {line {opt SET DATA} TYPE column-type }
                               {line SET DEFAULT expr }
                               {line DROP DEFAULT }
+                              {line DROP AUTOINCREMENT }
```
nirbhayc, line 776: Revert changes in all the bubble diagrams that are not touched.

sqlite/ext/comdb2/columns.c
nirbhayc: Also update docs to reflect this new column.

sqlite/src/comdb2build.c
```diff
@@ -2757,6 +2759,22 @@ static const char *type_comdb2_str[] = {TYPE_MAPPING};
...
 
+static inline int validAutoIncrementColumnCheck(struct comdb2_column *column)
+{
+    if ( column->type == SQL_TYPE_LONGLONG )
+        return 0;
+    if ( gbl_permit_small_sequences ) {
+        switch ( column->type ) {
+            case SQL_TYPE_SHORT:
+            case SQL_TYPE_INT:
+            case SQL_TYPE_LONGLONG:
```
nirbhayc, line 2770: This one’s superfluous.

sqlite/src/vdbeaux.c
```diff
@@ -3679,13 +3679,14 @@ int sqlite3VdbeCursorMoveto(VdbeCursor **pp, int *piCol){
...
 **     11                                  datetime
+**    SQLITE_MAX_U32-2                     nextsequence
 **    SQLITE_MAX_U32-1                     intervaldsus
 **    SQLITE_MAX_U32                       datetimeus
 **
-** The (SQLITE_MAX_U32-1) and SQLITE_MAX_U32 were added in R6.  Prior
-** versions of Comdb2 will not understand those serial types.  Note for
-** future expansion: When introducing new serial types greater than 11,
-** please update sqlite3IsFixedLengthSerialType accordingly.
+** (SQLITE_MAX_U32-2) was added in 7.0.  (SQLITE_MAX_U32-1) and SQLITE_MAX_U32
```
nirbhayc, line 3686: Added in 8.0 :)

tests/auth.test/t09.expected
```diff
@@ -114,6 +115,7 @@
 (candidate='LZ4')
 (candidate='MATCH')
 (candidate='NATURAL')
+(candidate='NEXTSEQUENCE')
```
nirbhayc, line 118: Drop this and only keep autoinc?

markhannum: It ends up conflicting with sqlite's autoincrement- I will document this function though.

PR #2371, opened 2020-07-28
Add ability to autofill a column with guid at record insertion time

adizaimi:
This feature is similar to autoincrement, except that the value assigned is a guid (which is stored as byte[16]). 
To have an autoasigned guid column, create table with a byte[16] field with dbstore set to "GUID",
like the following schema:
    
```
schema
{
    byte        auto_guid[16] dbstore=GUID
    int         i
}
```
Or with DDL:
```
> cdb2sql adidb 'create table t1(auto_guid byte(16) DEFAULT GUID, i int)'
> cdb2sql adidb 'insert into t1(i) values(1),(2)'
> cdb2sql adidb 'select guid_str(auto_guid) as str, * from t1'
(str='9341d907-c843-4ec6-844f-2a461379c3e1', auto_guid=x'9341d907c8434ec6844f2a461379c3e1', i=1)
(str='ed21148a-85af-4368-ae46-0787f900ee6e', auto_guid=x'ed21148a85af4368ae460787f900ee6e', i=2)
```


riverszhang89:
I think we should disallow creating a GUID bytearray other than 16 bytes, instead of allowing such a schema but silently ignoring the dbstore setting.

sqlite/src/parse.y
```diff
@@ -1329,6 +1329,10 @@ term(A) ::= CTIME_KW(OP). {
   A = sqlite3ExprFunction(pParse, 0, &OP, 0);
 }
 
+term(A) ::= GUID_FUNC(OP). {
```
mistachkin, line 1332: Could you please add the special comdb2 %ifdef / %endif pair here to help the automated tooling?

sqlite/tool/mkkeywordhash.c
```diff
@@ -230,6 +230,9 @@ static Keyword aKeywordTable[] = {
   { "GLOB",             "TK_LIKE_KW",      ALWAYS                 },
   { "GROUP",            "TK_GROUP",        ALWAYS                 },
   { "GROUPS",           "TK_GROUPS",       WINDOWFUNC             },
+#if defined(SQLITE_BUILDING_FOR_COMDB2)
```
mistachkin, line 233: Since this is an entirely new keyword (and not an "override" or an existing one), could you please move it to the comdb2 keywords section (below)?

tests/guid.test/runit
```diff
@@ -91,5 +91,65 @@ mult=`grep "^count" sel2.out | grep -v "=1" | wc -l`
...
+    byte        auto_guid[17] dbstore="GUID_FUNC"
+    int         i
+} ' &> create1.txt
+
+
+
+cdb2sql ${CDB2_OPTIONS} $dbnm default 'create table t2  { 
+schema 
+{
+    byte        auto_guid[16] dbstore="GUID"
```
nirbhayc, line 108: Shouldn't this be a keyword? IMO, we should differentiate between a literal and a keyword used in this context.

`dbstore = "GUID"` should store literal "GUID" .
`dbstore = GUID` should generate one.

This should also apply to CURRENT_TIMESTAMP, I think.

csc2/macc_so.c
```diff
@@ -3143,27 +3120,26 @@ static int dyns_get_field_option_comn(char *tag, int fidx, int option,
...
                     return 0;
-                } else if (*value_type == CLIENT_BYTEARRAY &&
-                           vbsz >= tables[tidx].sym[fidx].szof) {
+                } else if (*value_type == CLIENT_BYTEARRAY && vbsz >= tables[tidx].sym[fidx].szof && 
+                           tables[tidx].sym[fidx].szof == 16 && 
+                           (strcmp(tables[tidx].sym[fidx].fopts[i].value.strval, "GUID") == 0 ||
+                           strcmp(tables[tidx].sym[fidx].fopts[i].value.strval, "GUID_FUNC") == 0)) {
+                    memcpy(valuebuf, tables[tidx].sym[fidx].fopts[i].value.strval,
+                           strlen(tables[tidx].sym[fidx].fopts[i].value.strval));
+                    *value_sz = strlen(tables[tidx].sym[fidx].fopts[i].value.strval);
```
nirbhayc, line 3140: length can be cached in a local var.

nirbhayc: what happens when I do `guid byte[17] dbstore="GUID"`?

nirbhayc: perhaps a test for that?

adizaimi:
> I think we should disallow creating a GUID bytearray other than 16 bytes, instead of allowing such a schema but silently ignoring the dbstore setting.

Good idea, doing so.

sqlite/tool/mkkeywordhash.c
adizaimi: Done.

db/tag.c
```diff
@@ -5174,6 +5174,11 @@ static int init_default_value(struct field *fld, int fldn, int loadstore)
             if (strncasecmp(typebuf, "CURRENT_TIMESTAMP", 17) == 0)
                 is_null = 1; /* use isnull flag for current timestamp since
                                 null=yes is used for dbstore null */
+        } else if(opttype == CLIENT_BYTEARRAY) {
+            opttype = CLIENT_CSTR;
+            if (strncasecmp(typebuf, "GUID", 4) == 0)
```
nirbhayc, line 5179: we have used strcmp() (case sensitive) in macc_so.c. Should change this to use that as well? IIUC csc2 parsing is mostly case sensitive.

tests/guid.test/runit
adizaimi: Currently that gives error because csc2 format needs quotes. We can fix that later I suppose.

csc2/macc_so.c
adizaimi: Yes, this can all be cached--but did not want to pollute in this PR. 
Updated with failure message--can actually test for that.

db/tag.c
adizaimi: Good point -- probably strcmp is fine.

csc2/macc_so.c
nirbhayc: I meant, you could hold `strlen(tables[tidx].sym[fidx].fopts[i].value.strval))` into a var instead of calling it twice.

csc2/macc_so.c
```diff
@@ -3143,27 +3128,26 @@ static int dyns_get_field_option_comn(char *tag, int fidx, int option,
...
                     return 0;
-                } else if (*value_type == CLIENT_BYTEARRAY &&
-                           vbsz >= tables[tidx].sym[fidx].szof) {
+                } else if (*value_type == CLIENT_BYTEARRAY && vbsz >= tables[tidx].sym[fidx].szof && 
+                           tables[tidx].sym[fidx].szof == 16 && 
+                           (strcmp(tables[tidx].sym[fidx].fopts[i].value.strval, "GUID") == 0 ||
+                           strcmp(tables[tidx].sym[fidx].fopts[i].value.strval, "GUID_FUNC") == 0)) {
```
nirbhayc, line 3145: As we wouldn't be back-porting it, it'd also be good to add CHECK_LEGACY_SCHEMA here.

csc2/macc_so.c
adizaimi: oh, good point. thanks.

db/tag.c
adizaimi: Done.

sqlite/src/parse.y
adizaimi: Done. Thanks.

csc2/macc_so.c
adizaimi: Done.

csc2/macc_so.c
adizaimi: Done.

sqlite/src/func.c
```diff
@@ -2700,6 +2700,7 @@ void sqlite3RegisterBuiltinFunctions(void){
     VFUNCTION(guid_str,          0, 0, 0, guidStrFunc      ),
     FUNCTION(guid,               1, 0, 0, guidFromStrFunc  ),
     FUNCTION(guid_str,           1, 0, 0, guidFromByteFunc ),
+    VFUNCTION(guid_func,         0, 0, 0, guidFunc         ),
```
nirbhayc, line 2703: Why do we need this? We already have `VFUNCTION(guid, 0, 0, 0, guidFunc),` wouldn't that be sufficient?

sqlite/src/func.c
```diff
@@ -2700,6 +2700,7 @@ void sqlite3RegisterBuiltinFunctions(void){
     VFUNCTION(guid_str,          0, 0, 0, guidStrFunc      ),
     FUNCTION(guid,               1, 0, 0, guidFromStrFunc  ),
     FUNCTION(guid_str,           1, 0, 0, guidFromByteFunc ),
+    VFUNCTION(libuuid_call,      0, 0, 0, guidFunc         ),
```
nirbhayc, line 2703: I don't get the point behind adding this one. We already have 'guid' vfunc registered above.

adizaimi:
A better and more generic version of this is at #2383.

PR #2372, opened 2020-07-28
Return head and tail ages for queues, limit scans

mponomar:
Cap the scan depth for comdb2_queues.  Stop after a configurable number of records, and report that as the queue depth.  Prevents long running scans when there are very deep queues.  Also reports the ages of the newest and oldest item on the queue.  For queues that have odh we still report the full queue depth, since that's available without a scan.

### DIFF
--- a/bdb/queuedb.c
+++ b/bdb/queuedb.c
@@ -636,7 +636,7 @@ int bdb_queuedb_stats(bdb_state_type *bdb_state,
     DBC *dbcp2 = NULL;
     uint8_t ver = 0;
     size_t item_length = 0;
-    unsigned int epoch = 0, first_seq = 0, last_seq = 0;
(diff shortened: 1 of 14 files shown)


bdb/queuedb.c
```diff
@@ -1420,3 +1426,79 @@ int bdb_trigger_unpause(bdb_state_type *bdb_state)
...
+    } else {
+        struct bdb_queue_found qfnd;
+        p_buf = (uint8_t *) queue_found_get(&qfnd, p_buf, p_buf_end);
+        if (p_buf)
+            *epoch = qfnd.epoch;
+    }
+    if (p_buf == NULL) {
+        // failed to decode?
+        *bdberr = BDBERR_MISC;
+        return -1;
```
mistachkin, line 1487: Should this be "goto done" here?

mponomar: Thanks, good catch.

PR #2374, opened 2020-07-28
Address deadlocks for tests running in standalone mode

markhannum:
Berkley will leave locks on a B-tree after a cursor is closed if there is a transaction associated with the cursor.  This is true even for curtran readonly transactions.  We can get around this by creating a cursor using a NULL transaction, and then replacing the lockid in that cursor with the c_replace_lockid function (as there is no berkley transaction, the close-cursor function discards it's pagelocks).  This PR modifies the Berkley access routines in bdb/lite.c such that curtran_gettran transactions (used only from an SQL thread) will not leave pages locked when the cursor is closed.

### DIFF
--- a/bdb/lite.c
+++ b/bdb/lite.c
@@ -50,6 +50,22 @@
 #include "locks.h"
 #include <logmsg.h>
 
+static int get_cursor_lite(bdb_state_type *bdb_state, tran_type *tran, DBC **outdbcp)
+{
+    DB *db = bdb_state->dbp_data[0][0];
+    DBC *dbcp = NULL;
+    int rc;
+
+    if (tran && tran->is_curtran) {
+        rc = db->cursor(db, NULL, &dbcp, 0);
+        dbcp->c_replace_lockid(dbcp, tran->tid->txnid);
+    } else {
+        rc = db->cursor(db, tran ? tran->tid : NULL, &dbcp, 0);
+    }
+    (*outdbcp) = dbcp;
+    return rc;
+}
+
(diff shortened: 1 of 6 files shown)


bdb/lite.c
```diff
@@ -377,8 +410,7 @@ static int bdb_lite_fetch_keys_int(bdb_state_type *bdb_state, tran_type *tran,
         return -1;
     }
 
-    db = bdb_state->dbp_data[0][0];
-    rc = db->cursor(db, tid, &dbcp, 0);
+    rc = get_cursor_lite(bdb_state, tran, &dbcp);
```
adizaimi, line 413: can you check below codepaths? seems we return from this function on some errors without closing the cursor.

markhannum: AH THANK YOU ADI!

bdb/lite.c
```diff
@@ -668,16 +700,13 @@ int bdb_lite_list_records(bdb_state_type *bdb_state,
...
     DBC *dbcp;
     DBT key, data;
     int rc = 0;
 
     *bdberr = 0;
 
-    db = bdb_state->dbp_data[0][0];
-
-    rc = db->cursor(db, NULL, &dbcp, 0);
+    rc = get_cursor_lite(bdb_state, NULL, &dbcp);
```
adizaimi, line 709: Same here -- we should close always after c_get()?

markhannum: Wait, I think those get closed at the end of the function, no?

PR #2375, opened 2020-07-29
Small cleanup in dyns_get_table_field_option function

adizaimi:
In this cleanup:
- Refactor function to be more readable
- Use pointer instead of repeat structure-member-dereferencing train
- Reuse length of string instead of recomputing it

### DIFF
--- a/csc2/macc_so.c
+++ b/csc2/macc_so.c
@@ -75,9 +75,6 @@ static int dyns_get_field_arr_dims_comn(char *tag, int fidx, int *dims,
                                         int ndims, int *nodims);
 static int dyns_field_depth_comn(char *tag, int fidx, dpth_t *dpthinfo,
                                  int ndpthsinfo, int *ndpthout);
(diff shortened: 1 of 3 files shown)


csc2/macc_so.c
```diff
@@ -3044,142 +3040,118 @@ int dyns_get_table_field_option(char *tag, int fidx, int option,
...
+            if (*value_type == CLIENT_BYTEARRAY && vbsz >= length) {
+                memcpy(valuebuf, bytes, length);
+                *value_sz = length;
+                return 0;
             }
-            default:
+            return -1;
+        }
+        case CLIENT_CSTR: {
+            if (*value_type == CLIENT_BYTEARRAY && vbsz >= sym->szof) {
```
adizaimi, line 3126: changed the order of the ifs here, the one on the left is done a little further down on the right.

adizaimi:
Thank you Joe!

PR #2379, opened 2020-07-31
Update SQLite to the 3.33.0 release.

mistachkin:


db/sqlglue.c
```diff
@@ -8194,15 +8207,15 @@ int sqlite3BtreeCursor(
...
+        Pgno pgno = iTable;
         if (forOpen) {
             /*
             ** NOTE: When being called to open a temporary (table) cursor in
             **       response to OP_OpenRead, OP_OpenWrite, or OP_ReopenIdx
             **       opcodes by the VDBE, the temporary table may not have
             **       been created yet.  Attempt to do that now.
             */
-            if (hash_find(pBt->temp_tables, &pgno) == NULL) {
+            if (hash_find(pBt->temp_tables, &iTable) == NULL) {
```
dorinhogea, line 8218: pgno?

sqlite/src/status.c
```diff
@@ -351,7 +364,7 @@ int sqlite3_db_status(
     */
     case SQLITE_DBSTATUS_CACHE_SPILL:
       op = SQLITE_DBSTATUS_CACHE_WRITE+1;
-      /* Fall through into the next case */
+      /* no break */ deliberate_fall_through
```
dorinhogea, line 367: how does this compile?

db/sqlglue.c
mistachkin: Technically, it's not of the right type (even though both variables should have the same value at this point.

sqlite/src/status.c
mistachkin: The "deliberate_fall_through" thing is a macro that either expands to nothing or to a special GCC construct that disables the compiler warning.

dorinhogea:
I would suggest sticking with sqlite_master, to avoid rolling out issues. Tools and cross-db sql will break.

PR #2384, opened 2020-08-03
Disallow out of range longlong values.

mponomar:
It's possible to use SQL to get u_longlong values that are perfectly valid, in range u_longlong values but are not valid longlong values and therefore cannot be read back with SQL.  Provide an option to disallow that.

### DIFF
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -1002,6 +1002,17 @@ static int mem_to_ondisk(void *outbuf, struct field *f, struct mem_info *info,
             NULL /*blob */, out + f->offset, f->len, f->type, 0, &outdtsz,
             &f->convopts, NULL /*&outblob[nblobs] blob */);
     } else if (m->flags & MEM_Real) {
+        /* Prevent u_longlong fields from getting out of range (for longlong value).
+         * It's still possible to get them in with tags, but this prevents a situation
+         * where SQL lets you insert a value you can't subsequently read back.
+         * SQLite switches to double for values with numeric affinity that are out of range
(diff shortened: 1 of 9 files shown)


db/sqlglue.c
```diff
@@ -1002,6 +1002,17 @@ static int mem_to_ondisk(void *outbuf, struct field *f, struct mem_info *info,
             NULL /*blob */, out + f->offset, f->len, f->type, 0, &outdtsz,
             &f->convopts, NULL /*&outblob[nblobs] blob */);
     } else if (m->flags & MEM_Real) {
+        /* Prevent u_longlong fields from getting out of range (for longlong value).
+         * It's still possible to get them in with tags, but this prevents a situation
+         * where SQL lets you insert a value you can't subsequently read back.
+         * SQLite switches to double for values with numeric affinity that are out of range
+         * for int, regardless of the source type, so we only need to cover the MEM_Real case. */
+        if (gbl_disallow_sql_ull_values && f->type == SERVER_UINT && f->len == 9 &&
+                m->u.r < ULLONG_MAX && (uint64_t)m->u.r > LLONG_MAX) {
```
nirbhayc, line 1011: Is `m->u.r < ULLONG_MAX` check superfluous?

mponomar: You're right, it is.  Conversion will fail for values out of that range.

nirbhayc (dismissed):
Leave it ON by default? Okay to merge.

mponomar:
I can leave it on by default and disable in legacy_defaults.

PR #2385, opened 2020-08-04
cdb2jdbc: turn on verify-retry and return per-statement effects

riverszhang89:
This is a compromise between Comdb2's OCC model and what JDBC users normally expect. With the verify-retry mechanism, one can't rely on per-statement query effects in a transaction. So we lose a bit of accuracy (harmlessly, I hope) to be more in line with the JDBC land.


akshatsikarwar:
Why don't we disable verify-retry instead? Remain correct, accurate and in-line with normal expectation?

riverszhang89:
> Why don't we disable verify-retry instead? Remain correct, accurate and in-line with normal expectation?

Folks expect both verify-retry and per-stmt effects.

PR #2386, opened 2020-08-04
Fix timer trap deadlock

riverszhang89:
"sync log-delete off \<N\>" deadlocks itself when re-enabling log deletion.

(DRQS 161450311)

### DIFF
--- a/tests/sync_log_delete_off_reenable.test/runit
+++ b/tests/sync_log_delete_off_reenable.test/runit
@@ -1,0 +1,21 @@
+# TODO consolidate all leak regression tests into one
+
+#!/usr/bin/env bash
+bash -n "$0" | exit 1
+
+dbnm=$1
+
+set -e
+
+# Make sure we talk to the same host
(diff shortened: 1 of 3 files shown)


tests/sync_log_delete_off_reenable.test/runit
```diff
@@ -0,0 +1,21 @@
...
+# Make sure we talk to the same host
+host=`cdb2sql --tabs ${CDB2_OPTIONS} $dbnm default 'SELECT comdb2_host()'`
+
+
+# Turn off log deletion and enable it back in 1 minute.
+cdb2sql $dbnm --host $host 'exec procedure sys.cmd.send("sync log-delete off 1")'
+
+# Give it 5 extra seconds here to paper over slow machines.
+sleep 65
+
```
adizaimi, line 19: Would you want to do some reads/writes before and after just to test sanity?

riverszhang89: Unfortunately we do not have a message trap for displaying the setting. Adding one would be out of the scope of this PR.

riverszhang89:
@mponomar @adizaimi Thank you for the review!

PR #2389, opened 2020-08-05
Abort when we cant convert from client to server type

adizaimi:
All types should have a conversion from client to server. In the case that we add new types, we should add conversion as well. 
So we should abort here to catch those cases.

### DIFF
--- a/db/tag.c
+++ b/db/tag.c
@@ -303,7 +303,7 @@ int client_type_to_server_type(int type)
     case CLIENT_SEQUENCE:
         return SERVER_SEQUENCE;
     default:
-        return type;
+        abort();
     }
 }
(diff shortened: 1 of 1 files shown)


db/tag.c
```diff
@@ -301,7 +301,7 @@ int client_type_to_server_type(int type)
     case CLIENT_VUTF8:
         return SERVER_VUTF8;
     default:
-        return type;
+        abort();
```
mponomar, line 304: Maybe assert() that this isn't reachable instead?

adizaimi: Asserts aren't enabled for normal builds afaik -- only enabled for DEBUG builds.

PR #2391, opened 2020-08-07
README instructions also give successful "make test" on Ubuntu 20.04, Debian 10

danshearer:
all-tests takes a *very* long time :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ On every machine in the cluster:
 
 2. Install prerequisites: 
    
-   **Ubuntu 16.04, 16.10, 17.04, Windows Subsystem for Linux (WSL)**
+   **Ubuntu 16.04, 16.10, 17.04, 20.04, Debian 10, Windows Subsystem for Linux (WSL)**
         
    ```
    sudo apt-get install -y  \


adizaimi:
Thanks for verifying and the PR. In fact all my local testing is done in Debian : ).
all-tests do indeed take very long, and longer as we add more tests.
We were considering to include a smaller subset for quicker basic testing.

danshearer:
> We were considering to include a smaller subset for quicker basic testing.

Many tests could be done in parallel with very little work.

adizaimi:
> > We were considering to include a smaller subset for quicker basic testing.
> 
> Many tests could be done in parallel with very little work.

That's what we do -- we may need to update docs to make things more clear.
How do you run tests currently? 

We simply 
```
cd tests/
make -j 8 -k 
```
Will run all tests in 8-way parallel. Currently should take abount 1.5hrs, depending on hardware.

mponomar:
You'll want to run `make test-tools` in the build directory first.

mponomar:
The test suite expects some of the tests to have a cluster, since we rarely run Comdb2 standalone. If you want a somewhat friendlier way to run tests,

- Copy contrib/dev-util/cluster somewhere in your PATH
- run `cluster init`.  You can leave defaults for all the prompts except the path to source.
- run `cluster build`, `cluster run`
- run `cluster test` to kick off tests

This works on ubuntu, with limited success on other distributions.  We use a similar procedure to run our tests (ie: docker + make test -j ...).

PR #2394, opened 2020-08-07
Make testrunner.c compile (ansi strictness), remove python script

adizaimi:
* testrunner.c in some systems fails to compile due to ansi strictness,
  so fixing Makefile and .c for that.
* Updating the tests/Readme how to use the testrunner to run tests.
* Deleted obsolete python script tests/tools/testrunner
    

adizaimi:
Thank you Mike!

PR #2396, opened 2020-08-08
Fix check connections race

mohitkhullar:
Fix check connections race

### DIFF
--- a/db/appsock_handler.c
+++ b/db/appsock_handler.c
@@ -265,7 +265,6 @@ static void appsock_work(struct thdpool *pool, void *work, void *thddata)
     struct appsock_thd_state *state = thddata;
     appsock_work_args_t *w = (appsock_work_args_t *)work;
     int keepsocket = 0;
-
     thrman_setfd(state->thr_self, sbuf2fileno(w->sb));
     thd_appsock_int(w, &keepsocket, state->thr_self);
     thrman_setfd(state->thr_self, -1);
@@ -314,6 +313,7 @@ void appsock_handler_start(struct dbenv *dbenv, SBUF2 *sb, int admin)
     static int last_thread_dump_time = 0;
     static int last_thread_dump_warn_time = 0;
     int nconns = 0;
+    int curr_appsock_conns = 0;
     int maxconns;
     time_t now;
 
@@ -385,11 +385,11 @@ void appsock_handler_start(struct dbenv *dbenv, SBUF2 *sb, int admin)
     total_appsock_conns++;
     Pthread_mutex_lock(&appsock_conn_lk);
     active_appsock_conns++;
+    curr_appsock_conns = active_appsock_conns;
     Pthread_mutex_unlock(&appsock_conn_lk);
-    if (active_appsock_conns >
-        bdb_attr_get(thedb->bdb_attr, BDB_ATTR_MAXSOCKCACHED)) {
-        logmsg(LOGMSG_WARN,
-               "TOO many active socket connections. current limit %d\n",
+
+    if (curr_appsock_conns > bdb_attr_get(thedb->bdb_attr, BDB_ATTR_MAXSOCKCACHED)) {
+        logmsg(LOGMSG_WARN, "TOO many active socket connections. current limit %d\n",
                bdb_attr_get(thedb->bdb_attr, BDB_ATTR_MAXSOCKCACHED));
     }
 }
--- a/db/sqlinterfaces.c
+++ b/db/sqlinterfaces.c
@@ -4621,60 +4621,61 @@ errors:
 
 */
 
+static int check_done_func(void *obj)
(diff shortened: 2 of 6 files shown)


db/sqlinterfaces.c
```diff
@@ -4629,8 +4629,11 @@ int check_active_appsock_connections(struct sqlclntstate *clnt)
         int num_retry = 0;
         int rc = -1;
     retry:
+        /* Don't count this connection until we are done with the check */
+        active_appsock_conns--;
```
mponomar, line 4633: Don't we need appsock_conn_lk here to modify active_appsock_conns? We can also go over the limit if the dirty check fails for a bunch of connections all at once.

mohitkhullar: yes. we have another check and then retry.

db/appsock_handler.c
```diff
@@ -265,6 +268,33 @@ static void appsock_work(struct thdpool *pool, void *work, void *thddata)
...
+    Pthread_mutex_lock(&appsock_conn_lk);
+    if (active_appsock_conns >= bdb_attr_get(thedb->bdb_attr, BDB_ATTR_MAXAPPSOCKSLIMIT)) {
+        if (close_lru_client(active_appsock_conns, NULL, bdb_attr_get(thedb->bdb_attr, BDB_ATTR_MAXAPPSOCKSLIMIT)) != 0) {
+            num_retry++;
+            if (num_retry < 5) {
+                Pthread_mutex_unlock(&appsock_conn_lk);
+                sleep(1);
+                goto retry; 
+            }
+            w->rejected = 1; /* Reject this query. */
```
mponomar, line 285: The only reason for rejected seems to be to force-allow admin connections.  We already know at this point that whether it's admin - we can skip this entire process for admin, and reject non-admin connections right at this point.

db/appsock_handler.c
```diff
@@ -265,6 +268,33 @@ static void appsock_work(struct thdpool *pool, void *work, void *thddata)
...
     appsock_work_args_t *w = (appsock_work_args_t *)work;
     int keepsocket = 0;
+    int curr_appsock_conns = 0;
+    w->rejected = 0;
+    int num_retry = 0;
+
+retry:
+    Pthread_mutex_lock(&appsock_conn_lk);
+    if (active_appsock_conns >= bdb_attr_get(thedb->bdb_attr, BDB_ATTR_MAXAPPSOCKSLIMIT)) {
+        if (close_lru_client(active_appsock_conns, NULL, bdb_attr_get(thedb->bdb_attr, BDB_ATTR_MAXAPPSOCKSLIMIT)) != 0) {
```
mponomar, line 278: close_lru_client doesn't actually close - it signals the appsock thread to close.  So even if we found a suitable connection, we still end up allowing more than the maximum number during a spike.

db/appsock_handler.c
```diff
@@ -265,6 +268,33 @@ static void appsock_work(struct thdpool *pool, void *work, void *thddata)
...
+            num_retry++;
+            if (num_retry < 5) {
+                Pthread_mutex_unlock(&appsock_conn_lk);
+                sleep(1);
```
mponomar, line 282: Maybe we should just reject and not retry?  The API will try on a different node, which is likely to be faster than waiting here.  The API already has logic to back off and retry.

db/sqlinterfaces.c
```diff
@@ -4621,60 +4621,52 @@ __func__, cnonce);
...
-                // lru_clnt->statement_timedout = 1; already done
-                shutdown(fd, SHUT_RD);
-                logmsg(
-                    LOGMSG_WARN,
-                    "%s: Closing least recently used connection fd %d , total "
-                    "%d \n",
-                    __func__, fd, active_appsock_conns - 1);
-                rc = 0;
-            }
+            shutdown(fd, SHUT_RD);
```
mponomar, line 4645: We always call this with clnt == NULL, so I don't think we can rely on this not finding our own connection.

mohitkhullar: Used with clnt here.. to make it work with other plugins https://github.com/bloomberg/comdb2/pull/2396/files#diff-4932981bbf2c2bd6bf0267d1a6d39173R4664

db/appsock_handler.c
mohitkhullar: We don't want client to keep trying and thus increasing the load, we have to give error back.

db/appsock_handler.c
mohitkhullar: Yes, the thread doesn't close immediately, however number of actual open connections decreases

db/appsock_handler.c
mohitkhullar: Yes, we have to give error back based on plugin

PR #2397, opened 2020-08-10
Auxiliary function to check if the user is OP

nirbhayc:
.


riverszhang89:
I'm triggering a roborivers build now.

PR #2398, opened 2020-08-10
Auxiliary function to check if the user is OP

nirbhayc:


nirbhayc:
Cherry-picked to master branch.

PR #2402, opened 2020-08-13
Move freeing of the array in FreeAllBuiltinFunctions

adizaimi:
For a cleaner cleanup of arr, move the freeing in same file as it was allocated.


adizaimi:
Thank you Joe.

PR #2406, opened 2020-08-13
Use macro to expand enum and representation of osql types

adizaimi:
This was left as a todo from PR #2224 (edfeafe9): use macros to expand osql types.


adizaimi:
Thank you Akshat.
Joe pointed out that the assert should check for <= 0 and I obliged.

PR #2410, opened 2020-08-14
Fix datacopy race in create_sqlmaster_record

markhannum:
This is a WIP fix for the datacopy race that was affecting ddisncal: create_sqlmaster_record modifies a schema's datacopy array "in-place", even for schema's that are not being schema-changed (and therefore do not have a writelock on the table).


markhannum:
The grace-period-on-exit core is a printf bug (it happens against master).

PR #2412, opened 2020-08-18
warning: Fix build on ubuntu 20.04 with clang 10.0

akshatsikarwar:
/silent

### DIFF
--- a/sqlite/ext/comdb2/schistory.c
+++ b/sqlite/ext/comdb2/schistory.c
@@ -99,9 +99,7 @@ static int get_status(void **data, int *npoints)
         char str[22];
         sprintf(str, "%0#16"PRIx64, hist[i].seed);
         sc_hist_ents[i].seed = strdup(str);
-
-        if (hist[i].errstr)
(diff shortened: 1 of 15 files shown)


sqlite/ext/comdb2/schistory.c
```diff
@@ -99,9 +99,7 @@ static int get_status(void **data, int *npoints)
         char str[22];
         sprintf(str, "%0#16"PRIx64, hist[i].seed);
         sc_hist_ents[i].seed = strdup(str);
-
-        if (hist[i].errstr)
-            sc_hist_ents[i].error = strdup(hist[i].errstr);
+        sc_hist_ents[i].error = strdup(hist[i].errstr);
```
adizaimi, line 102: won't this crash if errstr is null?

akshatsikarwar: It won't be null because it is not a pointer - clang was (correctly) warning that test is always true.

akshatsikarwar:
Thanks guys.

PR #2416, opened 2020-08-19
Fix timeout for tests: send kill signal if still running

adizaimi:
Also timeout on unsetup send commands.


adizaimi:
thank you Rivers!

PR #2417, opened 2020-08-19
Set max number of trigger threads

mohitkhullar:
Set max number of trigger threads/Fix fdb deadlock

Signed-off-by: Mohit Khullar <mkhullar1@bloomberg.net>


db/trigger.c
```diff
@@ -189,11 +191,25 @@ int trigger_unregister(trigger_reg_t *t)
 
 static void *trigger_start_int(void *name_)
 {
-    GET_BDB_STATE_CAST(bdb_state, void *);
+    static int num_trigger_threads = 0;
+    int allow_new_thread = 1;
     char name[strlen(name_) + 1];
+    trigger_reg_t *reg;
+
+    Pthread_mutex_lock(&trig_thd_cnt_lk);
```
akshatsikarwar, line 199: I think it is better if we move these checks into `trigger_start`

akshatsikarwar: That way you can do check before thread is even created.

db/fdb_fend.c
```diff
@@ -1960,7 +1960,13 @@ int fdb_cursor_move_master(BtCursor *pCur, int *pRes, int how)
...
 
 search:
-    __lock_wrlock_shared(fdb);
+    //__lock_wrlock_shared(fdb);
+
+    /* At this point fdb->users already is increased to 2 by this thread and
+     * there can be multiple threads waiting for users to decrease.
+     * One of them gets write lock while others wait for the write lock,
+     * so this thread will never get a read lock */
+    Pthread_rwlock_wrlock(&fdb->h_rwlock);
```
dorinhogea, line 1969: This is not right, it will serialize access to sqlite master.   users is incremented to allow lock free access, but not free the structure!  This is not a source of deadlock

mohitkhullar: Close to 1000 other thread  in _add_table_and_stats_fdb one of which has wrlock and other waiting at https://github.com/bloomberg/comdb2/blob/master/db/fdb_fend.c#L1476 , the one which has wrlock fails and gets back to the wait, leaving the lock for one of the 1000 theads. This thead never gets opportunity to proceed.

dorinhogea: How do we allow 1000 sqlite engines?

dorinhogea: Only one sqlite engine will get to populate a fdb cache, the rest are simply waiting for the result.  This is by design.  No need to have every engine try to grab same remote schema information.

mohitkhullar: Its lua trigger threads here, a new thread is created if the previous thread is stuck and not consuming queue
https://github.com/bloomberg/comdb2/blob/master/db/trigger.c#L132

mohitkhullar:
can be reproduced by
for i in {1..100}; do time cdb2sql db1 local "select name from db2.sqlite_master"   & done

mohitkhullar:
also with this
for i in {1..100}; do time cdb2sql drqs2db local "select * from mohitdb1.t join mohitdb1.t$i"   & done
where t$i don't exist

db/fdb_fend.c
dorinhogea: We found the deadlock culprit and opened a new PR for it https://github.com/bloomberg/comdb2/pull/2455.
What happens it, due to the fact that sqlite_master has no entry in fdb cache, the sqlite table hash remains unpopulated during the initial fdb creation.  When sqlite later on tries to access sqlite_stats, as part of the fdb creation and table attachment, it has no entry for sqlite_stats and trigger a nested fdb creation inside the fdb creation process.  This obviously deadlocks since the access is exclusive.
Lets remove the changes for locking in remsql and keep only the trigger limit for this PR.

mohitkhullar:
Opening new PR to preserve commits related to deadlock
https://github.com/bloomberg/comdb2/pull/2465

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>350/364 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>incremental_backup_load [setup core dumped]<br>sirace [setup core dumped]<br>queuedb_rollover [setup core dumped]<br>maxtable [setup core dumped]<br>sc_async_constraints_logicalsc_generated [setup core dumped]<br>timepart_retention1 [setup core dumped]<br>simple_timepart_reptimeout_generated [setup core dumped]<br>simple_timepart [setup core dumped]<br>selectv_deadlock [setup core dumped]<br>sc_timepart [setup core dumped]</p><p><a href='http://comdb2.s3-website-us-east-1.amazonaws.com/tests/f18a7af/detail.txt'>Detail</a> | <a href='http://comdb2.s3-website-us-east-1.amazonaws.com/tests/f18a7af/logs.tar.gz'>Log</a> | <a href='http://ec2-34-202-216-70.compute-1.amazonaws.com/restricted/index.php?id=2417'>Resubmit</a></p>

riverszhang89:
It was hung in roborivers. I had to kill it just now.

PR #2429, opened 2020-08-24
Add a knob to control subquery flattening optimization

nirbhayc:


riverszhang89:
I think appending "LIMIT -1" to a sub-query also disables the flattening optimization.

nirbhayc:
> I think appending "LIMIT -1" to a sub-query also disables the flattening optimization.

Yes, that it right. However, the problem looks general enough (see the included test) I thought it will be good it we have an option to control this specific optimization on the server side.

nirbhayc:
Thanks for the review @mistachkin !

nirbhayc:
Cherry-picked to master branch.

PR #2434, opened 2020-08-25
Fix remsql when the network/remote node fails.

dorinhogea:
Fix broken sequence in transaction that rollback early
    (crash in fdb_cursor_open, or otherwise access a freed dbtran->dtran)

Fixes for the retry in the first remote access (creation of fdb, populating stats):
    preserve rootpage/sql_hint while reopening the remote connection
    fix the sqlite stats retry.

Fix case when sqlite3LockStmtTables failed (for example, old table versions)
    sql cached engines were removed from hash but kept in the [no]param_stmt_list

Fix crashes when I/O happens in between remsql sessions (printing random memory area)

Add io error generator and testcase to eliminate side-effects of failing networks

### DIFF
--- a/bdb/file.c
+++ b/bdb/file.c
@@ -123,6 +123,8 @@ static const char NEW_PREFIX[] = "new.";
 
 static pthread_once_t ONCE_LOCK = PTHREAD_ONCE_INIT;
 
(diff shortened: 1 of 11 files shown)


dorinhogea:
Thank you

PR #2437, opened 2020-08-25
Eventlog: change reql events rollat to take MB instead of bytes

adizaimi:
Instead of 'reql events rollat N' to be in bytes, we should have
this in MB so it does not risk rolling every few sql stmts.

Added a debug tunable to set rollat in bytes for testing purposes.

### DIFF
--- a/db/eventlog.c
+++ b/db/eventlog.c
@@ -50,8 +50,7 @@ extern void cson_snap_info_key(cson_object *obj, snap_uid_t *snap_info);
 static char *gbl_eventlog_fname = NULL;
 static char *eventlog_fname(const char *dbname);
 int eventlog_nkeep = 2; // keep only last 2 event log files
-#define EVENTLOG_ROLLAT_MINSIZE 1024 * 2        // need minsize to avoid cleanup flood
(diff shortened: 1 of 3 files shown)


db/db_tunables.h
```diff
@@ -1261,6 +1261,9 @@ REGISTER_TUNABLE(
     "Deliberately allow insertion without constraint check to debug db_verify",
     TUNABLE_BOOLEAN, &gbl_debug_skip_constraintscheck_on_insert, INTERNAL, NULL,
     NULL, NULL, NULL);
+REGISTER_TUNABLE("debug.eventlog_rollat_size",
```
riverszhang89, line 1264: move this to [debug_switches](https://github.com/bloomberg/comdb2/blob/master/util/debug_switches.c)?

adizaimi: Hmm, we have way too many ways of setting debug parameters. Will do, thanks.

adizaimi: Removed it totally. Thanks for suggestion.

adizaimi:
Thanks  Rivers.

PR #2439, opened 2020-08-26
Eventlog: cleanup old log files outside of lock

adizaimi:
Move calling eventlog_roll_cleanup() (which keeps only the last n most recent
eventlog files) outside of lock in order to not hold up other sql engines which
want the lock to write to the eventlog file.
Cleanup now occurs after creating the new file as opposed to before, so updating
the testcase to account for one less logfile after the cleanup.


adizaimi:
Thank you Rivers!

PR #2440, opened 2020-08-26
Fix null/empty parameter binding (7.0)

riverszhang89:
Honor `isnull` if present. Otherwise treat a 0-length parameter as NULL.

(DRQS 162977752)

### DIFF
--- a/plugins/newsql/newsql.c
+++ b/plugins/newsql/newsql.c
@@ -1213,23 +1213,21 @@ static int newsql_param_value(struct sqlclntstate *clnt,
     param->name = val->varname;
     param->pos = val->has_index ? val->index : 0;
     param->type = newsql_to_client_type(val->type);
-
+    int len = val->value.len;
     void *p = val->value.data;
 
-    if (val->has_isnull && val->isnull) {
+    /* The bound parameter is from an old client which does not send isnull,
+       and its length is 0. Treat it as a NULL to keep backward-compatible. */
+    if (len == 0 && !val->has_isnull) {
         param->null = 1;
         return 0;
     }
 
-    if (val->value.data == NULL) {
-        if (param->type != CLIENT_BLOB) {
-            param->null = 1;
-            return 0;
-        }
-        p = (void *)"";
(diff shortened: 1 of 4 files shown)


plugins/newsql/newsql.c
```diff
@@ -1213,23 +1213,21 @@ static int newsql_param_value(struct sqlclntstate *clnt,
...
         return 0;
     }
 
-    if (val->value.data == NULL) {
-        if (param->type != CLIENT_BLOB) {
-            param->null = 1;
-            return 0;
-        }
-        p = (void *)"";
+    if (val->isnull) {
```
nirbhayc, line 1226: Also test for `val->has_isnull` ?

riverszhang89: `val->isnull` being true implies `val->has_is_null` being true too.

riverszhang89:
> Looks good. Also port it to jdbc? Or that's not affected?

Thanks for the review! The pull request is an R7 port of https://github.com/bloomberg/comdb2/pull/2436, which fixes cdb2api as well.

plugins/newsql/newsql.c
riverszhang89: Basically `isnull` is tri-state.

| state | has_isnull | isnull |
| - | - | - |
| not applicable | 0 | 0 |
| NULL | 1 | 1 |
| 0-length | 1 | 0 |

riverszhang89:
The branch passed robomark too: `cdb2test   Aug 26 10:44:43 2020  in-comdb2db isnull7.R20200826.6`

@nirbhayc Thanks again for the review!

plugins/newsql/newsql.c
nirbhayc: I was wondering if it's safe to read 'isnull' when 'has_isnull' is false. Turns out protobuf initializes the field with the default value. So, should be good. Thanks for clarifying.

riverszhang89: Anytime!

PR #2443, opened 2020-08-28
Compatibility mode for fdb requests

mponomar:
Compatibility mode for running fdb queries against older databases - generate an old request type (rqid instead of uuid-based) even if noenv-messages is enabled.

### DIFF
--- a/db/config.c
+++ b/db/config.c
@@ -333,6 +333,7 @@ static char *legacy_options[] = {
     "enable_sql_stmt_caching none",
     "enable_tagged_api",
     "env_messages",
+    "fdb_use_rqid on",
     "init_with_time_based_genids",
     "legacy_schema on",
     "logmsg level info",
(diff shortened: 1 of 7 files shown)


db/config.c
```diff
@@ -368,6 +368,7 @@ static char *legacy_options[] = {
     "init_with_queue_ondisk_header off",
     "init_with_queue_compr off",
     "usenames",
+    "noenv_fdb_messages off"
```
nirbhayc, line 371: It's a sorted list. :)

mponomar: I keep forgetting...

db/sql.h
```diff
@@ -141,6 +141,7 @@ typedef struct osqlstate {
     /* == sql_thread == */
     char *host;              /* matching remote node */
     unsigned long long rqid; /* per node offload request session */
+    unsigned long long rmtqid; /* per node offload request session */
```
dorinhogea, line 144: Why are we adding a new field here? Can we use rqid?

mponomar: When it's uuid-based, we set rqid to OSQL_RQID_USE_UUID that tells the system to use uuids instead.  So I wanted a different id for remote requests for compatibility with R6.

mponomar:
Dorin is taking care of this differently in #2468

PR #2444, opened 2020-08-28
Provide a hook for plugins to specify a different fastseed generator.

mponomar:
Add a hook into comdb2fastseed() that plugins can override.

### DIFF
--- a/db/envstubs.c
+++ b/db/envstubs.c
@@ -56,6 +56,8 @@ static inline void fastseed_set_dup(int dup) { fastseed_dup = dup; }
 
 extern int gbl_mynodeid;
 
+uint64_t (*external_fastseed)(void) = NULL;
+
 /* TODO: fastseed - compatibility mode only - remove */
 uint64_t comdb2fastseed(void)
 {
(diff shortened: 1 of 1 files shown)


db/envstubs.c
```diff
@@ -65,6 +67,9 @@ uint64_t comdb2fastseed(void)
     int seed[2];
     uint64_t out;
 
+    if (external_fastseed)
+        return external_fastseed();
+
     node = gbl_mynodeid;
```
riverszhang89, line 73: My understanding is that `comdb2fastseed` is only needed in the bbenv build. Since you're dropping `fastseed64` into bb-plugins, maybe we could delete the rest of code here?

mponomar: It's also used by local replicants to generate unique record ids, so we're stuck with it for a bit.

riverszhang89: You mean the seqno? I thought it was just genid.

riverszhang89: Yes, you're right. It can be configured to use either fastseed or genid.

PR #2451, opened 2020-08-31
Patch from Richard which addresses the unnecessary btree hits for OP_NoHope

markhannum:
https://www.sqlite.org/src/info/8301da31d08b7467


markhannum:
This will be merged with Joe's changes (coming shortly)

PR #2453, opened 2020-08-31
Delete eventlog files in code rather than using script

adizaimi:
Instead of using system() command and a script to delete old eventlog
files and keep only the last N, use unlink() after sorting and filtering
directory content.

### DIFF
--- a/db/eventlog.c
+++ b/db/eventlog.c
@@ -91,6 +91,10 @@ static inline void free_gbl_eventlog_fname()
     gbl_eventlog_fname = NULL;
 }
 
+static int strptrcmp(const void *p1, const void *p2) {
+      return strcmp(*(char *const *)p1, *(char *const *)p2);
+}
+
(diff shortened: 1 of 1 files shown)


adizaimi:
Thank you Nirbhay!

db/eventlog.c
```diff
@@ -91,40 +91,59 @@ static inline void free_gbl_eventlog_fname()
...
 
-    // fname should look like '/dir/<dbname>.events.' : see eventlog_fname()
-    int len = strlen(fname);
+    int cnt = 100;
+    int num = 0;
+    char **arr = malloc((cnt * sizeof(char *)));
+ 
+    DIR *d = opendir(dname);
+    struct dirent *de;
+    while ( (de = readdir(d)) ) {
```
riverszhang89, line 122: `readdir` -> `bb_readdir`

PR #2454, opened 2020-08-31
Stop loading stats during prepare while holding the schema lock.

mistachkin:
These changes prevent the stat tables from being loaded while the schema lock is held.


akshatsikarwar:
This is wrong - a schemachange can sneak in and drop tables we are about to operate on.

PR #2455, opened 2020-08-31
fix case when first remote table acccessed is sqlite_master

dorinhogea:

There are no fdb cache entries for sqlite_master.  When first remote table accessed is sqlite_master, we end up looking up an entry that normally doesn't exist, and skip sqlite stat addition to sqlite hashes.  When this happens, sqlite code that tries to load stats trigger a nested remote table attach that deadlocks the db.  

Mohit pointed the issue to me, and opened up https://github.com/bloomberg/comdb2/pull/2417 to address a more general issue.


mohitkhullar:
Looks good

mohitkhullar (approved):
looks good

dorinhogea:
Thank you Mohit

PR #2446, opened 2020-08-28
fix ioerror during remsql when dohsql is used

dorinhogea:
Signed-off-by: Dorin Hogea <dhogea@bloomberg.net>

Use ioerror test with dohsql + remsql; fixes for the retry and the locking when shard processing fails due to remote I/O failure.

PR #2457, opened 2020-09-01
Ported https://github.com/bloomberg/comdb2/pull/2446 from master

dorinhogea:
See https://github.com/bloomberg/comdb2/pull/2446


dorinhogea:
Wrong base

PR #2499, opened 2020-09-17
One line change to initialize Mem before usage

aakash10292:
Just a one line addition to Init a Mem structure before using it. 
In the loop before the line of change, we initialize Mem before using it. We seem to have missed it for the genid case

### DIFF
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -831,6 +831,7 @@ static int ondisk_to_sqlite_tz(struct dbtable *db, struct schema *s, void *inp,
         genid = rrn;
 
     if (genid) {
+        memset(&m[fnum], 0, sizeof(Mem));
         m[fnum].u.i = genid;
         m[fnum].flags = MEM_Int;
 


db/sqlglue.c
```diff
@@ -831,6 +831,7 @@ static int ondisk_to_sqlite_tz(struct dbtable *db, struct schema *s, void *inp,
         genid = rrn;
 
     if (genid) {
+        memset(&m[fnum], 0, sizeof(Mem));
```
riverszhang89, line 834: This shouldn't really matter, right? Since the type is `MEM_Int`, we only need `Mem.u.i` in `SerialPut`.

aakash10292: Agreed. Nevertheless, safer to initialize now than expose ourselves to uninitialized memory bugs in the future , when/if things change

adizaimi: I think the intention here is to only access what we mean (u.i when MEM_Int is set) and any other usage is a bug -- in fact memsetting could cover up for such bug. It can also be less performant, altho I can't tell with what frequency this is called.

PR #2506, opened 2020-09-23
stop sending unused query_costs bytes to master

dorinhogea:

Currently we send client_query_costs from replicant to master, but master never reads the bytes.  In addition, client_query_costs are computed when statement is done, after we send the info to the master, so in effect we send old info.  This patch removes sending bytes until we decide to either fix the query_costs, or remove the code altogether.


mponomar (approved):
Thanks Dorin.   Looks good.  I can't recall how (if ever?) this was used.

dorinhogea:
Thank you Mike

PR #2542, opened 2020-10-15
Make sure ECDHE ciphers are enabled

sorindumitru:
Without calling SSL_CTX_set_ecdh_auto when the server uses
an ecdsa certificate only the ECDH_* ecdsa ciphers will be
allowed, even though the ECDHE ciphers are in the list of
acceptable ciphers.

This fixes an issue with connecting to comdb2 from runtimes
that only support ECDHE, not ECDH.

The faction call is only required for openssl versions prior to
1.1.

bug fix, but could also be seen as a feature to support ECDHE ciphers.

When connecting to comdb2 with a TLS library that does not support ECDH_* ciphers, a handshake failure would occur.
One such way is through go's `crypto/tls` package.

I think the easiest way to reproduce this is to generate an ecdsa certificate for comdb2 and set:
```
ssl_cipher_suites ECDHE
```
in its config. This forces the server to only accept ECDHE ciphers, but since they are not enabled all connections will fail.

comdb2 would only accept ECDH_* ciphers, now it acceps both ECDH_* and ECDHE_*.

### DIFF
--- a/util/ssl_support.c
+++ b/util/ssl_support.c
@@ -391,6 +391,11 @@ int SBUF2_FUNC(ssl_new_ctx)(SSL_CTX **pctx, ssl_mode mode, const char *dir,
     if (protocols != 0)
         SSL_CTX_set_options(myctx, protocols);
 
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+    /* Make sure ECDHE ciphers are enabled */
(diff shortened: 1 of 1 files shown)


riverszhang89:
Thank you for the suggestion! It works perfectly. I've made changes that are slightly more complete than this pull request, in https://github.com/bloomberg/comdb2/pull/2545.

PR #2610, opened 2020-12-08
net: Set bufsize for appsocks

akshatsikarwar:
/silent


akshatsikarwar:
I found some performance discrepancy when doing client benchmarks and realized that new-net was not setting buffer size for appsock fds. This makes it identical to old-net way of accepting new socket.

PR #2614, opened 2020-12-09
Minor summarizer changes for theoretically better disk performance

riverszhang89:
Remove unnecessary `fdatasync()` and pass a correct argument to `fadvise()`.

### DIFF
--- a/bdb/summarize.c
+++ b/bdb/summarize.c
@@ -289,6 +289,13 @@ int bdb_summarize_table(bdb_state_type *bdb_state, int ixnum, int comp_pct,
     unsigned long long recs_looked_at = 0;
     int fd = -1;
     int last, now;
+#ifdef POSIX_FADV_SEQUENTIAL
+    /* Release page cache every FADVISE_THRESH many pages. We could make it
+       a tunable, but for now, leave it hardcoded. */
+    off_t nread = 0;
+    const static size_t FADVISE_THRESH = 1024;
+    int usedio = bdb_attr_get(bdb_state->attr, BDB_ATTR_DIRECTIO);
+#endif
 
     if (comp_pct > 100 || comp_pct < 1) {
         *bdberr = BDBERR_BADARGS;
@@ -361,15 +368,20 @@ int bdb_summarize_table(bdb_state_type *bdb_state, int ixnum, int comp_pct,
         goto done;
     }
 
-#if defined(_IBM_SOURCE) || defined(__linux__)
+#ifdef POSIX_FADV_SEQUENTIAL
     // inform kernel that we will be accessing file sequentially
-    // and that we won't need the file to be cached
-    fdatasync(fd);
-    posix_fadvise(fd, 0, 0, POSIX_FADV_DONTNEED | POSIX_FADV_SEQUENTIAL);
+    (void)posix_fadvise(fd, 0, 0, POSIX_FADV_SEQUENTIAL);
 #endif
 
     last = comdb2_time_epoch();
     for (rc = read(fd, page, pgsz); rc == pgsz; rc = read(fd, page, pgsz)) {
+#ifdef POSIX_FADV_SEQUENTIAL
+        /* Periodically hint the OS to release pages we've read. Only do so
+           when directio is enabled for this operation will likely force out
+           useful cached pages otherwise. */
+        if (usedio && ((++nread) % FADVISE_THRESH) == 0)
+            (void)posix_fadvise(fd, (nread - FADVISE_THRESH) * pgsz, FADVISE_THRESH * pgsz, POSIX_FADV_DONTNEED);
+#endif
         /* If it is not a leaf page, continue reading the file. */
         if (!ISLEAF(page))
             continue;


bdb/summarize.c
```diff
@@ -361,11 +361,9 @@ int bdb_summarize_table(bdb_state_type *bdb_state, int ixnum, int comp_pct,
         goto done;
     }
 
-#if defined(_IBM_SOURCE) || defined(__linux__)
+#if defined(POSIX_FADV_SEQUENTIAL)
     // inform kernel that we will be accessing file sequentially
-    // and that we won't need the file to be cached
-    fdatasync(fd);
```
riverszhang89, line 367: Analyze doesn't care if dirty pages in the kernel page cache have been sync'd to the disk device. So no need to fsync here.

bdb/summarize.c
```diff
@@ -361,11 +361,9 @@ int bdb_summarize_table(bdb_state_type *bdb_state, int ixnum, int comp_pct,
...
-    posix_fadvise(fd, 0, 0, POSIX_FADV_DONTNEED | POSIX_FADV_SEQUENTIAL);
```
riverszhang89, line 368: `fadvise()` flags cannot be combined. Plus we probably do not want to use `POSIX_FADV_DONTNEED` for it'll clear pages from the index out of kernel page cache, which may be undesirable.

bdb/summarize.c
adizaimi: I think the fdatasync() was done so advise would be applied; not sure if needed nowadays.
https://insights.oetiker.ch/linux/fadvise/ seems to use it for that purpose too.

bdb/summarize.c
adizaimi: The idea was to have fewest number of pages in kernel cache to be nice to the system--these pages are not needed at all once we have processed them. The fact that we can't combine the flags is indeed unfortunate.

roborivers (approved):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p><a href='http://comdb2.s3-website-us-east-1.amazonaws.com/tests/53f199f1/detail.txt'>Detail</a> | <a href='http://comdb2.s3-website-us-east-1.amazonaws.com/tests/53f199f1/logs.tar.gz'>Log</a> | <a href='http://ec2-34-202-216-70.compute-1.amazonaws.com/restricted/index.php?id=2614'>Resubmit</a></p>

bdb/summarize.c
riverszhang89: It's only *recommended* when using `POSIX_FADV_DONTNEED`. Like I said, there's no reason for analyze to care if data has been sync'd to device. So it's an overkill to fsync here.

bdb/summarize.c
riverszhang89: >> these pages are not needed at all once we have processed them

This isn't entirely true. And this is why `POSIX_FADV_DONTNEED` may not be appropriate here. Pages processed by analyze may be needed by the normal operations such as an index lookup or modification. Forcing these pages out therefore therefore isn't desirable.

riverszhang89: In fact the optimization here makes very little sense with DIRECTIO which bypasses kernel page cache :)

riverszhang89:
I've made changes to 1) hint the OS that we'll be reading index files sequentially; 2) hint the OS to release page cache every 1024 pages. From the end user's point of view, there's very little difference (i.e., same runtime). However it does improve the kernel memory usage.

My benchmark is as below:

```
table tbl sz       7.95GB   0% (dta 2.41GB, ix0 1.42GB, ix1 1.44GB, ix2 2.69GB)
```

Analyze the table with 20% sampling rate.

| Branch | free mem high water mark (`free(1)`)| free pages low water mark  (`free(1)`)| difference |
|-|---------------------------------|---------------------------------|-------------|
|This PR| 247841 MiB | 245856 MiB | -1985 MiB |
|Trunk| 247860 MiB | 241978 MiB | -5882 MiB |

So this pull request requires roughly only 1/3 memory of the trunk.

bdb/summarize.c
riverszhang89: Even with `O_DIRECT`, `fsync()` will likely still have a performance impact here. From `fsync` man page:

> This includes writing through or flushing a disk cache if present.

`O_DIRECT` bypasses kernel page cache, however the backing device may have its own cache. According to the man page, `fsync()` will flush the device cache too.

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>1/388 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>basic_rowlocks_generated [core dumped]</p><p><a href='http://comdb2.s3-website-us-east-1.amazonaws.com/tests/62f9a5d6/detail.txt'>Detail</a> | <a href='http://comdb2.s3-website-us-east-1.amazonaws.com/tests/62f9a5d6/logs.tar.gz'>Log</a> | <a href='http://ec2-34-202-216-70.compute-1.amazonaws.com/restricted/index.php?id=2614'>Resubmit</a></p>

adizaimi:
I don't remember what benchmarks we did at the time of adding current implementation. It is unfortunate that or-ing the options does not work--the intention was good. It is nice that doing two fadvise separately works, and great that now your patch improves things measurably. We could be more aggressive about advising to release pages, but this is probably good enough for now : ). Good catch there about backing off when the db is not running with directio.

bdb/summarize.c
```diff
@@ -361,15 +368,20 @@ int bdb_summarize_table(bdb_state_type *bdb_state, int ixnum, int comp_pct,
...
+    (void)posix_fadvise(fd, 0, 0, POSIX_FADV_SEQUENTIAL);
 #endif
 
     last = comdb2_time_epoch();
-    for (rc = read(fd, page, pgsz); rc == pgsz; rc = read(fd, page, pgsz)) {
+    for (nread = 1, rc = read(fd, page, pgsz); rc == pgsz; ++nread, rc = read(fd, page, pgsz)) {
+#ifdef POSIX_FADV_SEQUENTIAL
+        /* Periodically hint the OS to release pages we've read. Only do so
```
adizaimi, line 379: How about nread++; here and declare (and init to 0) in the POSIX_FADV_SEQUENTIAL block above.

riverszhang89: Yes, done.

adizaimi (approved):
LGTM. Thanks Rivers!

riverszhang89:

PR #2643, opened 2021-01-12
Add autocommit flag to parameters

hgeraldino:
Allow passing the autocommit flag as a connection parameter

  Feature

    Add the ability to pass the autocommit flag as a connection parameter

  Some frameworks (JPA, Spring ORM, Hibernate, etc) work at a Datasource level. Connection lifecycle is handled by the framework itself; users don't have a mechanism to customize Connection objects so properties like `autocommit` can't be changed programmatically. 
  
  Adding a parameter to the JDBC URL allows users to set values for this property at the Datasource level.


riverszhang89 (approved):

riverszhang89:
Thank you for the pull request!

PR #2890, opened 2021-06-24
ssl_new_ctx(): fix openssl version ifdef

c4rlo:
I happened to notice that the OPENSSL_VERSION_NUMBER check seems wrong. As [documented](https://www.openssl.org/docs/man1.1.0/man3/OPENSSL_VERSION_NUMBER.html), it is structured as:

```
 MNNFFPPS: major minor fix patch status
```

And this code should be checking whether openssl is at version 1.0.2 or greater, since that is where `SSL_CTX_set_ecdh_auto()` was introduced (see [changelog](https://www.openssl.org/news/changelog.html)).

Note: this is completely untested.

* What is the type of the change (bug fix, feature, documentation and etc.) ? bug fix
* What are the current behavior and expected behavior, if this is a bugfix ? `SSL_CTX_set_ecdh_auto()` is incorrectly not being called with openssl version 1.0.2, even though it is available, so the fallback code is used instead. Not a big deal, but still wrong.
* What are the steps required to reproduce the bug, if this is a bugfix ? I have not attempted to reproduce anything.
* What is the current behavior and new behavior, if this is a feature change or enhancement ? See above.

### DIFF
--- a/util/ssl_support.c
+++ b/util/ssl_support.c
@@ -380,7 +380,7 @@ int SBUF2_FUNC(ssl_new_ctx)(SSL_CTX **pctx, ssl_mode mode, const char *dir,
 #endif /* HAVE_CRL */
 
 #ifndef OPENSSL_NO_ECDH
-#if OPENSSL_VERSION_NUMBER >= 0x10200000L
+#if OPENSSL_VERSION_NUMBER >= 0x10002000L
     SSL_CTX_set_ecdh_auto(myctx, 1);
 #else
     EC_KEY *ecdh = EC_KEY_new_by_curve_name(NID_X9_62_prime256v1);


util/ssl_support.c
```diff
@@ -380,7 +380,7 @@ int SBUF2_FUNC(ssl_new_ctx)(SSL_CTX **pctx, ssl_mode mode, const char *dir,
 #endif /* HAVE_CRL */
 
 #ifndef OPENSSL_NO_ECDH
-#if OPENSSL_VERSION_NUMBER >= 0x10200000L
+#if OPENSSL_VERSION_NUMBER >= 0x10002000L
```
sorindumitru, line 383: This should actually be:
```suggestion
#if OPENSSL_VERSION_NUMBER < 0x10100000L
```

The [changelog](https://github.com/openssl/openssl/blob/master/CHANGES.md#changes-between-102h-and-110--25-aug-2016) for version 1.1 has this to say about this function:
```
SSL_{CTX_}set_ecdh_auto() has been removed and ECDH is support is always enabled now. If you want to disable the support you should exclude it using the list of supported ciphers. This also means that the "-no_ecdhe" option has been removed from s_server.
```

riverszhang89:
Thank you, Carlo! Yes, the check here is supposed to be 1.0.2.

util/ssl_support.c
riverszhang89: Yes. Your approach is definitely more complete. `SSL_CTX_set_ecdh_auto()` isn't gone in 1.1.0. It's made a no-op ([man page](https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_ecdh_auto.html)). So what Carlo has is probably good enough.

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>1/406 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>comdb2sys_pagesize_generated [setup failure]</p><p><a href='http://comdb2.s3-website-us-east-1.amazonaws.com/tests/c64a5461/detail.txt'>Detail</a> | <a href='http://comdb2.s3-website-us-east-1.amazonaws.com/tests/c64a5461/logs.tar.gz'>Log</a> | <a href='http://ec2-34-202-216-70.compute-1.amazonaws.com/restricted/index.php?id=2890'>Resubmit</a></p>

riverszhang89 (approved):

PR #2894, opened 2021-07-02
Call sql_tick in direct-count

riverszhang89:
We rely on sql_tick() to relinquish locks, send back smartbeats, honor query limits, detect disconnected client, etc. Direct-count does not call sql_tick() on berkdb cursor move, which means that it is not getting any of these useful features. This patch fixes it.

### DIFF
--- a/bdb/cursor.c
+++ b/bdb/cursor.c
@@ -8613,13 +8613,21 @@ struct count_arg {
     DB *db;
     int64_t count;
     int rc;
+    /* the sql_thread struct from parent thread. sql_tick() needs it. */
+    void *sqlthd;
 };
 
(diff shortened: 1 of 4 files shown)


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p>

nirbhayc (approved):
Good catch!

riverszhang89:

PR #2966, opened 2021-09-28
Implicitly promote a String operand to Datetime for Datetime subtraction

riverszhang89:
Attempt to promote a String operand to Datetime if the other operand is a Datetime, and fall back to SQLite's default handling if the conversion fails.


mponomar (approved):

sqlite/src/vdbe.c
```diff
@@ -1935,6 +1935,20 @@ case OP_Remainder: {           /* same as TK_REM, in1, in2, out3 */
     pOut->z = pOut->zMalloc = NULL;
   }
 
+  /* Subtraction only: If one operand is a String and the other is a Datetime,
+     attempt to promote the String operand to Datetime. */
+  if( ( pOp->opcode==OP_Subtract )&&( (pIn1->flags|pIn2->flags)&MEM_TypeMask )==(MEM_Str|MEM_Datetime) ) {
```
nirbhayc, line 1940: Enclose this in a ```#ifdef SQLITE_BUILDING_FOR_COMDB2``` block.

nirbhayc, line 1940: I am wondering if we should check for ```(pIn1->flags|pIn2->flags)&MEM_TypeMask )==MEM_Datetime``` instead? That's is, proceed with the conversions iff at least one of these operands is datetime?

riverszhang89: It's already protected by `SQLITE_BUILDING_FOR_COMDB2`.

nirbhayc: Yes, I see it now. Wasn't visible in the patch.

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>1/411 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>sc_truncate [core dumped]</p>

sqlite/src/vdbe.c
riverszhang89: I assume you mean combinations of a primitive type and a Datetime, as other combinations (e.g., Datetime - Datetime, Datetime + Interval) are well-defined in this routine.

I'm a bit reluctant to convert primitive types here for a couple reasons:

1) Whether or not we convert a number, the result will be lossless in the sense that no information is dropped from either type (By returning an interval, you get a period of time; By returning an integer, you get the duration). In other words, you'll be getting essentially the same content, but of 2 different types.

2) Whether we convert a number to datetime, or convert a datetime to a number, they are both logical - one operand is converted to the type of the other operand, which is in line with most programming languages.

3) And perhaps more importantly, this will change the return type from a number to an Interval, which may break existing applications. That's why I'd err on the side of caution and keep it just for String and Datetime.

nirbhayc: Sure, that makes sense.

By the way, while looking at the code I found that `pIn1` & `pIn2` are actually *right and left* operands (flipped, unlike stated in the comments few lines above). Unrelated to your changes but nonetheless thought of pointing.

nirbhayc (approved):
Okay to merge

riverszhang89:
@mponomar @nirbhayc Thank you for the review!

PR #2986, opened 2021-10-08
Don't let multiple threads update disk space stats

mponomar:
```
#!/bin/bash

for i in {1..10}; do
    yes "select value from comdb2_metrics where name='diskspace'" | cdb2sql -tabs -s mikedb local - | grep -v 'cdb2sql>' &
done
```

10 tables, 1M rows of 1 int column each.

Without patch:
2067310574
2067302382
1994369006
2165090286
2385356782
2115930094

With patch:
1994427563
1994427563
1994427563
1994427563
1994427563
1994427563
1994427563
1994427563


nirbhayc (approved):
Okay to merge.

PR #2992, opened 2021-10-18
block renaming time partitions

dorinhogea:
Renaming time partitions is not supported.  Return proper error if clients accidentally try to rename a partition.


nirbhayc (approved):
Looks good. Should we also block renaming individual shards?

dorinhogea:
You cannot do any schema change on individual shards

PR #3018, opened 2021-11-03
remove always that is not always

dorinhogea:
Creating a user view and dropping a user view (i.e. not a tpt) triggers a trigger ALWAYS check in debug mode that crashes tunables.  Trivial fix.


aakash10292 (approved):

dorinhogea:
Thank you!

PR #3051, opened 2021-12-01
Fix build

riverszhang89:
Fix build failure caused by function redefinitions.


nirbhayc (approved):

riverszhang89:
@nirbhayc Thanks for the review!

PR #3056, opened 2021-12-02
Check for read/write permissions once per table in query

chands10:
Check for read/write permissions only once per table in a query instead of on every cursor move

### DIFF
--- a/db/db_access.c
+++ b/db/db_access.c
@@ -197,6 +197,10 @@ int access_control_check_sql_write(struct BtCursor *pCur,
 
     struct sqlclntstate *clnt = thd->clnt;
 
+    if (pCur->permissions & ACCESS_WRITE) {
+        return 0;
+    }
+
     if (gbl_uses_accesscontrol_tableXnode) {
         rc = bdb_access_tbl_write_by_mach_get(
             pCur->db->dbenv->bdb_env, NULL, pCur->db->tablename,
@@ -252,6 +256,7 @@ int access_control_check_sql_write(struct BtCursor *pCur,
         }
     }
 
+    pCur->permissions |= ACCESS_WRITE;
     return 0;
 }
 
@@ -264,6 +269,9 @@ int access_control_check_sql_read(struct BtCursor *pCur, struct sql_thread *thd)
 
     if (pCur->cursor_class == CURSORCLASS_TEMPTABLE)
         return 0;
+    if (pCur->permissions & ACCESS_READ) {
+        return 0;
+    }
 
     if (gbl_uses_accesscontrol_tableXnode) {
         rc = bdb_access_tbl_read_by_mach_get(
@@ -322,6 +330,7 @@ int access_control_check_sql_read(struct BtCursor *pCur, struct sql_thread *thd)
         }
     }
 
+    pCur->permissions |= ACCESS_READ;
     return 0;
 }
 
--- a/bdb/bdb_api.h
+++ b/bdb/bdb_api.h
@@ -1740,11 +1740,11 @@ int bdb_set_sc_start_lsn(tran_type *tran, const char *table, void *plsn,
 int bdb_delete_sc_start_lsn(tran_type *tran, const char *table, int *bdberr);
 
 enum {
-    ACCESS_INVALID = 0,
-    ACCESS_READ = 1,
-    ACCESS_WRITE = 2,
-    ACCESS_DDL = 3,
-    ACCESS_USERSCHEMA = 4
+    ACCESS_INVALID = 1,
+    ACCESS_READ = 2,
+    ACCESS_WRITE = 4,
+    ACCESS_DDL = 8,
+    ACCESS_USERSCHEMA = 16
 };
 
 int bdb_tbl_access_write_set(bdb_state_type *bdb_state, tran_type *input_trans,
--- a/db/sql.h
+++ b/db/sql.h
@@ -1065,6 +1065,8 @@ struct BtCursor {
     int tableversion;
 
     void *query_preparer_data;
+
+    int permissions; /* permissions for read/write access to table */
 };
 
 struct sql_hist {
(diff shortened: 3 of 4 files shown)


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>2/421 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>sirandtest2<br>timepart_trunc</p>

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p>

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>4/421 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>selectv_recom [setup failure]<br>logfill<br>sc_timepart<br>sc_swapfields_logicalsc_generated</p>

db/db_access.c
```diff
@@ -264,6 +269,9 @@ int access_control_check_sql_read(struct BtCursor *pCur, struct sql_thread *thd)
 
     if (pCur->cursor_class == CURSORCLASS_TEMPTABLE)
         return 0;
```
nirbhayc, line 271: This change is good and addresses the issue. However, I think we should also first check for `clnt->authgen != gbl_bpfunc_auth_gen` and invoke `check_sql_access()` on mismatch and then fallback doing the other checks only if `check_sql_access()` succeeds. `check_sql_access()` would take care of setting `clnt->authgen`.

Additionally, an authgen mismatch also implies that we should clear `pCur->permissions` so that it triggers a fresh read/write auth check.

nirbhayc (changes requested):

db/db_access.c
chands10: So as in check if `clnt->authgen != gbl_bpfunc_auth_gen` at the top of the functions, and if they are not equal, then clear the existing permissions in `pCur->permissions`? If we do this, then we would be able to reauthenticate for the current table, and set `clnt->authgen = gbl_bpfunc_auth_gen` at this point. However, I do not think we would be able to clear the permissions for the rest of the tables since those are in separate `pCur`s. If we didn't set `clnt->authgen = gbl_bpfunc_auth_gen` at this point then we would be able to clear permissions for all of the tables in a query, but would go through the whole procedure for every cursor move again.

riverszhang89: I don't understand why authgen isn't sufficient and a new flag must be introduced. Seems like you can also bump gbl_bpfunc_auth_gen for revoke/grant, and check per-clnt authgen against gbl_bpfunc_auth_gen for any mismatch?

chands10: The authgen's are equal in the beginning of a query at this point of the code so then the authentication will always pass if we just check authgens. 

Also if you switch users, change your password, and then run another query, the authgens will be equal again once this part of the code is reached, even though gbl_bpfunc_authgen would be incremented after changing password (since a different part of the code will then set `clnt->authgen = gbl_bpfunc_authgen`), and we run into the same issue as above.

Also I haven't tested this case but I was also thinking what if we run a query that involves two tables, one which we don't have access to. Assume that the authgens are not equal at the start, so we run through `check_sql_access()`. It authenticates for the first table, and then we set `clnt->authgen = gbl_bpfunc_authgen` in the same function. But then we run `check_sql_access()` for the second table, the one we should not have access to. Then I believe we should have `clnt->authgen == gbl_bpfunc_authgen` and will authenticate access when it should fail.

chands10: And if you have read access to a table but not write access, then select from this table, we will authenticate and set `clnt->authgen = gbl_bpfunc_authgen`. And when we try to insert into this table, we will still have `clnt->authgen == gbl_bpfunc_authgen`

chands10: One option we could do if we still want to check for permissions in the middle of a query is create an authgen variable within `pCur` and compare that against `gbl_bpfunc_authgen`/update it as needed, but I would still need the permissions variable I also created since I have to check whether read/write access has been granted due to the case above. And would have to increment `gbl_bpfunc_auth_gen` on revoke/grant. Not sure if this is something I should pursue.

mponomar: I think it's reasonable not to recheck permissions in the middle of a query.  If they change, the new permissions will apply to the next query.

nirbhayc: > I think it's reasonable not to recheck permissions in the middle of a query. If they change, the new permissions will apply to the next query.

I just wanted to keep the old behavior (unsure if that was deliberate design decision). But, I am ok with the idea of not rechecking in the middle of the query execution; checking at every cursor operation seemed like an overkill in first place.

nirbhayc: > However, I do not think we would be able to clear the permissions for the rest of the tables since those are in separate `pCur`s

Right. However, per last comments, I think the current changes in the PR should be sufficient.

nirbhayc: > bump gbl_bpfunc_auth_gen for revoke/grant

https://github.com/bloomberg/comdb2/pull/3060

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>socksql_master_swings<br>phys_rep</p>

nirbhayc (approved):
Looks good.

PR #3067, opened 2021-12-23
Reduce libevent overhead

riverszhang89:
The patch gets rid of a context switch from appsock_timer to appsock_rd, and reduces the number of event_add/del calls.

### DIFF
--- a/net/net_appsock.h
+++ b/net/net_appsock.h
@@ -43,14 +43,13 @@ extern int gbl_libevent_appsock;
 #define check_appsock_rd_thd() check_thd(appsock_rd_thd)
 #define check_appsock_timer_thd() check_thd(appsock_timer_thd)
 
-#define event_once(a, b, c)                                                    \
-    ({                                                                         \
-        int erc;                                                               \
-        if ((erc = event_base_once(a, -1, EV_TIMEOUT, b, c, NULL)) != 0) {     \
-            logmsg(LOGMSG_ERROR, "%s:%d event_base_once failed\n", __func__,   \
-                   __LINE__);                                                  \
-        }                                                                      \
-        erc;                                                                   \
+#define evtimer_once(a, b, c)                                                                                          \
+    ({                                                                                                                 \
+        int erc;                                                                                                       \
+        if ((erc = event_base_once(a, -1, EV_TIMEOUT, b, c, NULL)) != 0) {                                             \
+            logmsg(LOGMSG_ERROR, "%s:%d event_base_once failed\n", __func__, __LINE__);                                \
+        }                                                                                                              \
+        erc;                                                                                                           \
     })
 
 #endif
--- a/net/sqlwriter.c
+++ b/net/sqlwriter.c
@@ -32,7 +32,7 @@
 #define MSEC(x) (x * 1000) /* millisecond -> microsecond */
 
 //heartbeat will tick every
-static struct timeval heartbeat_time = {.tv_sec = 0, .tv_usec = MSEC(10)};
+static struct timeval heartbeat_time = {.tv_sec = 0, .tv_usec = MSEC(100)};
 
 //send heartbeat if no data every (seconds)
 #define min_hb_time 1
@@ -66,21 +66,10 @@ struct sqlwriter {
     unsigned wr_continue : 1;
 };
 
-static void sql_enable_trickle(struct sqlwriter *writer)
-{
-    event_del(writer->heartbeat_ev);
-    event_add(writer->heartbeat_trickle_ev, NULL);
-}
-
-static void sql_disable_trickle(struct sqlwriter *writer)
-{
-    event_add(writer->heartbeat_ev, &heartbeat_time);
-    event_del(writer->heartbeat_trickle_ev);
-}
+static void sql_trickle_cb(int fd, short what, void *arg);
 
 static void sql_enable_flush(struct sqlwriter *writer)
 {
-    sql_disable_heartbeat(writer);
     struct timeval recover_ddlk_timeout = {.tv_sec = 1};
     event_add(writer->flush_ev, &recover_ddlk_timeout);
 }
@@ -102,15 +91,14 @@ static void sql_timeout_cb(int fd, short what, void *arg)
         writer->do_timeout = 0;
         writer->timed_out = 1;
     }
-    event_del(writer->timeout_ev);
-    sql_enable_trickle(writer);
     Pthread_mutex_unlock(&writer->wr_lock);
 }
+
+static void sql_heartbeat_cb(int fd, short what, void *arg);
 
 void sql_enable_heartbeat(struct sqlwriter *writer)
 {
     event_add(writer->heartbeat_ev, &heartbeat_time);
-    event_del(writer->heartbeat_trickle_ev);
 }
 
 void sql_disable_heartbeat(struct sqlwriter *writer)
@@ -122,7 +110,7 @@ void sql_disable_heartbeat(struct sqlwriter *writer)
 void sql_enable_timeout(struct sqlwriter *writer, int timeout_sec)
 {
     if (!writer->timeout_ev) {
-        writer->timeout_ev = event_new(appsock_timer_base, -1, EV_TIMEOUT, sql_timeout_cb, writer);
+        writer->timeout_ev = event_new(appsock_timer_base, -1, 0, sql_timeout_cb, writer);
     }
     struct timeval timeout = {.tv_sec = timeout_sec};
     event_add(writer->timeout_ev, &timeout);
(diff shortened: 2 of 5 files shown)


roborivers (commented):
Lint check:

```diff


@@ db/sqlinterfaces.c @@
+ Please make sure that a mutex is unlocked before returning.
```

roborivers (approved):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>socksql_master_swings</p>

roborivers (commented):
Lint check:

```diff


@@ db/sqlinterfaces.c @@
+ Please make sure that a mutex is unlocked before returning.
```

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>1/421 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>reorder_inserts_noreorder_generated</p>

akshatsikarwar:
This is looking good...

net/sqlwriter.c
```diff
@@ -418,14 +399,9 @@ struct sqlwriter *sqlwriter_new(struct sqlwriter_arg *arg)
...
 
-    int flush_flags = EV_WRITE | EV_TIMEOUT | EV_PERSIST;
-    writer->flush_ev = event_new(writer->wr_base, arg->fd, flush_flags, sql_flush_cb, writer);
-
-    int hb_flags = EV_TIMEOUT | EV_PERSIST;
-    writer->heartbeat_ev = event_new(appsock_timer_base, -1, hb_flags, sql_heartbeat_cb, writer);
-
-    int trickle_flags = EV_WRITE | EV_PERSIST;
-    writer->heartbeat_trickle_ev = event_new(appsock_timer_base, arg->fd, trickle_flags, sql_trickle_cb, writer);
+    writer->flush_ev = event_new(writer->wr_base, arg->fd, EV_WRITE, sql_flush_cb, writer);
```
akshatsikarwar, line 402: Wouldn't it be better to keep this persistent - so we don't have to `event_add` in a loop in `sql_flush_int`

plugins/newsql/newsql_evbuffer.c
```diff
@@ -153,7 +153,7 @@ static int newsql_done_cb(struct sqlclntstate *clnt)
         appdata->query = NULL;
     }
     if (sql_done(appdata->writer) == 0) {
-        event_once(appsock_timer_base, newsql_read_again, appdata);
+        newsql_read_again(-1, 0, appdata);
```
akshatsikarwar, line 156: Should this (disable heartbeat/timeout) be push down into `sql_done` which will performs it under lock so it doesn't race with heartbeat?

net/sqlwriter.c
riverszhang89: The flush event doesn't gets fired repeatedly. Generally it triggers only on the last row. So it makes sense to make it not persistent such that it'll be automatically removed from queue once triggered.  If the event is PERSIST, we'll have to call event_del to remove it from the event base, which is a lot more expensive.

You're right that we may have to `event_add` in the while loop if the event isn't PERSIST. However the chances of  `writer->bad` are relatively low, so in most cases we only `event_add` once.

akshatsikarwar:
I guess we also need corresponding change in `fastsql`

plugins/newsql/newsql_evbuffer.c
riverszhang89: `event_del` would block if the callback is being executed in the event base. Therefore it's safe to call `newsql_read_again()` here without holding `wr_lock` or scheduling the function via `appsock_timer_base`.

net/sqlwriter.c
akshatsikarwar: _However the chances of writer->bad_
I wasn't thinking about `writer->bad`. Won't this require scheduling flush a few times to write the entire buffered response. I was thinking one (event_add + event_del) vs (several event_add). What does a table scan look like? Or even a typical request (I don't know what that is.. 10k rows or 1K each?) Put a counter and test?

plugins/newsql/newsql_evbuffer.c
akshatsikarwar: 👍

roborivers (commented):
Lint check:

```diff


@@ db/sqlinterfaces.c @@
+ Please make sure that a mutex is unlocked before returning.
```

net/sqlwriter.c
riverszhang89: We're doing [a while loop](https://github.com/bloomberg/comdb2/blob/master/net/sqlwriter.c#L156) in `sql_flush_cb()` to write out the evbuffer. So unless there's a slow reader, the `flush_ev` event gets scheduled once per flush (Though `evbuffer_write()` may be called multiple times during each run).

However I'm removing this while loop to make this logic more non-blocking-like: if insufficient bytes are written, exit  `sql_flush_cb()`; the callback will be picked up again on next EV_READ event. I've also changed flush_ev back to EV_PERSIST. My benchmark shows that one (event_add + event_del) and several event_add's here perform almost identically.

net/net_appsock.h
```diff
@@ -43,7 +43,7 @@ extern int gbl_libevent_appsock;
 #define check_appsock_rd_thd() check_thd(appsock_rd_thd)
 #define check_appsock_timer_thd() check_thd(appsock_timer_thd)
 
-#define event_once(a, b, c)                                                    \
+#define evtimer_once(a, b, c)                                                    \
```
riverszhang89, line 46: I renamed event_once to evtimer_once as libevent has its own [event_once](http://www.wangafu.net/~nickm/libevent-2.0/doxygen/html/event__compat_8h.html#a074cf31709035ec2b7b660374bbd3b90).

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>1/421 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>upsert</p>

roborivers (commented):
Lint check:

```diff


@@ db/sqlinterfaces.c @@
+ Please make sure that a mutex is unlocked before returning.
```

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>2/421 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>phys_rep<br>upsert</p>

net/sqlwriter.c
akshatsikarwar: I don't think that while loop is sufficient. That only fills up sending tcp buffer and more than that amount can be easily outstanding - even if reader is fast.

akshatsikarwar: The while loop exists to paper over limited number of iovecs that writev will process. So with lots of small rows, tcp buffer is not full, but one writev doesn't pick up all the iovecs. At least, that was what I observed when I added that loop. I don't think removing it achieves anything - Idea was, we are flushing in our own dedicated thread, why go through another epoll when tcp buff isn't full.

akshatsikarwar: Which is also why trickle tries to make the outstanding data contiguous.. But I'm less sure about how useful this is.

riverszhang89: Thank you! I wasn't aware of the writev limit. I put the while loop back.

roborivers (commented):
Lint check:

```diff


@@ db/sqlinterfaces.c @@
+ Please make sure that a mutex is unlocked before returning.
```

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p>

akshatsikarwar (approved):

riverszhang89:
@akshatsikarwar Thank you for the review!

PR #3107, opened 2022-01-27
fix non-blkseq schema change removal of old files

dorinhogea:
C&P error here probably.  We need to create a child txn so that we remove the old files.  If there is no blkseq, parent_trans is NULL (and we do not want to create an independent txn here!).
Instead, we need to pass the "real" parent transaction here, which is iq->sc_logical_tran->physical_tran == trans.


markhannum (approved):
Obviously correct- parent trans has to be NULL in this branch.

dorinhogea:
Thank you Mark!

mponomar (approved):

dorinhogea:
Thank you Mike!

PR #3114, opened 2022-02-03
when we use a virtual table/system table, get its lock

dorinhogea:
Master build does not get a table lock when using virtual tables (i.e. system tables).  7.0 had a fix for one use case in this commit 13adf2153, but that only addresses the simple "select from virtual table" case.  This patch makes sure we have a table lock for any type of statement that uses a virtual table.


dorinhogea:
Apparently master created an extra table lock array in vdbe, here: vTableLocks.  This PR does the same thing with a single line change, but I assume we wanna keep this for now. W/ing.

PR #3119, opened 2022-02-09
Allow a cheapstack for certain lockerids when they are freed

markhannum:
We're seeing an issue where our code aborts when attempting to get a lock from a lockerid which has been incorrectly freed.  The lockerid in each case has always been dbc->lid: a lockerid which is created when the cursor is allocated.  The code marks these cursors with the DB_LOCKER_TRACK_FREE, and will dump a cheapstack when that lockerid is freed on a tunable.  I'm hoping this gets us more information about how these lockerids are disappearing.


roborivers (commented):
Lint check:

```diff


@@ db/db_tunables.h @@
+ Please update the tunable documentation (docs/pages/config/config_files.md) accordingly.
```

PR #3125, opened 2022-02-10
use proper enum for alteronly everywhere

dorinhogea:
Simple, per title.


dorinhogea:
Thank you Mike!

PR #3130, opened 2022-02-15
Don't drain the cursor when closing the ResultSet or Connection 

hgeraldino:
* What is the type of the change (bug fix, feature, documentation and etc.) ? Feature flag
When executing a Stored Procedure that reads data from a Lua Consumer queue, events are removed from the queue if/when the ResultSet is closed even in cases where the application has not read them, which causes data loss. This flag will allow us to control if this 'draining' will happen when closing the ResultSet.

### DIFF
--- a/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Connection.java
+++ b/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Connection.java
@@ -19,8 +19,6 @@ import java.util.*;
 import java.util.concurrent.*;
 import java.util.logging.*;
 
-import com.bloomberg.comdb2.jdbc.Constants.*;
-
 /**
  * @author Rivers Zhang
  * @author Mohit Khullar
@@ -273,6 +271,14 @@ public class Comdb2Connection implements Connection {
     public void setStackAtOpen(boolean sendStack) {
         hndl.hasSendStack = true;
         hndl.setSendStack(sendStack);
+    }
+
+    public void setSkipResultSetDrain(boolean skipDrain) {
+        hndl.setSkipResultSetDrain(skipDrain);
+    }
+
+    public void setClearAck(boolean clearAck) {
+        hndl.setClearAck(clearAck);
     }
 
     public ArrayList<String> getDbHosts() throws NoDbHostFoundException{
--- a/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
+++ b/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
@@ -96,6 +96,8 @@ public class Comdb2Handle extends AbstractConnection {
     private List<String> sets;
 
     private boolean ack = false;
+    private boolean skipDrain = false;
+    private boolean clearAckOnClose = true;
 
     private boolean isRead;
     private String lastSql;
@@ -382,6 +384,14 @@ public class Comdb2Handle extends AbstractConnection {
         }
 
         stmteffects = val;
+    }
+
+    public void setSkipResultSetDrain(boolean val){
+        this.skipDrain = val;
+    }
+
+    public void setClearAck(boolean val){
+        this.clearAckOnClose = val;
     }
 
     public ArrayList<String> getDbHosts() throws NoDbHostFoundException{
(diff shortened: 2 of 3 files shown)


cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
```diff
@@ -2116,7 +2130,9 @@ public void close() throws IOException {
         if (opened == true) {
             synchronized (lock) {
                 if (opened == true) {
-                    if (ack) ack();
+                    if (ack && !skipDrain) {
```
riverszhang89, line 2133: This should be placed under its own tunable. Consider using "[clear_ack](https://github.com/bloomberg/comdb2/blob/master/cdb2api/cdb2api.h#L251)" to be in line with the C api.

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
```diff
@@ -13,20 +13,29 @@
    limitations under the License. */
 package com.bloomberg.comdb2.jdbc;
 
-import java.nio.ByteBuffer;
-import java.io.*;
-import java.util.*;
-import java.util.logging.*;
-import javax.net.ssl.*;
+import com.bloomberg.comdb2.jdbc.Cdb2Query.Cdb2BindValue;
```
riverszhang89, line 16: I would very much prefer to import a whole package instead of individual classes in the package. It's cleaner to read and much easier for non-IDE users to modify the code.

riverszhang89 (changes requested):

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
hgeraldino: Do you mean not passing it as an argument on the JDBC URL, but set it as a tunable on the database itself?

That change will be broader than what I intended, as it will affect all connections, not just those that need this particular behavior.

riverszhang89: skip_rs_drain controls whether unread rows are drained on close; clear_ack controls whether events are consumed on close. Two different arguments.

hgeraldino: Added a new `clear_ack` flag as suggested.

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Connection.java
```diff
@@ -275,6 +273,10 @@ public void setStackAtOpen(boolean sendStack) {
         hndl.setSendStack(sendStack);
     }
 
+    public void setSkipResultSetDrain(boolean skipDrain) {
+        hndl.setSkipResultSetDrain(skipDrain);
+    }
+
```
riverszhang89, line 279: I think you need a setter for `clearAck` here, too.

hgeraldino: yep, it's there

hgeraldino:
Reverted the `import` changes as suggested

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Connection.java
riverszhang89: I don't see it in this patch. You need a setter in the Connection class.

hgeraldino: You're right. Method has been added now

riverszhang89 (dismissed):

riverszhang89:
Looks good to merge. Could you please squash the commits into one?

hgeraldino:
> Looks good to merge. Could you please squash the commits into one?

After fetching changes from upstream, there are several commits in between so squashing is tricky. I can open a fresh new PR from HEAD, or you can squash the commits when merging the PR to master.

Let me now how you'd like to proceed.

riverszhang89:
@hgeraldino You could rebase or open a new PR. Whichever you prefer.

hgeraldino:
> @hgeraldino You could rebase or open a new PR. Whichever you prefer.

Rebased  branch & squashed the two commits. Should be mergeable now

riverszhang89 (approved):

PR #3133, opened 2022-02-15
Support providing connection parameters via properties 

hgeraldino:
Support providing connection parameters via properties (in addition to jdbc url params)

Signed-off-by: Hector Geraldino <hgeraldino@gmail.com>

* What is the type of the change (bug fix, feature, documentation and etc.) ? feature
JDBC allows passing connection parameters via DataSource properties. The properties are partially ignored by the current Driver implementation (are only used for addons), so this patch is to add support for configuring the connection using that mechanism. 
It's better to configure the connection via properties rather than adding flags to the jdbc url


hgeraldino:
> Would you please undo the import statements? Looks good to merge otherwise.

Done

riverszhang89 (approved):

PR #3213, opened 2022-03-30
use selectv for id query

zhengcao09:
* Bug fix. 
* We have a use case where the id for a table is generated using another table. The dialect should generate a 'select for update' to lock the table/row for the current TX. However the generated sql does not do that. comdb2 has a selectv (v at the end) which simulates the 'select for update' semantics but that is not generated.


aibolit (approved):

riverszhang89 (approved):

riverszhang89:
Thank you!

PR #3285, opened 2022-06-03
Drop index sqlite syntax support for time partition tables

chands10:
Signed-off-by: Salil Chandra <schandra107@bloomberg.net>

When debugging create index for tpts I noticed that comdb2CreateIndex was being called first on the tpt as a whole, and then eventually again for each shard (not recursively). And on each shard it would just go into this conditional and return
https://github.com/bloomberg/comdb2/blob/master/sqlite/src/comdb2build.c#L5651

But with my change for drop index support it looks like comdb2DropIndex is only called once (just on the tpt as a whole and not on each shard), and there is no similar conditional in this function. Still, calling select * from sqlite_master after this, I can see that all shards have dropped the index.

Even though there is different behavior between these two functions, similar to create index for tpts, I think that with my drop index change all sc-s for the tpt should share the same transaction to finalize the drop. Can this be confirmed?


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>5/434 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>socksql_master_swings<br>sc_truncate<br>reorder_inserts_noreorder_generated<br>ddl_no_csc2<br>sc_downgrade</p>

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>3/434 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>sc_downgrade [core dumped]<br>socksql_master_swings<br>phys_rep<br>sc_downgrade</p>

chands10:
Thank you!

PR #3295, opened 2022-06-08
config: Enable queuedb_genid_filename by default

akshatsikarwar:
/silent


akshatsikarwar:
This option is default in roborivers build - so failures above are unrelated to this change. Also passed robomark:
`cdb2test   Jun  8 10:39:11 2022  in-comdb2db queuedb_genid_filename-7.R20220608.2`

PR #3301, opened 2022-06-13
Fix partial reads/writes and slow writes

mohitkhullar:
Signed-off-by: mohitkhullar <mkhullar1@bloomberg.net>


mohitkhullar:
If one of the reader/writer thread closes, or there is a heartbeat timeout, we shutdown the fd, which can lead to partial reads/writers and we don't check them in some of the cases. In one case we can process a partial read.

akshatsikarwar (approved):

mohitkhullar:
my default changed away from squash and merge :(

PR #3339, opened 2022-07-19
Fix chunk transaction crash

riverszhang89:
A chunk transaction has an unwanted side effect of cleaning up expressional index keys on micro-transaction commit. Replicant may then send junk OSQL_INSIDX/OSQL_DELIDX schedules to the master and eventually crash in the temptable code.

The patch fixes it by latching the expressional index keys before spawning a micro-transaction and restoring them back onto clnt after the micro-transaction commits.

(DRQS 169860714)


mponomar (approved):

nirbhayc (approved):
LGTM

PR #3362, opened 2022-08-03
Disable the auto-consume logic in cdb2_close

### LINKED ISSUE #3361: Potential data loss with lua triggers
**Describe the bug**

When closing the connection in the middle of consuming items from a lua trigger, we would expect for the item to be acknowledged, and a subsequent re-opening of the connection to be able to resume to the next item in the queue, but instead it seems that the next items on the queue are not re-delivered.

**To Reproduce**
1. Compile cdb2api and install it
2. Create a mattdb local instance and start it
3. Add the config to connect to it (e.g. `echo "$COMDB2_DBNAME 1234 $(hostname -f)" > /opt/bb/etc/cdb2/config/comdb2db.cfg`)
4. Compile https://gist.github.com/gusmonod/0c43bc81e37a024d7139c8175c07daa5
5. Run it
6. See that it is hanging when trying to consume event of id 3
7. Insert into the `monitored` table (e.g. `/opt/bb/bin/cdb2sql mattdb dev "insert into monitored(id, message) values (5, 'five')"`)
8. Notice that the event consumed is not of id 5 and not of id 3

**Expected behavior**

We should be able to consume the event showing `3, NULL, bye` and not hang.

**Screenshots**

demo

**Environment (please complete the following information):**
 - Operating System and Version: Ubuntu Linux 20.04 running on Github Actions (so in azure)
 - Comdb2 version: https://github.com/gusmonod/comdb2/tree/hostname_length

**Additional context**

To see how we are creating the environment, compiling comdb2, creating and starting the local instance see: https://github.com/bloomberg/python-comdb2/blob/bf9f758ef7eb291fe5c0072fa0fb0e39bc5e421c/.github/workflows/build.yml

riverszhang89:
cdb2_close() attempts to drain unfetched rows in order to make the connection reusable for the next sockpool user. This logic however may undesirably consume events from a consumer. Disable it by default.

Fixes #3361.

### DIFF
--- a/cdb2api/cdb2api.c
+++ b/cdb2api/cdb2api.c
@@ -94,7 +94,7 @@ static int MIN_RETRIES = MIN_RETRIES_DEFAULT;
 #define CDB2_CONNECT_TIMEOUT_DEFAULT 100
 static int CDB2_CONNECT_TIMEOUT = CDB2_CONNECT_TIMEOUT_DEFAULT;
 
-#define CDB2_AUTO_CONSUME_TIMEOUT_MS_DEFAULT 2
+#define CDB2_AUTO_CONSUME_TIMEOUT_MS_DEFAULT 0
 static int CDB2_AUTO_CONSUME_TIMEOUT_MS = CDB2_AUTO_CONSUME_TIMEOUT_MS_DEFAULT;
 
 #define COMDB2DB_TIMEOUT_DEFAULT 2000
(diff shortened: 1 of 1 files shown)


roborivers (commented):
Lint check:

```diff


@@ cdb2api/cdb2api.c @@
+ Please port the change to cdb2jdbc if necessary.
+ If you add any internal configuration state to this file, please update the reset_the_configuration() function as well to include it.
```

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>9/440 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>sc_downgrade [core dumped]<br>socksql_master_swings<br>sc_truncate<br>phys_rep<br>ignore_bad_table<br>weighted_standing_queue<br>load_cache_dumpmax_generated<br>load_cache_autocache_generated<br>cdb2_close<br>sc_downgrade</p>

akshatsikarwar:
I was thinking it might be better to simply track if `cdb2_run_statment` has ever received a ping-pong request, and if so skip donating to sockpool (and so no need to drain connection). Don't want to disable it by default.

akshatsikarwar:
Also need to patch `consume_previous_query` in the API.

akshatsikarwar:
bb-cdb2api does not have auto-consume in `cdb2_close`, but does in `cdb2_run_statement`:
```
   else if (hndl->ack) {
       hndl->ack = 0;
       if (hndl->sb) {
           int fd = cdb2buf_fileno(hndl->sb);
           shutdown(fd, SHUT_RDWR);
       }
   }
   else {
       while(cdb2_next_record_int(hndl) == CDB2_OK);
   }
```

This is insufficient, as user might have called `cdb2_clear_ack`. I think we need to track if prior `cdb2_run_statement` was for a consumer or not.

riverszhang89:
Yes you're right. We need to patch these places for a mixed use of 'exec procedure' on a consumer and regular queries.

akshatsikarwar:
Disable for now - we can make it robust later.

akshatsikarwar (approved):

PR #3384, opened 2022-08-17
rep_mon thread can pstack a thread blocked in rep_process_message

markhannum:
Signed-off-by: Mark Hannum <mhannum72@gmail.com>

This is the master version of https://github.com/bloomberg/comdb2/pull/3383

### DIFF
--- a/bdb/rep.c
+++ b/bdb/rep.c
@@ -47,6 +47,9 @@
 #include <compat.h>
 #include "str0.h"
 #include <thrman.h>
+#ifdef _LINUX_SOURCE
+#include <sys/syscall.h>
+#endif
 
 #define REP_PRI 100     /* we are all equal in the eyes of god */
 #define REPTIME 3000000 /* default 3 second timeout on election */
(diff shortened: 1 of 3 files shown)


roborivers (commented):
Lint check:

```diff


@@ db/db_tunables.h @@
+ Please update the tunable documentation (docs/pages/config/config_files.md) accordingly.
@@ bdb/rep.c @@
+ Please remove debugging statements, and replace fprintf/printf with logmsg.
@@ bdb/rep.c @@
+ Please make sure that a mutex is unlocked before returning.
@@ bdb/rep.c @@
+ Please replace any pthread_xxx() function in your diff with corresponding Pthread_xxx() wrapper.
```

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>0/0 tests failed</b> &#x26A0;.</p>

bdb/rep.c
```diff
@@ -3729,6 +3732,85 @@ void bdb_set_seqnum(void *in_bdb_state)
...
+    }
+    hash_free(rep_mon_hash);
+    rep_mon_hash = NULL;
+    logmsg(LOGMSG_INFO, "Exiting replication monitor thread\n");
+    pthread_mutex_unlock(&rep_mon_lk);
+    return NULL;
+}
+
+static void add_rep_mon(struct rep_mon *rm)
+{
```
nirbhayc, line 3794: Wondering if we should NOOP this and `rem_rep_mon` functions for the default case `gbl_rep_mon_threshold == 0`, and then we could change the below `else if` into just an `else`.

bdb/rep.c
```diff
@@ -3729,6 +3732,85 @@ void bdb_set_seqnum(void *in_bdb_state)
...
+#endif
+    }
+    return 0;
+}
+
+static void *rep_mon(void *thd)
+{
+    pthread_mutex_lock(&rep_mon_lk);
+    if (rep_mon_hash) {
+        return NULL;
```
nirbhayc, line 3771: I believe you forgot to unlock `rep_mon_lk` here. However, I think `rep_mon_hash` will never be NULL here, so it's better to replace this with an assert.

bdb/rep.c
```diff
@@ -3729,6 +3732,85 @@ void bdb_set_seqnum(void *in_bdb_state)
...
+    logmsg(LOGMSG_INFO, "Exiting replication monitor thread\n");
+    pthread_mutex_unlock(&rep_mon_lk);
+    return NULL;
+}
+
+static void add_rep_mon(struct rep_mon *rm)
+{
+    pthread_mutex_lock(&rep_mon_lk);
+    if (rep_mon_hash) {
+        hash_add(rep_mon_hash, rm);
```
nirbhayc, line 3797: Signal here so that the rep_mon thread dumps the pstack immediately?

nirbhayc (changes requested):

bdb/rep.c
markhannum: Good catch!  The hash toggles to NULL if someone disables the tunable

nirbhayc: Ah ok, I thought it was not a dynamic tunable.

bdb/rep.c
markhannum: We shouldn't be stacking anything until it's been on the queue for at least 1-second anyway ..

nirbhayc: Which is ok as it's supposedly hung? Cool.

bdb/rep.c
markhannum: I think we have to use only the hash actually- imagine that the tunable is disabled after something is added to the hash .. because it's kept on the stack, the rep_mon thread might be looking at free'd stack memory if rem_rep_mon doesn't remove it ..

markhannum: Maybe we can use gbl_rep_mon_threshold for add_rep_mon ..

markhannum: Okay .. I've introduced an 'added' field to struct rep_mon .. if it is added to the rep_mon hash, this will be set to 1.  rem_rep_mon gets the mutex only if added is set to 1.

nirbhayc (approved):
okay to merge

PR #3401, opened 2022-09-02
Enable fail_dispatch on new clients by default

mohitkhullar:
We enable it only after read client's configuration, however for new clients the queries can be rejected without giving an error back, as we haven't read client's configuration yet, leading to >10 second delay. Since client sets it after reading server's shared mem flags, have it turned on by default from server side.

Signed-off-by: mohitkhullar <mkhullar1@bloomberg.net>

### DIFF
--- a/db/sqlinterfaces.c
+++ b/db/sqlinterfaces.c
@@ -5174,6 +5174,7 @@ void reset_clnt(struct sqlclntstate *clnt, int initial)
     }
     clnt->recno = 1;
     clnt->done = 1;
+    clnt->fail_dispatch = 1;
     strcpy(clnt->tzname, "America/New_York");
     clnt->dtprec = gbl_datetime_precision;
     bzero(&clnt->conninfo, sizeof(clnt->conninfo));


mohitkhullar:
With inline_poll enabled, the socket will be rejected just after 'fastsql' and subsequent query request won't be able to write. Without inline_poll, the socket will be rejected after query is sent, and after 5 second read timeout for reply, we might again hang during flush (at close).

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>3/440 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>sc_truncate<br>ignore_bad_table<br>reorder_inserts_noreorder_generated</p>

mohitkhullar:
We might not see any log on server side, because its seen only when we queue the requests, which is the next step after server sends reject request on failed dispatch.

PR #3487, opened 2022-10-25
Fix race in akq that could prevent node from exiting

riverszhang89:
The following events would cause akq_worker to cond_wait indefinitely:

```

     THD-1 (akq_worker)        | THD-2 (akq_enqueue, eg "select 1")   | THD-3 (akq_stop, eg "send exit")
--------------------------------------------------------------------------------------------------------
T0 | cond_wait for work        |                                      |
--------------------------------------------------------------------------------------------------------
T1 |                           | cond_signal akq_worker to do work    |
--------------------------------------------------------------------------------------------------------
T2 | wake up, not being told   |                                      |
   | to stop, so do work       |                                      |
--------------------------------------------------------------------------------------------------------
T3 |                           |                                      | cond_signal akq_worker to stop;
   |                           |                                      | the signal isn't delivered as
   |                           |                                      | akq_worker isn't waiting
--------------------------------------------------------------------------------------------------------
T4 | cond_wait for work        |                                      |
--------------------------------------------------------------------------------------------------------

```

The fix is really simple: before waiting for work, akq_worker checks whether it needs to stop (so in the table above akq_worker would exit at T4).


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>4/448 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>phys_rep<br>ignore_bad_table<br>reorder_inserts_noreorder_generated<br>selectv_rcode_disable_svonly_nop_generated</p>

akshatsikarwar:
I think the check for `stop_work` should be just once, before we call `cond_wait`?

akshatsikarwar:
nitpicking - but we don't need the superfluous comments -- 
`if (q->stop_work) { /* Check if we're supposed to exit */`
stop work, means, stop work :) pretty self explanatory..

akshatsikarwar (approved):

riverszhang89:
> ~I think the check for `stop_work` should be just once, before we call `cond_wait`?~ I think it does make sense to check before and after wait so we can exit sooner..

Sorry I missed this. This was what I had in mind too. I've changed it back and added a comment so we know why we're checking it twice.

akshatsikarwar (approved):
LGTM

PR #3524, opened 2022-11-09
Print stack for string-ref if its leaked

markhannum:
Signed-off-by: Mark Hannum <mhannum72@gmail.com>

cheapstack_sym_char_array prints stack-trace to memory.  STACK_STRING_REFERENCES uses this to print stack-trace of allocating thread if a stringref is leaked.


riverszhang89 (approved):
LGTM!

PR #3551, opened 2022-11-21
Allow auto analyze on IAM enabled databases

mohitkhullar:
Will need only for connect, for rest in_sqlite_init/no_transaction (7.0) takes care of it
{DRQS 170823925}
Signed-off-by: mohitkhullar <mkhullar1@bloomberg.net>

PR #3569, opened 2022-12-05
Bypass IAM auth check for analyze

riverszhang89:
The `access_control_check_sql_read()` function is invoked on each cursor move, in case that authentication changes in the middle of the query. Therefore the same logic in #3551 should probably be applied to the function, as well.


mohitkhullar:
Thanks, we need to add a check before we reset app_data/get_authdata since that isn't being set to NULL

riverszhang89:
@mohitkhullar Thanks for the review!

PR #3374, opened 2022-08-08
Improving libevent memory use

riverszhang89:
Libevent-based network can easily run out of memory on a congested network (rebuilding a bloby table, for instance).
This patch fixes it.

(DRQS 169989112)

PR #2608, opened 2020-12-08
Enable per-thread malloc for libevent net

riverszhang89:
Also rename THREAD_TYPE to ENABLE_PER_THREAD_MALLOC to better explain what the macro really does.

PR #3604, opened 2023-01-03
Reducing libevent memory footprint

riverszhang89:
Libevent pre-allocates at least 1KiB memory. On a congested network the pre-allocation can be a bit too aggressive and may lead to a fragmented heap. This patch addresses it.

(DRQS 171049568)

### DIFF
--- a/net/net_evbuffer.c
+++ b/net/net_evbuffer.c
@@ -1697,38 +1697,43 @@ static void get_stat_evbuffer(struct event_info *e, net_queue_stat_t *stat)
     }
 }
 
+
+static void free_wrapper(const void *unused0, size_t unused1, void *ptr)
+{
+    free(ptr);
+}
+
 static ssize_t readv_evbuffer(struct evbuffer *buf, int fd)
 {
-#   define NVEC 8
-    struct iovec v[NVEC];
-#   ifdef FIONREAD
+    void *p;
     int avail;
+#   ifdef FIONREAD
     (void)ioctl(fd, FIONREAD, &avail);
     if (avail <= 0) avail = TCP_BUFSZ;
 #   else
 #   pragma message("FIONREAD not available")
     avail = TCP_BUFSZ;
 #   endif
-    const int nv = evbuffer_reserve_space(buf, avail, v, NVEC);
-    if (nv <= 0) {
+
+    /* Request the memory directly from the system -
+       Libevent internally pre-allocates at least 1KiB memory. Network data is
+       transferred from rd_buf to akq_buf, on message boundary, via evbuffer_remove_buffer().
+       However, evbuffer_remove_buffer() will copy an entire chunk of memory to avoid memcpy,
+       whenever possible. This can lead to a fragmented heap on a congested network. */
+    if ((p = malloc(avail)) == NULL) {
         errno = ENOMEM;
         return -1;
     }
-    const ssize_t sz = readv(fd, v, nv);
-    if (sz <= 0) {
-        if (errno == EAGAIN || errno == EWOULDBLOCK) return 0;
-        return sz;
-    }
-    ssize_t n = sz;
-    for (int i = 0; i < nv; ++i) {
-        if (v[i].iov_len < n) {
-            n -= v[i].iov_len;
-            continue;
-        }
-        v[i].iov_len = n;
-        n = 0;
-    }
-    evbuffer_commit_space(buf, v, nv);
+
+    ssize_t sz = read(fd, p, avail);
+    if (sz <= 0 && (errno == EAGAIN || errno == EWOULDBLOCK))
+        sz = 0;
+
+    if (sz > 0 && evbuffer_add_reference(buf, p, sz, free_wrapper, p) != 0)
+        sz = -1;
+
+    if (sz <= 0)
+        free(p);
     return sz;
 }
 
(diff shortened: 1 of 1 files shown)


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>2/459 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>socksql_master_swings<br>phys_rep</p>

net/net_evbuffer.c
```diff
@@ -1697,38 +1697,43 @@ static void get_stat_evbuffer(struct event_info *e, net_queue_stat_t *stat)
...
-        v[i].iov_len = n;
-        n = 0;
-    }
-    evbuffer_commit_space(buf, v, nv);
+
+    ssize_t sz = read(fd, p, avail);
+    if (sz <= 0 && (errno == EAGAIN || errno == EWOULDBLOCK))
+        sz = 0;
+
+    if (sz > 0 && evbuffer_add_reference(buf, p, sz, free_wrapper, p) != 0)
```
akshatsikarwar, line 1732: this does become extra malloc/free for each read though.

net/net_evbuffer.c
```diff
@@ -3036,35 +3041,39 @@ int write_list_evbuffer(host_node_type *host_node_ptr, int type,
...
     Pthread_mutex_lock(&e->wr_lk);
     if (!e->flush_buf) {
        rc = -3;
     } else if ((rc = skip_send(e, nodrop, 0)) == 0) {
-        if ((rc = evbuffer_add_buffer(e->flush_buf, buf)) == 0) {
+        sz = e->wirehdr_len;
+        for (int i = 0; i < n; ++i)
+            sz += iov[i].iov_len;
+
+        const int nv = evbuffer_reserve_space(e->flush_buf, sz, v, 1);
```
akshatsikarwar, line 3056: Idea here was to do malloc without holding wr_lk

net/net_evbuffer.c
akshatsikarwar: Original problem being, too many calls to malloc and on profiling always catching it in memory allocation. Not sure want to go back to it.

riverszhang89: That makes sense. I don't know when the profiling was done but there have been a couple changes lately:

1. Per-thread malloc was enabled for event-based net in #2608. This helps reduce contention on the "net" allocator.

2. We used to allocate 8MiB memory here in `readcb()`. `Malloc()` would request directly from the system [using `mmap()`](https://github.com/bloomberg/comdb2/blob/master/dlmalloc/dlmalloc.c#L582) which can be quite expensive. Furthermore, memory obtained this way cannot be reused, as `free()` would simply call `munmap()`. This may explain why we were always seeing malloc allocation when profiling. This was changed to `FIONREAD` in https://github.com/bloomberg/comdb2/pull/3374, which is a much cheaper request to work with.

Furthermore, on a congested network, `reserve_space()` may wind up doing malloc for each read, too. Consider a slow network where messages arrive one at a time:

```c
1st msg

readcb()
  \
  reserve_space(rd_buf) 
  /* rd_buf has no available memory, so malloc(1024) */
    \
    readv()
      \
      evbuffer_remove_buffer(rd_buf, akq_buf, lengthof(1st_msg) 
      /* Transfer the entire "chain" to akq_buf, to avoid memcpy */
      
2nd msg

readcb()
  \
  reserve_space(rd_buf) 
  /* rd_buf has no available memory (was transferred to akq_buf), so malloc(1024) again!!! */
    \
    readv()
      \
      evbuffer_remove_buffer(rd_buf, akq_buf, lengthof(1st_msg) 
      /* Transfer the entire "chain" to akq_buf, to avoid memcpy */
  
```

riverszhang89: What's the benchmark? I can certainly test again.

net/net_evbuffer.c
riverszhang89: Gotcha. I think I can malloc, memcpy outside `wr_lk`, and do an `add_reference()` while holding `wr_lk`. Would that be better?

net/net_evbuffer.c
akshatsikarwar: Should we even be optimizing for degenerate case of slow or congested network? This is doubling the number of malloc/free for all cases. I saw the original commit which capped FIONREAD to 8MB. That seemed reasonable - but also if there is more to be read from the socket why not read a bit more at once (cap higher still..)

net/net_evbuffer.c
akshatsikarwar: add_reference is additional malloc/free calls and again under lock. I'm not sure why we are touching write_list tbh.

net/net_evbuffer.c
akshatsikarwar: benchmark is anything causing lots of reads - so replicants writing lots of bp-logs and leader writing lots of txn-logs.

akshatsikarwar: Tests were certainly after your per thread changes. It was not concurrency related so I don't think it did much in that way. Problem is that running malloc is expensive.

akshatsikarwar: should we instead look at `pool_getablk` to manage larger allocations? so they can be reused?

akshatsikarwar: Or perhaps cap FIONREAD to DEFAULT_MMAP_THRESHOLD_MAX (instead of TCP_BUFSZ) so it can be reused.

akshatsikarwar:
https://gist.github.com/akshatsikarwar/0af255dbc035046e93bd96b358580cbc
^^ I'm trying to validate this:
>```
> 1st msg
>
> readcb()
>   \
>   reserve_space(rd_buf)
>   /* rd_buf has no available memory, so malloc(1024) */
>     \
>     readv()
>       \
>       evbuffer_remove_buffer(rd_buf, akq_buf, lengthof(1st_msg)
>       /* Transfer the entire "chain" to akq_buf, to avoid memcpy */
>
> 2nd msg
>
> readcb()
>   \
>   reserve_space(rd_buf)
>   /* rd_buf has no available memory (was transferred to akq_buf), so malloc(1024) again!!! */
>     \
>     readv()
>       \
>       evbuffer_remove_buffer(rd_buf, akq_buf, lengthof(1st_msg)
>       /* Transfer the entire "chain" to akq_buf, to avoid memcpy */
>```

`malloc` on every call is killing performance in the test above. If anything, we should look at reserving minimum amount if lots of small messages are expected (not the small size that `FIONREAD` might report). I will keep looking to see what the original problem is and how it can be addressed.

Also, discovered that threshold for `mmap` is much higher and we should not be hitting that case (of not being able to use `mspace` chunks.) Need to figure out what was observed.

riverszhang89:
Closing this pull request, for Akshat's had a much better fix for this!!!

PR #3615, opened 2023-01-06
Fix remote write in higher isolation level

riverszhang89:
An FDB transaction's rqid isn't properly generated under noenv_message, in transaction isolation levels higher than socksql. This patch fixes it.


dorinhogea (approved):
Looks good to me

riverszhang89:
@dorinhogea Thanks for the review!

PR #3616, opened 2023-01-06
Fix remote write memory leaks

riverszhang89:
Each remote-write transaction leaks `sizeof osql_sqlthr_t` many bytes on the remote node. Additionally, if a remote-write transaction modifies an off-page field (blob/vutf8), it ends up leaking the new data it puts in, on the remote node, as well.

The patch fixes them.

(DRQS 170951962)


dorinhogea (approved):
This is good, thank you!

riverszhang89:
@dorinhogea Thank you so much for the review!

PR #3632, opened 2023-01-19
Don't bump dbopen_gen on analyze

mponomar:
Analyze running between statements in a stored procedure can cause it to return an error, if the stored procedure happens to be running on the leader node.


markhannum (approved):
Good to merge

PR #3641, opened 2023-01-24
libevent: Set decom flag directly

akshatsikarwar:
/silent


akshatsikarwar:
Skip calling into legacy net to set decom flag.

mponomar (approved):

PR #3642, opened 2023-01-26
Fix incorrect index position caused by loss of precision

riverszhang89:
Consider the following statements.

```sql
CREATE TABLE t (i int, j int)
CREATE INDEX t_ij on t(i, j)
INSERT INTO t VALUES (1.1, 1), (1.1, 2), (2, 3)
SELECT * FROM t WHERE i = 1.1 AND j = 1
```

The last SELECT statement returns wrong rows, as shown below:

```sql
(i=1, j=1)
(i=1, j=2)
```

To understand the issue, first consider the following statement.
```
SELECT * FROM t WHERE i = 1 AND j = 1
```

The cursor on the index is first positioned on the smallest entry that's greater than or equal to (1, 1), that is, the very 1st row in this table. We then continue traversing the index till we see an entry greater than the search key. Hence we stop at (1, 2), returning (1, 1) to client.

However, when searching (1.1, 1), the floating point number is rounded down to 1 (just like how it was inserted). Therefore, instead of being positioned on (2, 3), the cursor is placed on (1, 1), again. Now, we're going to traverse the index till we meet an entry greater than (1.1, 1), that is, (2, 3) (table has (1, 1), (1, 2) and (2, 3)).

To fix this, a new `roundup` conversion flag is introduced. When lit, `mem_to_ondisk`, instead of rounding the given value down, it rounds it up to the next higher value in its lower precsion data type. For example, 1.1 would be rounded up to 2; "2023-01-01T000000.000001" would be rounded up to "2023-01-01T000000.001".

(DRQS 171250145)

### DIFF
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -1064,6 +1064,7 @@ static int mem_to_ondisk(void *outbuf, struct field *f, struct mem_info *info,
     int outdtsz;
     int rc = 0;
     uint8_t *out = (uint8_t *)outbuf;
+    dttz_t dttz;
 
     if (m->flags & MEM_Null) {
         /* for local replicants, we need to supply a value here. */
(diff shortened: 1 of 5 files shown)


roborivers (commented):
Lint check:

```diff


@@ sqlite/src/vdbeInt.h @@
+ If the order of members in struct Mem is changed, columnNullValue (defined in sqlite/vdbeapi.c) needs to be reordered accordingly.
```

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>2/461 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>simple_timepart<br>verify_error</p>

akshatsikarwar:
See https://github.com/bloomberg/comdb2/pull/3643

PR #3676, opened 2023-02-13
need custom flag, cannot use resume for new shard sc-s

dorinhogea:
When creating new schema changes for missing shards during partition schema change resume, we cannot use the ->resume flag to skip the quota check (since that skips some critical sc initialization code).   Use a custom flag in this case.


riverszhang89 (approved):

dorinhogea:
Thank you Rivers!

PR #3698, opened 2023-02-22
block new query requests in remsql if incoherent

dorinhogea:
More upgrades to remsql, if node is incoherent, and the cursor is not part of a transaction, block it.

PR #3699, opened 2023-02-22
block new query requests in remsql if incoherent

dorinhogea:
Port https://github.com/bloomberg/comdb2/pull/3698 to 7.0


dorinhogea:
Thank you Mike!

PR #3725, opened 2023-03-09
Flush the log when recovering a checkpoint to avoid an assert failure

markhannum:
Checkpoints are flushed out-of-band, so this might not have made it to disk yet.


mohitkhullar (approved):
lgtm

PR #3729, opened 2023-03-09
carray: Update documentation

akshatsikarwar:
/skipbuild

### DIFF
--- a/docs/pages/programming/c_api.md
+++ b/docs/pages/programming/c_api.md
@@ -363,11 +363,11 @@ int cdb2_bind_array(cdb2_hndl_tp *hndl, const char *name, cdb2_coltype type, con
 ```
 
 Description:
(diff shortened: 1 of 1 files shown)


mponomar (dismissed):

docs/pages/programming/c_api.md
```diff
@@ -363,11 +363,11 @@ int cdb2_bind_array(cdb2_hndl_tp *hndl, const char *name, cdb2_coltype type, con
 ```
 
 Description:
-`Carray` is a table-valued function with a single column (named "value") and zero or more rows, corresponding to the values in the array. The "value" of each row in the carray() is taken from a C-language array supplied by the application via parameter binding by calling `cdb2_bind_array`. In this way, the carray() function provides a convenient mechanism to bind C-language arrays to SQL queries. There can only be a single `carray` parameter in a given SQL statement. This means only one array parameter can be used in a single SQL statement.
+`cdb2_bind_array` enables passing C-language array of values to a SQL query or to a stored procedure. In an SQL query, the parameter is passed to `carray` table valued function. `Carray` has a single column (named "value") and zero or more rows, corresponding to the values in the array. The "value" of each row in the carray() is taken from a C-language array supplied by the application.
```
chands10, line 366: You have "a SQL query" and "an SQL query" on the same line

chands10 (approved):

PR #3752, opened 2023-03-25
Use maven-dependency-plugin to download protoc protobuf compiler

hgeraldino:

It's a "feature", in the sense that it configures Maven to download the protoc compiler as part of the build process, instead of having developers download & install protobuf out-of-band.

Users won't have to figure out how to download a compatible version of protoc to build this project.

On a clean setup (a MacBook Pro M1 in my case), it is not possible to build the project with mvn clean package out of the box.

It is better, as it improves the SDLC experience.


riverszhang89 (approved):
Sorry for taking so long to get to this pull request. It runs flawlessly on my laptop. The pull request is good to merge!

Issue #3755, opened 2023-03-27
OOM error on Comdb2 JDBC driver


hgeraldino:
**Describe the bug**
Each `Comdb2Connection`  has an internal `Comdb2Handle` that is reused across all statements for the given connection. Because of this,  there's a potential for the internal `List<String> set` commands that is used to keep track of the `set` commands to grow unbounded and cause OOM errors. This is not a theoretical problem: we've encountered this issue when running the latest `2.6.0` version of the JDBC driver.

**To Reproduce**
Please refer to the https://github.com/hgeraldino/comdb2-oom-example repository, which has a sample Integration Test that can be used to reproduce the problem.

**Expected behavior**
Each `java.sql.Statement` should only care about the `set` SQL statements of its own.

**Screenshots**
The referenced project has a Screenshot of a heap dump taken from one of our running applications. It shows the enormous number of String references kept by the `Comdb2Handle` .

**Environment (please complete the following information):**
 - Operating System and Version: 
```bash
$ uname -a
Linux <hostname> 3.10.0-1160.80.1.el7.x86_64 #1 SMP Sat Oct 8 18:13:21 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
$ java -version
openjdk version "17.0.6" 2023-01-17
OpenJDK Runtime Environment Temurin-17.0.6+10 (build 17.0.6+10)
OpenJDK 64-Bit Server VM Temurin-17.0.6+10 (build 17.0.6+10, mixed mode, sharing)
```

Brunosampaiomartins:
SQL Error: Internal jdbc driver error
  Internal jdbc driver error
  Internal jdbc driver error
    Java heap space
    Java heap space
    
    Alguém sabe pq esse erro?

hgeraldino:
> SQL Error: Internal jdbc driver error Internal jdbc driver error Internal jdbc driver error Java heap space Java heap space
> 
> ```
> Alguém sabe pq esse erro?
> ```

Might be related, but it's hard to tell without a heap dump. I'd suggest [taking a heap dump](https://www.baeldung.com/java-heap-dump-capture) and examining its contents in order to pinpoint the cause

PR #3756, opened 2023-03-27
Maintain a unique set of "set" SQL statements

hgeraldino:
See issue https://github.com/bloomberg/comdb2/issues/3755 for info about this patch

The proposed change splits the `set` SQL statements into two separate sets:
* A `pendingSetStmts` that keeps track of the `set` SQL statements that have been added by the API/caller but not executed just yet
* A `sentSetStmts` with the list of "set" statements that have been executed. This separate list is maintained so already executed "set" statements can be queued/executed again in the case of a re-connection.

Both sets are implemented as linked lists, as elements are removed before they are added again. This is done in order to avoid having an ever growing number of set commands added to the list, which causes OOM errors in applications with lots of activity.

### DIFF
--- a/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
+++ b/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
@@ -13,16 +13,12 @@
    limitations under the License. */
 package com.bloomberg.comdb2.jdbc;
 
-import java.nio.ByteBuffer;
 import java.io.*;
 import java.util.*;
 import java.util.logging.*;
 import javax.net.ssl.*;
 import java.text.MessageFormat;
 
-import com.google.protobuf.*;
-
-import com.bloomberg.comdb2.jdbc.SockIO;
 import com.bloomberg.comdb2.jdbc.Cdb2Query.*;
 import com.bloomberg.comdb2.jdbc.Constants.*;
 import com.bloomberg.comdb2.jdbc.Sqlquery.*;
@@ -93,7 +89,8 @@ public class Comdb2Handle extends AbstractConnection {
 
     HashMap<String, Cdb2BindValue> bindVars;
     HashMap<Integer, Cdb2BindValue> bindVarsByIndex;
-    private List<String> sets;
+    private List<String> sentSetStmts; // Collection of "set" SQL statements sent for this Connection
+    private List<String> pendingSetStmts; // Collection of "set" SQL statements pending to be sent
 
     private boolean ack = false;
     private boolean skipDrain = false;
@@ -112,7 +109,6 @@ public class Comdb2Handle extends AbstractConnection {
     private int snapshotOffset;
     private boolean isHASql = false;
     private int isRetry;
-    private int nSetsSent;
     private int errorInTxn = 0;
     private boolean readIntransResults = true;
     private boolean firstRecordRead = false;
@@ -205,7 +201,8 @@ public class Comdb2Handle extends AbstractConnection {
        without discovering twice. */
     public Comdb2Handle() {
         super(new ProtobufProtocol(), null);
-        sets = new ArrayList<String>();
+        sentSetStmts = new LinkedList<String>();
+        pendingSetStmts = new LinkedList<String>();
 
         /* CDB2JDBC_STATEMENT_QUERYEFFECTS and CDB2JDBC_VERIFY_RETRY
            are used by the Jepsen tests to change the driver's behaviors. */
@@ -229,12 +226,12 @@ public class Comdb2Handle extends AbstractConnection {
 
         String userEnv = System.getenv("COMDB2_USER");
         if (userEnv != null) {
-            sets.add("set user " + userEnv);
+            addSetStatement("set user " + userEnv);
         }
 
         String passwordEnv = System.getenv("COMDB2_PASSWORD");
         if (passwordEnv != null) {
-            sets.add("set password " + passwordEnv);
+            addSetStatement("set password " + passwordEnv);
         }
 
         uuid = UUID.randomUUID().toString();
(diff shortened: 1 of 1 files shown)


cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
```diff
@@ -93,7 +89,8 @@ public class Comdb2Handle extends AbstractConnection {
 
     HashMap<String, Cdb2BindValue> bindVars;
     HashMap<Integer, Cdb2BindValue> bindVarsByIndex;
-    private List<String> sets;
```
riverszhang89, line 96: Wondering what if you just change the List to a Set?

hgeraldino: Thanks for reviewing this @riverszhang89 

My first thought was to change this field type to `LinkedHashSet` (which is ordered), but faced two issues:

1. The `java.util.Set` interface doesn't provide a `subList(...)` (or equivalent) method, which is what was being used when deciding which `set` commands to append to the `SQLQuery`. Of course we can provide our own implementation of this method, but it won't help because...
2. A side effect of the current behavior is that `set` commands configured in one `Statement` could be propagated to subsequent statements. The code made an attempt to avoid this by keeping a 'counter' (nSetsSent) of the `set` commands already sent, but that counter gets reset in some situations (e.g. if there's a re-connection), which means that there's a chance for any `Statement` to send `set` commands configured for a previous/different statement, which makes the behavior non-deterministic.

Keeping the collection of `set` commands per statement and clearing it once the statement is executed provides stronger guarantees IMO.

riverszhang89: A set statement isn't directly sent to server. It's embedded in the next SQL statement. A counter keeps track of the last sent set command in the list so that the API knows what set statements need to be packed into the next SQL query. For instance,

```sql
SET TIMEZONE UTC
SELECT NOW() -- counter is at 0; so pack 'SET TIMEZONE UTC' into this query
SET MAXQUERYTIME 1
SET VERIFYRETRY OFF
SELECT SLEEP(10) -- counter is at 1; so pack 'SET MAXQUERYTIME 1' and 'SET VERIFYRETRY OFF' into this query
```

Set statements are also per-connection, in Comdb2. There isn't a so called 'multiple-statement-execution' in comdb2. You can certainly create multiple java.sql.statement objects on a single java.sql.connection, but they end up sharing each other's set statements. This is by design. This is why the counter resets to 0 if server disconnects, that the API can restore all set statements previously issued, when failing over to the next node.

On a second thought, I don't think we can use linkedhashset either (at least not without overriding its add()). Consider the following statements:

```sql
SET TIMEZONE UTC
SET TIMEZONE EST
SET TIMEZONE UTC
```
Say the timezone is changing back and forth (maybe because it's loaded from a config file, or maybe it's coming from an UI where the user switches timezone a couple times). Now, with a List, the final timezone is UTC; but with a linkedhashset, the 2nd 'SET TIMEZONE UTC' is getting ignored, leaving the final timezone EST.

How about just removing the string from the list before adding it? something like this:
```java
String set_statement;
sets.remove(set_statement);
sets.add(set_statement);
```

This way you'd avoid inserting duplicates into the list, and at the same time maintain the order of insertion.

hgeraldino: Ok makes sense. I'm not a fan of the delete-then-add sequence (items on the backing array get reshuffled constantly), but will do it if that's the only option

hgeraldino: Updated the title & description, also re-implemented the patch.

I still had to add an additional data structure to keep track of the `set` commands that have been executed, instead of having a single list with a position/counter. Using the single list + counter didn't work in some cases. Following your example:

```sql
SET TIMEZONE UTC #T0
SET TIMEZONE EST #T1
SET TIMEZONE UTC #T2
```

* When the SET command in T0 is executed, the counter `nSetsSent` value will be set to 1
* Then, at T1 , the counter `nSetsSent` value will increase to 2
* At T2, if we delete-then-insert the SET command setting the time zone to UTC, the size of `sets` remains the same (2 items), so using `nSetsSent` to calculate the sub-list of unsent commands won't return the correct value (will return zero elements in this case, which will cause the last SET not to be executed).

I think the PR maintains the same principle:
* There's a list of SET statements we need to add to the query
* The list cannot grow forever, so it's safe to remove duplicates while maintaining order
* We don't want to re-send the same SET commands already set for the connection
* If the connection ever closes and we need to reconnect, we need a way for replaying the SET commands already sent

riverszhang89:
@hgeraldino I like your idea of keeping 2 lists of set commands. Can you please rebase and squash commits into one?

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>9/473 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>cldeadlock<br>logfill<br>ruleset_no_fp_generated<br>ruleset<br>enospc<br>mem_tracker<br>queueing_sql<br>sc_timepart_logicalsc_generated<br>analyze_exit</p>

riverszhang89 (approved):

riverszhang89:
@hgeraldino Thanks for the pull request!!!

hgeraldino:
> [Coding style](https://bloomberg.github.io/comdb2/contrib.html#code-style) check: Success ✓.Smoke testing: Error ⚠.Cbuild submission: Error ⚠.Regression testing: 9/473 tests failed ⚠.
> 
> The first 10 failing tests are: cldeadlocklogfillruleset_no_fp_generatedrulesetenospcmem_trackerqueueing_sqlsc_timepart_logicalsc_generatedanalyze_exit

Rebased/squashed

PR #3772, opened 2023-04-08
Fix for Dockerfile smell DL3007

grosa1:

Refactoring of Dockerfile for writing best practices

None

None

The Dockerfile contains a best practice violation (smell) that could lead to reliability issues

Setting a specific version tag ensures the build reproducibility

* Description:

Hi!
The Dockerfile placed at "tests/netloss.test/Dockerfile.dev" contains the best practice violation [DL3007](https://github.com/hadolint/hadolint/wiki/DL3007) detected by the [hadolint](https://github.com/hadolint/hadolint) tool.

The smell DL3007 occurs when the tag "latest" is used instead of a specific version tag for the base image.
In this pull request, we propose a fix for that smell generated by our fixing tool. We have verified that the patch is correct before opening the pull request.
To fix this smell, specifically, we use a heuristic approach that selects the most probable version tag for the base image in order to replace the "latest" tag. In detail, it selects the most recent image tag which corresponds to the same image digest that currently corresponds to the "latest" tag.

This change is only aimed at fixing that specific smell. If the fix is not valid or useful, please briefly indicate the reason and suggestions for possible improvements.

Thanks in advance

### DIFF
--- a/tests/netloss.test/Dockerfile.dev
+++ b/tests/netloss.test/Dockerfile.dev
@@ -1,4 +1,4 @@
-FROM ubuntu:latest
+FROM ubuntu:22.04
 
 RUN apt-get update && \
   apt-get install -y \


nirbhayc (commented):
I think switching from latest to the latest LTS is a good change.

@grosa1 In order for us to merge this change you'll need to signoff the patch.

riverszhang89 (approved):

PR #3781, opened 2023-04-14
{170477625}: Fixing dbstore function

riverszhang89:
Rebuilding a new field with a dbstore function crashes the database, as there's no `SERVER_to_SERVER` routine for the `SERVER_FUNCTION` datatype. This patch fixes it.

There're a few places where such a dbstore function can be evaluated for existing records. 

1) It can be evaluated per record at sc-convert-record time. This approach obviously requires a rebuild.

2) It can be evaluated once at sc-commit-time. The advantage of this is that it works with both instant and non-instant schema change. The disadvantage is that all existing rows get the same value, which may cause other problems down the line (uniqueness violation, for instance).

3) It can be evaluated per bdb-cursor move (this is where a literal dbstore is processed). The problem with this approach is that selecting the same row at different times may produce different values for this field (`now()`, `guid()`, etc.).

4) It can be not evaluated at all, leaving the new field for all existing rows NULL. This may cause other problems down the line (nullability violation, for instance).

None of these are perfect but option 1 seems to make the most sense. This patch also implements that.

### DIFF
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -13173,3 +13173,68 @@ int comdb2_is_field_indexable(const char *table_name, int fld_idx) {
     }
     return 1;
 }
+
+/* Evaluate an SQL function, and convert its SQLite Mem to ondisk data. */
+int sql_func_to_ondisk(char *func, void *out, struct field *f, struct convert_failure *fail_reason, const char *tzname)
+{
+    char *sql;
+    struct sqlclntstate clnt;
+
+    struct schema_mem sm = {0};
+    struct schema empty_schema = {0};
+    Mem empty_mem = {{0}};
+    Mem mout = {{0}};
+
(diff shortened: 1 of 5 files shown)


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>1/474 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>sc_timepart_logicalsc_generated</p>

riverszhang89:
new idea: always require a dbstore function to be null-able. existing records get null (and hence won't lose uniqueness). the dbstore function is evaluated only for new records.

PR #3832, opened 2023-05-16
Update commit LSN map test case

morgando:
This PR makes the following changes to the commit LSN map test:

- Replaces the checkpoint command with the flush command to force a checkpoint where it's needed.
- Adds some debug output.
- Adds a debug procedure `clm_delete_logfile <n>` to delete a transactions that committed in a specific logfile from the map. Uses this procedure to test map deletion instead of triggering the same process via log deletion. This cuts the test time by 5+ minutes and should prevent it from sometimes timeing out.

### DIFF
--- a/tests/commit_lsn_map.test/runit
+++ b/tests/commit_lsn_map.test/runit
@@ -74,18 +74,14 @@ function update_tranlog_copy
 function verify_child_commit_lsn_is_parent_commit_lsn
 {
 	update_tranlog_copy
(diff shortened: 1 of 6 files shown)


tests/commit_lsn_map.test/runit
```diff
@@ -233,100 +238,48 @@ function test_add
...
-	    # Saw something that needed to be archived: break out of loop
-	    else
-
-		# Break out of this loop
-		seen_archives=1
-
-	    fi
+		sleep 3
 
+		res=$(cdb2sql ${CDB2_OPTIONS} $dbnm default 'select MAX(commitlsnfile) from comdb2_transaction_commit')
```
riverszhang89, line 247: there's a --tab option that returns rows without parentheses.

riverszhang89 (approved):

PR #3837, opened 2023-05-19
Fixing gcc v9.4 warning 'format-truncation'

JamesWP:
In both cases we either increase the size of the buffer to be a little more generous or check the error and return an error to the caller

> Warn about calls to formatted input/output functions such as snprintf
> and vsnprintf that might result in output truncation. When the exact
> number of bytes written by a format directive cannot be determined
> at compile-time it is estimated based on heuristics that depend on
> the level argument and on optimization. While enabling optimization
> will in most cases improve the accuracy of the warning, it may also
> result in false positives. Except as noted otherwise,
> the option uses the same logic -Wformat-overflow.
>
> Link: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#:~:text=Wformat%2Dtruncation%20%C2%B6

After the change, the code compiles without warning.

OS: Ubuntu 20.04.5 LTS

### DIFF
--- a/db/views.c
+++ b/db/views.c
@@ -792,12 +792,16 @@ static int _generate_new_shard_name(const char *oldname, char *newname,
         /* get the length of the number sufix */
         suffix_len = _shard_suffix_str_len(maxshards);
 
-        snprintf(newname, newnamelen, "%s%.*d", oldname, suffix_len, nextnum);
+        if (0 > snprintf(newname, newnamelen, "%s%.*d", oldname, suffix_len, nextnum)) {
+	    return VIEW_ERR_GENERIC;
+	}
     } else {
         char hash[128];
         len = snprintf(hash, sizeof(hash), "%u%s", nextnum, oldname);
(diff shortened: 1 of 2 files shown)


db/views.c
```diff
@@ -792,12 +792,16 @@ static int _generate_new_shard_name(const char *oldname, char *newname,
         /* get the length of the number sufix */
         suffix_len = _shard_suffix_str_len(maxshards);
 
-        snprintf(newname, newnamelen, "%s%.*d", oldname, suffix_len, nextnum);
+        if (0 > snprintf(newname, newnamelen, "%s%.*d", oldname, suffix_len, nextnum)) {
```
riverszhang89, line 795: Thanks for the pull request! I think it should check for `snprintf(...) >= newnamelen`, instead. snprintf returns number of bytes written. If output is truncated, it returns number of bytes which would've been written as if output had not been truncated. So any return value >= `newnamelen` means truncation.

riverszhang89 (approved):

PR #3822, opened 2023-05-09
{171141266}: Ensure that heartbeat is stopped before sending last row

riverszhang89:
A stored procedure may send a heartbeat after the last row, which can break the SSL protocol when sockpool is present, as shown below.

|    | SERVER               |     CLIENT-1                                 |     CLIENT-2                            |
|----|----------------------|----------------------------------------------|-----------------------------------------|
| T0 | send ROW_LAST        |//////////////////////////////////////////////|/////////////////////////////////////////|
| T1 |//////////////////////| recv ROW_LAST, this is considered an OK_DONE |/////////////////////////////////////////|
| T2 | send HEARTBEAT       |//////////////////////////////////////////////|/////////////////////////////////////////|
| T3 |//////////////////////| cdb2_close; connection returned to sockpool  |/////////////////////////////////////////|
| T4 |//////////////////////|//////////////////////////////////////////////| cdb2_open; reuse the connection from T3; start SSL protocol; send SSLCONN packet |
| T5 | recv SSLCONN; call SSL_accept           |//////////////////////////////////////////////|/////////////////////////////////////////|
| T6 |//////////////////////|//////////////////////////////////////////////| read 1st byte of the HEARTBEAT from T2; since it's not 'N', call SSL_connect.   now read the remainder of the heartbeat, which isn't part of SSL negotiation. SSL_connect fails.                      |
| T8 | SSL_accept fails     |//////////////////////////////////////////////|/////////////////////////////////////////|

This patch fixes it by switching off the ready_for_heartbeats flag before sending ROW_LAST, so that no heartbeats will go out after ROW_LAST and potentially be read by the next process that reuses the connection. The SSL-ability byte check in the API is also changed from `rc == 'N'` to `rc != 'Y'`, to avoid mishandling of a heartbeat packet (as shown at T6 in the table above).

PR #3868, opened 2023-06-13
Do not donate socket if SSL_shutdown fails

riverszhang89:
SSL_shutdown() will fail if there is unread data on an SSL connection. The patch changes the API to check the return code from SSL_shutdown(), and to only donate the underlying socket after SSL_shutdown() succeeds.

While #3822 successfully reproduces a case in 7.0 that misleads the API to incorrectly donate a undrained socket to sockpool, I've struggled to get it to happen in 8.0. Therefore, in this patch I've also added instrumentation to catch any protobuf response type other than LAST_ROW and immediately before a RESET from sockpool, as we believe that such a packet causes the behavior in the first place.

### DIFF
--- a/cdb2api/cdb2api.c
+++ b/cdb2api/cdb2api.c
@@ -2181,10 +2181,12 @@ static int try_ssl(cdb2_hndl_tp *hndl, SBUF2 *sb)
 
     /* The node does not agree with dbinfo. This usually happens
        during the downgrade from SSL to non-SSL. */
-    if (rc == 'N') {
+    if (rc != 'Y') {
         if (SSL_IS_OPTIONAL(hndl->c_sslmode)) {
             hndl->c_sslmode = SSL_ALLOW;
-            return 0;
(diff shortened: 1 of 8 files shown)


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>6/478 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>sc_truncate<br>sc_resume<br>morestripe<br>incremental_backup_usenames<br>queuedb_rollover_noroll1_generated<br>sc_timepart_logicalsc_generated</p>

mohitkhullar (approved):
lgtm

riverszhang89:
@mohitkhullar Thanks for the review!

PR #3885, opened 2023-06-30
add support for table alias in dohsql

dorinhogea:
Per title; example:
Before:
 c "select t.a from t left join (select a from t) tt where tt.a = t.a union all  select 1 from t left join (select a from t) limit 2"
[select t.a from t left join (select a from t) tt where tt.a = t.a union all  select 1 from t left join (select a from t) limit 2] failed with rc -3 no such column: tt.a
Now:
 c "select t.a from t left join (select a from t) tt where tt.a = t.a union all  select 1 from t left join (select a from t) limit 2"
(a=1)
(a=1)


riverszhang89 (approved):

dorinhogea:
Thank you Rivers!

PR #3909, opened 2023-07-14
RFC: PRAGMA CHUNK

riverszhang89:
A chunk "transaction" requires to be wrapped in a transaction explicitly using BEGIN and COMMIT, though it's not atomic. This usage has caused a lot of confusion. This patch proposes a new syntax for chunk transactions, using SQLite's PRAGMA:

```
SET TRANSACTION CHUNK N
PRAGMA CHUNK=ON -- begin a chunk tran
INSERT INTO T VALUES(...)
PRAGMA CHUNK=OFF -- finish a chunk tran
```

The CHUNK pragma and `BEGIN..COMMIT` behave mostly identically, except that the CHUNK pragma disables the skip-row optimization. To make these two work together, it would require the client to recognize the pragma and treat it as a BEGIN, and that means client-side parsing and a new client feature to be backwards compatible. Since chunk is mostly used in scheduled jobs that run only periodically where performance isn't the greatest concern, I think I'm okay with this tradeoff.

### DIFF
--- a/db/sql.h
+++ b/db/sql.h
@@ -747,6 +747,7 @@ struct sqlclntstate {
     struct query_effects effects;
     struct query_effects log_effects;
     struct query_effects chunk_effects;
+    int chunk_pragma; /* is it a chunk pragma? */
     int64_t nsteps;
 
     struct user current_user;
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -13194,4 +13194,3 @@ int comdb2_is_field_indexable(const char *table_name, int fld_idx) {
     }
     return 1;
 }
-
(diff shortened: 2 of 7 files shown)


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>3/482 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>sc_transactional_rowlocks_generated<br>queuedb_rollover_noroll1_generated<br>queuedb_rollover</p>

PR #3912, opened 2023-07-18
Physrep: Fix an issue with 'reverse connection'

nirbhayc:
'reverse connection' did not work properly when 'sources' table was hosted in the source node/cluster (as against in a separate replication metadb). This was due to a restrictive faulty check which kept reverse connection manager from starting.

### DIFF
--- a/berkdb/log/log_put.c
+++ b/berkdb/log/log_put.c
@@ -2324,23 +2324,11 @@ __log_rep_put(dbenv, lsnp, rec)
 			goto err;
 		}
 
-                // TODO (NC): enable?
-#if 0
-		/*
-		 * If we changed files and we're in a replicated
-		 * environment, we need to inform our clients now that
-		 * we've dropped the region lock.
-		 *
-		 * Note that a failed NEWFILE send is a dropped message
-		 * that our client can handle, so we can ignore it.  It's
-		 * possible that the record we already put is a commit, so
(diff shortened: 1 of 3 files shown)


roborivers (commented):
Lint check:

```diff


@@ db/reverse_conn.c @@
+ Please replace any pthread_xxx() function in your diff with corresponding Pthread_xxx() wrapper.
```

berkdb/rep/rep_method.c
```diff
@@ -1704,9 +1705,11 @@ __rep_flush(dbenv)
...
 		goto err;
 
-	/* treat the end of the log as perm */
-	(void)__rep_send_message(dbenv,
-	    db_eid_broadcast, REP_LOG, &lsn, &rec, DB_LOG_PERM, NULL);
+        if (last_lsn.file != lsn.file || last_lsn.offset != lsn.offset) {
+            last_lsn = lsn;
+            (void)__rep_send_message(dbenv,
+                db_eid_broadcast, REP_LOG, &lsn, &rec, DB_LOG_PERM, NULL);
+        }
```
nirbhayc, line 1712: This is needed (per Mike).

nirbhayc: Reverted

mponomar (approved):

PR #3922, opened 2023-07-26
rename unique_tag_key to thd_info_key to properly identify use

dorinhogea:
trivial rename; key maybe was originally used to create unique tag names, but the purpose evolved; new name reflect that.


mponomar (approved):

dorinhogea:
Thank you Mike, Aakash!

PR #3948, opened 2023-08-16
Remove xstring

aakash10292:
while investigating RDSICOMDB2-288, which turned out to be a false positive, it was discovered that we don't use this library anymore.


aakash10292:
akshat has a PR https://github.com/bloomberg/comdb2/pull/3947

PR #3954, opened 2023-08-21
skip parameterized queries in proxy fdb

dorinhogea:
Code currently does not push parameters breaking queries when fdb proxy is enabled.  Disable this optimization in cases where parameters are present (until support can be added).


nirbhayc (approved):

dorinhogea:
Thank you Nirbhay!

PR #3967, opened 2023-08-29
do not leak unpacked rows when more than one column

dorinhogea:
Remote sql proxy mode has a leak, unpacking a remote row repeatedly when more than column is needed.  Trivial fix.


morgando (approved):

dorinhogea:
Thank you Morgan!

PR #3971, opened 2023-09-02
{171428666}: Adding utf8_valid() sql function

riverszhang89:
The type system is also tweaked to accept trailing NUL characters in a UTF8 byte sequence.

### DIFF
--- a/db/types.c
+++ b/db/types.c
@@ -3695,6 +3695,21 @@ TYPES_INLINE int CLIENT_BLOB_to_CLIENT_PSTR2(
     return -1;
 }
 
+static int utf8_validate_trailing_zeros(const char *u, int max)
+{
+    int valid_len;
+
+    if (utf8_validate(u, max, &valid_len) != 0)
+        return -1;
+
+    /* utf8_validate() stops at the 1st NUL character. We want to permit trailing zeros */
+    for (; valid_len < max - 1; ++valid_len) {
+        if (u[valid_len] != '\0')
+            return -1;
+    }
+    return 0;
+}
+
 /**
  * Finds out where the input vutf8 string is stored and then determines where it
  * should be copied and copies it. Doesn't deal with NULLs.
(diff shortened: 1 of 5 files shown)


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>1/487 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>prepare</p>

akshatsikarwar:
I don't think the default behavior should be to accept trailing nulls. Perhaps as an option: `utf8_valid(column_name/literal, optional second argument to accept trailing zeros)`. Also, rename function so it is symmetrical to existing ones (`json_valid()`, etc). Reads better as well: `update foo set bar = null where utf8_valid(bar) == 0`

riverszhang89:
The change that accepts trailing zeros, is to work around a "corruption" (see DRQS linked), where a utf8 byte sequence has 2 trailing NUL characters. `utf8_validate()` stops at the 1st NUL, hence returning the 'actual-valid-length' as `strlen()-1`. Now, the type system verifies if the entire byte sequence is valid, by checking if actual-valid-length equals `strlen()`. And because of that, the type system rejects this utf8 sequence.

My guess is that years ago we had a bug in handling of NUL where we explicitly padded with a NUL terminator without checking if the payload is already NUL-terminated. I don't think it's possible to insert new data like this anymore, so the `utf8_validate_trailing_zeros` routine most definitely only has an effect on old data.

riverszhang89:
`utf8_validate()` behaves slightly differently from `json_valid()`. `utf8_validate()` isn't boolean. It returns 0 if the byte sequence validates okay, and return -(N-1) otherwise, where N is the 1st encountered malformed character. I think `utf8_validate()` would sound better, though I don't have a strong opinion about this.

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>2/511 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>sc_truncate_lockorder_generated<br>queuedb_rollover_noroll1_generated</p>

PR #3985, opened 2023-09-11
Fixing crash introduced by previous checkin

riverszhang89:
A child worker of a parallel-union-all query does not have `clnt->thd`. This patch adds a check for it (Thanks, Dorin!).


dorinhogea (approved):
LGTM!

riverszhang89:
@dorinhogea Thanks for the review!

PR #4002, opened 2023-09-20
{173038130}: Fixing clnt race

riverszhang89:
This is a regression from #3778. The idea of that pull request was to clear `sqlthd->clnt` in `signal_clnt_as_done()`, while holding the global `gbl_sql_lock` mutex. But dohsql child worker does not call `signal_clnt_as_done()`.

This patch adds back the code to clear `sqlthd->clnt`, in `sqlengine_work_shard()`.

### DIFF
--- a/db/dohsql.c
+++ b/db/dohsql.c
@@ -228,6 +228,13 @@ static void sqlengine_work_shard(struct thdpool *pool, void *work,
         handle_child_error(clnt, clnt->query_rc);
     }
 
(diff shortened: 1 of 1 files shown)


akshatsikarwar (approved):

roborivers (commented):
Lint check:

```diff


@@ db/dohsql.c @@
+ Please make sure that a mutex is unlocked before returning.
```

PR #4024, opened 2023-10-06
Add integration tier to the list of environments

hgeraldino:
Add a new entry to the list of recognized environments so support the `integration` tier

Using a connection string such as `jdbc:comdb2://integration/<dbname>?comdb2dbname=comdb3db` should work (it doesn't today)

Try connecting to any database in the `integration` tier - it shouldn't be possible without this change.


hgeraldino:
cc @akshatsikarwar @riverszhang89 

Happy weekend! :)

riverszhang89 (approved):
This works, however please take a look at https://github.com/bloomberg/comdb2/pull/3874, which I believe is more complete and does not require changing driver every time a new tier is introduced.

PR #4053, opened 2023-10-25
Fix `ar_compress` and `ar_decompress` not set bug

zixuanzhang-x:
## Issue (Bug)
The **current behavior** is that if `lz4` exists on the default path (`/opt/bb/bin/lz4`), `[[ ! -x $lz4 ]]` would evaluate to `false`, and `ar_compress` and `ar_decompress` wouldn't be set to correct value and therefore undefined.

The **expected behavior** is that `ar_compress` and `ar_decompress` should be set in all circumstances.

## Reproduce the bug
Run `copycomdb2` on a host where `/opt/bb/bin/lz4` does exist.

### DIFF
--- a/db/copycomdb2
+++ b/db/copycomdb2
@@ -144,21 +144,23 @@ lz4_rmt=
 do_recovery=yes
 
 [ -z "$comdb2ar" ] && comdb2ar="${PREFIX}/bin/comdb2ar"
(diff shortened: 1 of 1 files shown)


tvsaru (approved):
LGTM

markhannum (approved):
Thank you!  This looks great

PR #4067, opened 2023-11-01
schema change does not stop threads anymore

dorinhogea:
Left-over from the time when we stopped threads during schema changes.  Removing since it might collide with a pending exit.


mponomar (approved):

dorinhogea:
@mponomar thank you!

PR #4069, opened 2023-11-02
account for nameless chron-s

dorinhogea:
instead of crash


akshatsikarwar (approved):

dorinhogea:
@akshatsikarwar thank you!

PR #4076, opened 2023-11-06
when failing or aborting a schema change, need more cleaning

dorinhogea:
Fix in https://github.com/bloomberg/comdb2/pull/4054 introduced a regression; we leak the in memory tracking for schema change (as set by sc_set_running).  This is the fix.


chands10 (approved):

dorinhogea:
@chands10 thank you!

PR #4083, opened 2023-11-09
when a table is partitioned, name is aliased; needs special handling during unpartitioning

dorinhogea:
Partitioning an existing table creates an alias for it to serve as first shard; if the partition is merged back to the original table, special code is required to remove the alias.


mponomar (approved):

dorinhogea:
Thank you Mike!

PR #4097, opened 2023-11-27
cannot use blindly table name, need to use actual source item 

dorinhogea:
This never worked, but it broke the execution when join support for parallel sql execution was added. 
Since the expression in the return set, ExprList[], does not have the alias, we need to match cursor numbers of any TK_Column expression with the tables in SrcList[], which has the alias.


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>0/0 tests failed</b> &#x26A0;.</p>

mponomar (approved):

dorinhogea:
@mponomar thank you

PR #4101, opened 2023-11-29
eliminate redundant sqlite master records creation

dorinhogea:
Sql schema records created by create_sqlmaster_records & create_sqlite_master are updated in block processor upon return, we do not need to do it for each of the schema changes in do_X functions.

PR #4104, opened 2023-12-01
need to update schema records for comdb2sc too

dorinhogea:
Regression for https://github.com/bloomberg/comdb2/pull/4101
Once legacy interface is gone, all !is_osql code will be removed too so this addition will be gone too.  Until then, we need it!


mponomar (approved):

dorinhogea:
@mponomar thank you!

PR #4109, opened 2023-12-05
UPSERT: Fix a specific case of INSERT OR IGNORE failure

nirbhayc:
In the current UPSERT implementation we check of non-upsert targets (unique indices) for conflicts first; causing the INSERT to fail when there are conflicts in multiple unique indices (including the upsert target/index).

Example:
```
CREATE TABLE t1(i INT UNIQUE, j INT UNIQUE)$$
INSERT INTO t1 VALUES(1,1);
INSERT INTO t1 VALUES(1,1) ON CONFLICT(i) DO NOTHING; # This fails with dup error on index (j)
```

This behavior, however, is inconsistent with how Sqlite or Postgres handle the last INSERT where they allow it to succeed.

This patch fixes this behavior.


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>1/510 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>scindex</p>

riverszhang89 (approved):

PR #4112, opened 2023-12-06
Mike found this obvious bug that started to manifest recently; 173683794

dorinhogea:
obvious bug, pretty old too, started to manifest recently probably because schema change started to deadlock more often.
credit goes to @mponomar !


dorinhogea:
@mponomar thank you!

PR #4125, opened 2023-12-13
Detect uncommitable upserts

morgando:
A blocksql transaction that inserts the same key multiple times with upsert statements is not detected as uncommittable. So it retries the maximum number of times before failing.

The following is an example of such a transaction:
```
create table t(i int primary key)
begin
insert into t(i) values(1) on conflict (i) do update set i=2
insert into t(i) values(1) on conflict (i) do update set i=2
commit
```

The changes in this PR cause this type of transaction to fail as soon as extended checking kicks in: A hash structure tracks the keys that the running transaction has inserted. Once the running transaction has tried to insert the same key twice, the transaction is considered uncommittable and is failed.

### DIFF
--- a/db/indices.c
+++ b/db/indices.c
@@ -235,6 +235,33 @@ static int check_index(struct ireq *iq, void *trans, int ixnum, blob_buffer_t *b
     return 0;
 }
 
+/* Add unique key to per-transaction tracking structure.
+ *
+ * @return 1 : Fail. Key already exists in structure.
+ *         0 : Success.
+ */
+int track_record_index(struct ireq *iq, int ixnum, void *key, int ixkeylen)
+{
+    int is_dup;
+    char * store_key;
+
+    is_dup = 0;
+    store_key = calloc(sizeof(int) + sizeof(int) + MAXKEYLEN, sizeof(char));
+
+    memcpy(store_key, &iq->usedb->dbs_idx, sizeof(int));
+    memcpy(store_key + sizeof(int), &ixnum, sizeof(int));
+    memcpy(store_key + sizeof(int) + sizeof(int), key, ixkeylen);
+
+    char * s = hash_find(iq->vfy_idx_hash, store_key);
+    if (s) {
+        is_dup = 1;
+        free(store_key);
+    } else {
+        hash_add(iq->vfy_idx_hash, store_key);
+    }
+    return is_dup;
+}
+
 /* If a specific index has been used in the ON CONFLICT (idx) DO NOTHING
  * clause (aka upsert target/index), then we must first perform check for
  * that particular index and bail out (ignore) if there's a similar entry
@@ -305,6 +332,7 @@ int add_record_indices(struct ireq *iq, void *trans, blob_buffer_t *blobs,
                        int flags, int reorder)
 {
     int rc = 0;
+    int dup_txn_insert = 0;
     char *od_dta_tail = NULL;
     int od_tail_len;
     if (iq->osql_step_ix)
@@ -427,6 +455,10 @@ int add_record_indices(struct ireq *iq, void *trans, blob_buffer_t *blobs,
             rc = ix_addk(iq, trans, key, ixnum, *genid, *rrn, od_dta_tail,
                          od_tail_len, isnullk);
 
+            if (iq->vfy_idx_track) {
+                dup_txn_insert = track_record_index(iq, ixnum, key, ixkeylen);
+            }
+
             if (vgenid && rc == IX_DUP) {
                 if (iq->usedb->ix_dupes[ixnum] || isnullk) {
                     rc = ERR_VERIFY;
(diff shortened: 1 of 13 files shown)


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>3/511 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>sc_truncate<br>sc_inserts<br>ioerror_remsql</p>

db/indices.c
```diff
@@ -423,6 +423,24 @@ int add_record_indices(struct ireq *iq, void *trans, blob_buffer_t *blobs,
                 vgenid = 0; // no need to verify again
             }
 
+            if (iq->vfy_idx_track) {
+                char * store_key;
+                store_key = pool_getablk(iq->vfy_idx_pool);
+                memcpy(store_key, &ixnum, sizeof(int));
+                memcpy(store_key + sizeof(int), key, ixkeylen);
```
riverszhang89, line 430: The table itself needs to be part of the key, I believe (can probably use iq->usedb->dbs_idx), otherwise we may report false positives from upserts that write the same key value to the same ixnum but on different tables.

morgando: Yes that's true. Will update, thanks :)

tests/upsert.test/t01_upsert.trans
```diff
@@ -106,3 +106,28 @@
...
+1 create table t1(i int primary key)
+1 begin
+1 insert into t1(i) values(1) on conflict (i) do update set i=2
+1 insert into t1(i) values(100) on conflict (i) do update set i=101
+1 insert into t1(i) values(1) on conflict (i) do update set i=2
+1 insert into t1(i) values(10) on conflict (i) do update set i=11
+1 insert into t1(i) values(10) on conflict (i) do update set i=11
+1 insert into t1(i) values(10) on conflict (i) do update set i=11
+1 commit
+1 select * from t1
```
nirbhayc, line 133: You could also add ```select value from comdb2_metrics  where name like 'verify_replays'``` to the test to ensure we are not replaying unnecessarily.

morgando: That's perfect, thanks!

nirbhayc (approved):
Okay to merge.

tests/upsert.test/t01_upsert.expected
```diff
@@ -143,3 +143,35 @@ done
...
+done
+done
+done
+done
+done
+done
+done
+done
+done
+[commit] failed with rc 2 add key constraint duplicate key 'COMDB2_PK' on table 't1' index 0
```
riverszhang89, line 173: Could we return 299 (DUPLICATE), instead?

db/osqlcomm.c
```diff
@@ -6674,6 +6674,19 @@ int osql_process_packet(struct ireq *iq, unsigned long long rqid, uuid_t uuid,
...
+                if (iq->vfy_idx_track == 1 && iq->dup_key_insert == 1) {
+                    rc = ERR_UNCOMMITABLE_TXN;
+                    reqerrstr(iq, COMDB2_CSTRT_RC_DUP, "add key constraint "
+                                                   "duplicate key '%s' on "
+                                                   "table '%s' index %d",
+                          get_keynm_from_db_idx(iq->usedb, err->ixnum),
+                          iq->usedb->tablename, err->ixnum);
+                    err->blockop_num = step;
+                    err->ixnum = 0;
+                    err->errcode = ERR_UNCOMMITABLE_TXN;
```
riverszhang89, line 6686: These error codes (`rc` and err->errcode`) do not need overwriting. I believe this would give us the correct rcode for the unique constraint violation.

morgando: I decided to use `ERR_UNCOMMITABLE_TXN` because it makes it clear that the transaction failed on a verify because it was uncommittable. Using this return code causes the following line to be printed `[ERROR] Forced VERIFY-FAIL for uncommitable blocksql transaction`. This distinguishes it from a regular duplicate key constraint violation, where the transaction may still be committable on retry. The code that checks for duplicate updates on retries also uses this return code. 

What do you think?

tests/upsert.test/t01_upsert.expected
morgando: See other comment. I could change the message to make it clear on the client end that it is an uncommittable transaction (and therefore failed verify) due to the duplicate key violation, or I could get rid of all of the uncommittable flags and just have it be a regular duplicate key violation error.

db/osqlcomm.c
riverszhang89: That makes sense. `ERR_UNCOMMITABLE_TXN` it is then!

tests/upsert.test/t01_upsert.expected
riverszhang89: This is fine!

riverszhang89 (dismissed):

nirbhayc (dismissed):

nirbhayc (approved):

PR #4130, opened 2023-12-18
Print client API type in reqlog

morgando:
The changes in this PR add the client API to the information printed about each request in the request log.

### DIFF
--- a/db/reqlog.c
+++ b/db/reqlog.c
@@ -1591,6 +1591,10 @@ static void log_header_ll(struct reqlogger *logger, struct output *out,
 
     dumpf(logger, out, " rqid %s from %s", logger->id, reqorigin(logger));
 
+    if (logger->api_type) {
+        dumpf(logger, out, " clntapi %s", logger->api_type);
+    }
+
     if (out == long_request_out && is_running) {
         /* No need to include 'rc' if the request is still running. */
         dumpf(logger, out, " **running**\n");
@@ -1939,6 +1943,7 @@ void reqlog_long_running_clnt(struct sqlclntstate *clnt)
     logger.mask = logger.event_mask | logger.dump_mask;
 
     reqlog_set_clnt(&logger, clnt);
+    reqlog_set_api_type(&logger, clnt->plugin.api_type(clnt));
     reqlog_set_origin(&logger, "%s", clnt->origin);
     reqlog_set_vreplays(&logger, clnt->verify_retries);
     reqlog_set_sql(&logger, sql);
@@ -3066,6 +3071,11 @@ inline void reqlog_set_error(struct reqlogger *logger, const char *error,
     free(logger->error);
     logger->error = strdup(error);
     logger->error_code = error_code;
+}
+
+inline void reqlog_set_api_type(struct reqlogger *logger, const char *api_type)
+{
+    if (logger) logger->api_type = api_type;
 }
 
 inline int reqlog_get_error_code(const struct reqlogger *logger)
--- a/db/sql.h
+++ b/db/sql.h
@@ -400,6 +400,7 @@ int sp_column_val(struct response_data *, int, int, void *);
 void *sp_column_ptr(struct response_data *, int, int, size_t *);
 
 typedef int(plugin_func)(struct sqlclntstate *);
+typedef const char *(api_type_func)(struct sqlclntstate *);
 typedef int(response_func)(struct sqlclntstate *, int, void *, int);
 typedef void *(replay_func)(struct sqlclntstate *, void *);
 typedef int(param_index_func)(struct sqlclntstate *, const char *, int64_t *);
@@ -470,6 +471,7 @@ struct plugin_callbacks {
     plugin_func *peer_check; /* newsql_peer_check_evbuffer */
     auth_func *get_authdata; /* newsql_get_authdata */
     override_type_func *set_timeout; /* newsql_set_timeout_sbuf */
+    api_type_func *api_type; /* newsql_api_type */
 
     /* Optional */
     void *state;
@@ -536,6 +538,7 @@ struct plugin_callbacks {
         make_plugin_callback(clnt, name, peer_check);                          \
         make_plugin_callback(clnt, name, get_authdata);                        \
         make_plugin_callback(clnt, name, set_timeout);                         \
+        make_plugin_callback(clnt, name, api_type);                            \
         make_plugin_optional_null(clnt, count);                                \
         make_plugin_optional_null(clnt, type);                                 \
         make_plugin_optional_null(clnt, int64);                                \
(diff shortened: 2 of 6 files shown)


db/sql.h
```diff
@@ -52,6 +52,12 @@ enum transaction_level {
     TRANLEVEL_SNAPISOL = 12
 };
 
+enum api_type {
+    API_TYPE_UNDEFINED = 1,
+    API_TYPE_CDB2API = 2,
+    API_TYPE_COMDB2API = 3
+};
```
morgando, line 59: I don't love having this enum because it exposes non-open-source api information to the open-source code. Alternatively each plugin could just return an opaque string, then the reqlog would blindly print this string. We wouldn't be able to add any logic on top of that (logic to do something only if the api is a specific type), but it would keep comdb2api hidden from the open-source codebase. Thoughts?

riverszhang89: String sounds better. I doubt we would have any logic on top of it, and if we do, it'll most likely be implemented as another self-contained plugin callback.

db/reqlog.c
```diff
@@ -2041,6 +2041,9 @@ void reqlog_end_request(struct reqlogger *logger, int rc, const char *callfunc,
     if (!logger->in_request)
         goto out;
 
+    if (logger->api_type) {
+        reqlog_logf(logger, REQL_INFO, "Client API=%s", logger->api_type);
```
riverszhang89, line 2045: make this a single word. will make any parsing a lot easier

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>4/511 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>sc_transactional<br>queuedb_rollover_noroll1_generated<br>ddl_no_csc2<br>sc_downgrade</p>

db/reqlog.c
```diff
@@ -3068,6 +3071,11 @@ inline void reqlog_set_error(struct reqlogger *logger, const char *error,
     logger->error_code = error_code;
 }
 
+inline void reqlog_set_api_type(struct reqlogger *logger, const char *api_type)
+{
+    if (logger) logger->api_type = strdup(api_type);
```
riverszhang89, line 3076: I think this needs to be freed. But just pointing it to the string literal is probably cleaner.

plugins/newsql/newsql.h
```diff
@@ -87,6 +87,7 @@ void newsql_effects(CDB2SQLRESPONSE *, CDB2EFFECTS *, struct sqlclntstate *);
     clnt->plugin.local_check = newsql_local_check##_##name;                    \
     clnt->plugin.peer_check = newsql_peer_check##_##name;                      \
     clnt->plugin.set_timeout = newsql_set_timeout##_##name;                    \
+    clnt->plugin.api_type = newsql_api_type##_##name;                          \
```
akshatsikarwar, line 90: Don't need this.

plugins/newsql/newsql_evbuffer.c
```diff
@@ -224,6 +224,11 @@ static int newsql_set_timeout_evbuffer(struct sqlclntstate *clnt, int timeout_ms
     return 0; /* nop */
 }
 
+static const char *newsql_api_type_evbuffer(struct sqlclntstate *clnt)
```
akshatsikarwar, line 227: Don't need this.

plugins/newsql/newsql_sbuf.c
```diff
@@ -192,6 +192,11 @@ static int newsql_set_timeout_sbuf(struct sqlclntstate *clnt, int wr_timeout_ms)
     return sbuf_set_timeout(clnt, appdata->sb, wr_timeout_ms);
 }
 
+static const char *newsql_api_type_sbuf(struct sqlclntstate *clnt)
```
akshatsikarwar, line 195: Don't need this either.

db/sqlinterfaces.c
```diff
@@ -1233,6 +1233,8 @@ static void sql_statement_done(struct sql_thread *thd, struct reqlogger *logger,
 
     thd->crtshard = 0;
 
+    reqlog_set_api_type(logger, clnt->plugin.api_type(clnt));
```
akshatsikarwar, line 1236: Should we do this earlier (`run_stmt` perhaps?) We may log a query even before it has finished.

akshatsikarwar: Try running `select sleep(10)`. It should log twice. Once after the 5s threshold. And once on completion.

plugins/newsql/newsql.c
```diff
@@ -1462,6 +1462,11 @@ static int newsql_set_timeout(struct sqlclntstate *clnt, int timeout_ms)
     return 0;
 }
 
+static const char * newsql_api_type(struct sqlclntstate *clnt)
+{
+    return "cdb2sql";
```
akshatsikarwar, line 1467: `cdb2api`

db/reqlog.c
morgando: Thanks for catching this :sweat_smile: yeah I agree that the string literal is cleaner.

db/sqlinterfaces.c
morgando: It looks like the logger that is used to print stats about a running request is different from the one that prints stats about a completed request. `run_stmt` has access to the logger that is used to print stats about a completed request. `reqlog_long_running_clnt` is the function that sets up the logger when stats get printed during a long request, so I set the api type [there](https://github.com/bloomberg/comdb2/pull/4130/files#diff-6b1a6f166109d69c771d205177e9f5c0769e9b6cdc41fb93e54f45393c6d3d76R1946). 

I also moved this line to [`query_stats_setup`](https://github.com/bloomberg/comdb2/pull/4130/files#diff-c2934cd4869e4f80bc2ffe6bcb07cca56937370c1f6fbcfce0fe3d17388423feR2782) because it makes more sense to set the client api during initial setup. I don't think that this changes any behavior though since this value doesn't appear to be printed until the request completes.

akshatsikarwar (approved):

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>7/513 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>sc_truncate<br>sc_transactional_rowlocks_generated<br>queuedb_rollover_noroll1_generated<br>queuedb_rollover<br>op_makerecord<br>dupewrite<br>ddl_no_csc2</p>

PR #4142, opened 2024-01-05
Docs: Fix some quoting & brackets in cdb2_bind_array section

adamncasey:
One line of code wasn't completed and a couple of the quotes were using a non-ascii double quote which guessing by the syntax highlighting isn't accepted by most compilers.


riverszhang89 (approved):

riverszhang89:
@adamncasey Thanks for the patch!

PR #4164, opened 2024-01-19
fix table scan in a join when inner loop is remote and a table scan 

dorinhogea:
When inner join loop is a remote table scan, make sure a Rewind (i.e. a move to first row) returns at least one row if there is a predicate for the join.  Since we pass the predicate remotely for optimized filtering, returning no rows for a move to first row will be interpreted by sqlite as an empty table and it will stop the join prematurely.
We fix this by making sure a move to first row in table scan always return a row unless the remote table is empty.
I.e 
we change the remote query from 
SELECT *, rowid FROM remtable WHERE predicate
to 
SELECT *, rowid FROM remtable WHERE predicate union SELECT * from (SELECT *, rowid FROM remtable limit 1)

This is a corner case that is best to be avoided in general, since a inner loop that does a remote table scan will perform terrible in general, with or without remote filtering.

### DIFF
--- a/db/fdb_fend.c
+++ b/db/fdb_fend.c
@@ -2622,7 +2622,8 @@ static int fdb_cursor_close(BtCursor *pCur)
 }
 
 static char *_build_run_sql_from_hint(BtCursor *pCur, Mem *m, int ncols,
-                                      int bias, int *p_sqllen, int *error)
+                                      int bias, int *p_sqllen, int *error,
+                                      int at_least_one)
 {
     fdb_cursor_t *fdbc = pCur->fdbc->impl;
     char *tableName = NULL;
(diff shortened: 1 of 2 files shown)


mponomar (approved):

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>5/511 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>logfill<br>sc_truncate<br>sc_transactional_rowlocks_generated<br>phys_rep_tiered<br>queuedb_rollover</p>

dorinhogea:
@mponomar thank you!

PR #4168, opened 2024-01-23
do not allocate these array once for each column

dorinhogea:
Trivial fix, right now we are allocating and reallocating ixschema array once for each column in a table.  We do not need to do that obviously.


mponomar (approved):

dorinhogea:
@mponomar thank you!

PR #4170, opened 2024-01-24
{174095892}:  Fix crash in datetime/blob comparison

morgando:
Comparisons between datetimes and blobs are not explicitly handled. `sqlite3BlobCompare` ends up handling these comparisons by accident. If a struct holding the datetime column's data has been initialized such that a member unused by datetimes (`n`) is zeroed, then this function returns a meaningless result. If this member has not been initialized to zero, then this function may crash [on this memcmp](https://github.com/bloomberg/comdb2/pull/4170/files#diff-4e08d24e07e4228cce60f7ef9ebee88a31beae9a46cc23d9fc3a0bccab187719R4839) when it accesses `z` (which is a garbage address if the column is a datetime).

The changes in this PR make `sqlite3BlobCompare` assume that the value of `n` is zero for all datetime columns to make the crashing case behave like the non-crashing case. In the longer term, we can consider a more appropriate comparison between blobs and datetimes or throw an error if this comparison is attempted.


roborivers (commented):
Lint check:

```diff


@@ sqlite/src/vdbeaux.c @@
+ If a new fixed-length serial type is introduced, sqlite3IsFixedLengthSerialType (defined in sqlite/vdbeInt.h) needs to be changed accordingly.
```

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>3/512 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>queuedb_rollover_noroll1_generated<br>prepare<br>truncatesc_ufid_on_generated</p>

PR #4223, opened 2024-02-16
Abort on commit if this lockid previously received deadlock

markhannum:
Simple idea from Mike .. if a lockid has gotten a DEADLOCK, and then subsequently attempts to commit, then abort the database rather than possibly corrupting it.  I've defaulted this to ON to test under RR.


markhannum:
This code causes the 'onconflict' test in that branch to core if run against a version which does not have the fix .. as expected.

PR #4238, opened 2024-02-26
Multi table ddl should have UUID

dorinhogea:
Current multi-table ddl implementation uses uuid=0 to mark a standalone table schema change. It sets uuid=bplog->uuid if more than one table schema change is done in the same transaction.
Nevertheless, bplog->uuid==0 before we call START_SOCKSQL.
While this still does not fixes multi-table ddl resume, it is a pre-requisite to identify schema changes that are part of the same transaction.

### DIFF
--- a/db/osqlsqlthr.c
+++ b/db/osqlsqlthr.c
@@ -1781,12 +1781,6 @@ int osql_schemachange_logic(struct schema_change_type *sc,
             hash_add(clnt->ddl_tables, strdup(sc->tablename));
     }
 
-    if (!bdb_attr_get(thedb->bdb_attr, BDB_ATTR_SC_RESUME_AUTOCOMMIT) ||
-        in_client_trans(clnt)) {
-        sc->rqid = osql->rqid;
-        comdb2uuidcpy(sc->uuid, osql->uuid);
-    }
-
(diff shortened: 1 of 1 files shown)


dorinhogea:
NOTE: this bug has an unfortunate consequence when master bounces during a multi-table ddl transaction; the first schema change has uuid==0 and it is committed standalone, while the rest are recognized as a multi-table and executed but not finalized (which is supposed to be done by the replicant retrying the bplog).  Unfortunately, this part also fails for a secondary bug, which is addressed in a follow-up PR (no seed and host error).

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>5/520 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>delfile_block [setup failure]<br>systable_locking<br>queuedb_rollover<br>prepare<br>sc_downgrade</p>

dorinhogea:
reopened with shorter name here: https://github.com/bloomberg/comdb2/pull/4242

PR #4319, opened 2024-03-28
Tweaking send-stat-size formatting

riverszhang89:
The `diskspace_nollmeta` test starts failing because of a deleted space. Add it back.


morgando (approved):

riverszhang89:
@morgando Thanks for the review!

PR #4338, opened 2024-04-08
{174866577}: Fixing libevent stack overflow

riverszhang89:
The following pseudo code crashes the server, when sockpool is present

```
while true
  cdb2_open()
  /* do nothing */
  cdb2_close()
end
```

The server will call rd_hdr recursively, and eventually will overflow its call stack, as shown below.

```
12 rd_hdr
13 rd_hdr
14 rd_hdr
...
32672 rd_hdr
32673 rd_hdr
32674 rd_hdr
32675 event_process_active_single_queue
32676 event_process_active
32677 event_base_loop
32678 net_dispatch
32679 start_thread
32680 clone
```

This patch fixes it: after multiple consecutive sql resets are received from the same user, the user will be rescheduled back to the event queue, instead of being processed inline.

Note that open-cdb2api is unaffected by this, as it always does a dbinfo query which will break the server out of the recursion. On the other hand, bb-cdb2api which has the skip-dbinfo optimization is prone to this bug. I have added an internal `donate_ddos` test to demonstrate the bug.

### DIFF
--- a/db/db_tunables.c
+++ b/db/db_tunables.c
@@ -482,6 +482,13 @@ int gbl_new_leader_duration = 3;
 extern int gbl_transaction_grace_period;
 
 /*
+ * RESET is received on server twice, typically:
+ * sockpool resets once, and the client API resets again.
+ * For now, we need to allow at least 2 resets.
+ */
+int gbl_max_num_consecutive_sql_resets = 2;
+
+/*
   =========================================================
   Value/Update/Verify functions for some tunables that need
   special treatment.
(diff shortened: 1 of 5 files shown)


roborivers (commented):
Lint check:

```diff


@@ db/db_tunables.h @@
+ Please update the tunable documentation (docs/pages/config/config_files.md) accordingly.
```

akshatsikarwar (changes requested):
While this works, a better fix is to remove the recursive call altogether. E.g.: https://github.com/akshatsikarwar/comdb2/commit/8cb59b9381dbcdc700790d116c23e8c18fce2639

riverszhang89:
Agreed that removal of recursion is cleaner. I did it this way because in most cases, user data is ready to read immediately, followed by a RESET; therefore scheduling it back to the event queue seems a bit of waste of cycles.

PR #4391, opened 2024-04-30
{174996173}: Workaround for remote-insert that reads from the same remote

riverszhang89:
Dorin pointed out the bug, that if a query reads from a remote table and inserts the results into another table in the same remote database, the query will hang (e.g., INSERT INTO remote.t1 SELECT * FROM remote.t2).

This patch works it around by always using a temptable plan for such a query.


riverszhang89:
This patch is a short-term fix. Dorin has a better idea for chunk+remote: we can simply pass a chunk flag to remote in the beginning of the transaction (just like how cdb2api client does a 'set transaction chunk N'), from source to remote. Remote can then start committing in chunks by itself, without the need of source telling it when to commit. I'll work on that.

dorinhogea (approved):
lgtm! thank you!

riverszhang89:
```cdb2test   Apr 30 13:48:08 2024  in-comdb2db chunk_remote.R20240430.12```

riverszhang89:
@dorinhogea Thank you for the review!

PR #4400, opened 2024-05-03
sqlwriter flush

akshatsikarwar:
`sql_flush` clears `wr_continue` flags. Intention is to prevent heartbeat from trying to flush while sql-thread is doing the same. 

Due to the bug, heartbeat is disabled altogether if `sql_trickle_cb` happens to run at the same.

After sql-thread is done flushing it goes back to running the query. If there are no more rows produced in the next 5secs, the client will time out since heartbeats are no longer being sent.


roborivers (commented):
.

PR #4492, opened 2024-06-14
Fixing debug build

riverszhang89:
@LiSodium Lina pointed it out that my previous changes break the debug build. This patch adds the missing argument.


markhannum (approved):

riverszhang89:
@markhannum @LiSodium Thank you for the review!

PR #4493, opened 2024-06-14
[8.0] Fix debug build 8

riverszhang89:
Porting from #4492


markhannum (approved):

riverszhang89:
@markhannum Thanks again for the review!

PR #4535, opened 2024-07-08
Fix ddl offset error on where clause, add support for alter add index where

chands10:
Where clause cases.
Fix this case: `CREATE TABLE t10(i INT, INDEX i(i) WHERE i > 5)$$`
Also had memory leak here ^^
Previously would get `failed with rc 240 malformed database schema `

Add support for this case: `ALTER TABLE t1 ADD INDEX idx6(i) WHERE i > 5 $$`
SQL bubble documentation already shows alter add index where command is supported
https://bloomberg.github.io/comdb2/sql.html#alter-table-ddl
Previously this would pass but would not actually change the csc2


roborivers (commented):
Lint check:

```diff


@@ sqlite/src/parse.y @@
+ Custom tokens and rules must be defined after SQLite's.
+ Please update SQL documentation and bubble diagrams to reflect the changes you've made, if necessary.
```

aakash10292 (approved):
LGTM!

chands10:
Thanks Aakash!!!

PR #4623, opened 2024-08-20
Enforce commit LSN map dependency on utxnids

morgando:
The changes in this PR block commit_lsn_map from being enabled if utxnid logging is disabled.

### DIFF
--- a/tests/commit_lsn_map.test/runit
+++ b/tests/commit_lsn_map.test/runit
@@ -259,5 +259,60 @@ function test_delete
 	done
 }
 
+verify_tunable_value() {
+	local -r tunable=$1 exp_on_off=$2 node=$3
+	local res_on_off
+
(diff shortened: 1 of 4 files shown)


tests/commit_lsn_map.test/runit
```diff
@@ -259,5 +259,47 @@ function test_delete
...
+	then
+		verify_tunable_value "commit_lsn_map" "off" && \
+		verify_tunable_value "utxnid_log" "on"
+	elif [ "$TESTCASE" == "commit_lsn_map_utxnid_disabled_then_enabled_generated" ];
+	then
+		verify_tunable_value "commit_lsn_map" "on" && \
+		verify_tunable_value "utxnid_log" "on"
+	else
+		echo "No test for $TESTCASE"
+		return 1
```
morgando, line 300: Look into rewriting tunable checks in a single test-case to avoid overhead.

PR #4691, opened 2024-09-13
Address commit LSN map recovery bug

morgando:
Transactions are added to the commit LSN map during the backwards pass of recovery. The backward pass undoes transactions. Currently, the commit LSN map code assumes that all transactions undone in the backwards pass will be redone in the forward pass; however, there is no guarantee of this. The changes in this PR address this bug as follows:

When transactions are added to the commit LSN map during the backwards pass they are also marked as "in limbo." This is a way of saying that we are still uncertain about what status these transactions will finally end up in (whether they will be committed or undone).

Each transaction committed in the forward pass is taken out of limbo since we become certain that all of these transactions will be in a committed state once we finish recovery.

After the forward pass, all transactions still in limbo are taken out of limbo and removed from the commit LSN map because they were never committed during the forward pass.

### DIFF
--- a/berkdb/build/db.h
+++ b/berkdb/build/db.h
@@ -2860,7 +2860,7 @@ struct __utxnid {
 struct __logfile_txn_list {
 	u_int32_t file_num;
 	DB_LSN highest_commit_lsn;
-	LISTC_T(struct __utxnid) commit_utxnids;
+	hash_t *commit_utxnids;
 };
 
 struct __utxnid_track {
--- a/berkdb/build/db.h
+++ b/berkdb/build/db.h
@@ -2875,6 +2875,9 @@ struct __txn_commit_map {
 	DB_LSN highest_checkpoint_lsn;
 	hash_t *transactions;
 	hash_t *logfile_lists;
+
+	hash_t *utxnids_in_limbo;
+	DB_LSN highest_commit_lsn_recovered_so_far;
 };
 
 struct __mempv_cache_page_key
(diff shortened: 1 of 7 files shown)


morgando:
If **all** committed transactions that are unwound during pass 2 (backwards pass) of recovery and are not recommitted during pass 3 (forward pass) of recovery have commit records that **are guaranteed** to test positive on this condition (https://github.com/bloomberg/comdb2/blob/e0e6d10380753a8eb0dfe2d5f353d331886eb4f5/berkdb/txn/txn_rec.c#L455) in pass 2, then we can greatly simplify this code by removing the "in limbo" logic and simply removing every commit record that hits this condition from the commit LSN map.

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>368/577 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>lostwrite [setup failure]<br>sc_inserts [setup failure]<br>timepart_constraints [setup failure]<br>epochms_rollover<br>sc_resume<br>downgrade_loop<br>incremental_backup_usenames<br>incremental_backup<br>incremental_backup_load<br>analyze_exit_immediately</p>

PR #4690, opened 2024-09-13
Allow all socksql, recom, snapisol & serial

markhannum:
Consider all socksql, read-committed, snapshot, and serializable as non-tagged requests.

PR #4692, opened 2024-09-13
Consider all socksql, read-committed, snapshot, and serializable as non-tagged requests.

markhannum:
This is the 8.0 version of https://github.com/bloomberg/comdb2/pull/4690.


dorinhogea (approved):
lgtm,  thank you!

PR #4724, opened 2024-09-25
feat: correct location of the workflow to publish cdb2hibernatedialect

MikeLaptev:
Correct location of the workflow to publish `cdb2hibernatedialect`
- adding a condition for a workflow to run when changes are in the path `contrib/cdb2hibernatedialect/**` (https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-including-paths)
  - ❗ I've removed original condition to run on push to the `main` branch, to avoid situation when the workflow could be run on non-cdb2hibernatedialect changes (https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#using-multiple-events) , but instead of add an `if` condition to prevent job to be running if `github.ref` is not `main`. (https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution)
- for tasks which are building the binary and publishing it, the working directory is added - `./contrib/cdb2hibernatedialect` (https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#defaultsrunworking-directory)

### DIFF
--- a/.github/workflows/publication_cdb2hibernatedialect.yaml
+++ b/.github/workflows/publication_cdb2hibernatedialect.yaml
@@ -1,0 +1,59 @@
+name: 'Publication cdb2hibernatedialect'
+
+on:
+  push:
+    paths:
+      - 'contrib/cdb2hibernatedialect/**'
+  release:
+    types: [published]
+
+jobs:
+  publish:
+    name: 'Publish'
+    if: ${{ github.ref == 'refs/heads/main' }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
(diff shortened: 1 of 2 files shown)


MikeLaptev:
@riverszhang89 , just want to keep this in draft for awhile, but in my previous PR, I put the publication workflows into an incorrect folder.

MikeLaptev:
@riverszhang89 @aibolit , could you please have a look?

riverszhang89 (approved):

PR #4734, opened 2024-09-27
feat: removing publishing flow for cdb2hibernatedialect

MikeLaptev:
🗒️  should be done prior to https://github.com/bloomberg/comdb2/pull/4731


MikeLaptev:
🗒️  will be part of a different cumulative PR

PR #4746, opened 2024-10-02
block partitioning a table that is the target of a constraint

dorinhogea:
Per title.

PR #4747, opened 2024-10-02
block partitioning a table that is the target of a constraint

dorinhogea:
https://github.com/bloomberg/comdb2/pull/4746 for 8.0


dorinhogea:
@chands10 thank you

PR #4768, opened 2024-10-14
Address log delete races

morgando:
There are a few places where the log deletion lock isn't held over `backend_update_sync`. The changes in this PR update the implementation of `backend_update_sync` to acquire the lock for the duration of the call instead of requiring the caller to hold the lock.


riverszhang89 (approved):

morgando:
`cdb2test   Oct 15 06:54:11 2024  in-comdb2db logdel.R20241015.4`

PR #4840, opened 2024-11-15
use the openssl EVP API for SHA1 and AES instead of the deprecated lo…

frumioj:

This is sort of a bug-fix, but also improves security

Current behaviour is that you use a `#pragma` to remove warnings about using deprecated openssl features. This fix implements non-deprecated usage of the EVP APIs of openssl.

Try building and testing with the new code. Behaviour should be unchanged (still using AES and SHA-1 for encryption and key stretching) but the build should work correctly while not using the `#pragma`.

This new behaviour is better as it is more future-proof use of openssl, and is a more secure way to use the openssl APIs.

### DIFF
--- a/berkdb/crypto/aes_method.c
+++ b/berkdb/crypto/aes_method.c
@@ -9,11 +9,6 @@
  * astubble@rice.edu.
  */
 
-/* AES_* and SHA1_* functions are made deprecated as of OpenSSL 3.0.
-   Disable compiler warning for the entire file as it's not very likely
-   that we'll ever modify this file. */
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-
 #include "db_config.h"
 
 #ifndef lint
@@ -28,6 +23,7 @@ static const char revid[] = "$Id: aes_method.c,v 1.18 2003/04/28 19:59:19 bostic
 #include "db_int.h"
 #include "dbinc/crypto.h"
 #include "dbinc/hmac.h"
+#include <openssl/evp.h>
 
 static int __aes_derivekeys __P((DB_ENV *, DB_CIPHER *, u_int8_t *, size_t));
 
@@ -83,8 +79,15 @@ __aes_close(dbenv, data)
 	DB_ENV *dbenv;
 	void *data;
 {
-	__os_free(dbenv, data);
-	return (0);
+  AES_CIPHER *aes = (AES_CIPHER *)data;
+
+  // free the EVP openssl contexts
+  if (aes) {
+    EVP_CIPHER_CTX_free(aes->encrypt_ctx);
+    EVP_CIPHER_CTX_free(aes->decrypt_ctx);
+    __os_free(dbenv, aes);
+  }
+  return (0);
 }
 
 /*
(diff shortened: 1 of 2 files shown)


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>3/588 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>sc_inserts_logicalsc_generated [setup failure]<br>snap_ha_retry_newsi_generated<br>auth</p>

berkdb/crypto/aes_method.c
```diff
@@ -144,7 +159,16 @@ __aes_encrypt(dbenv, aes_data, iv, data, data_len)
 	if ((ret = __db_generate_iv(dbenv, (uint32_t*)orig)) != 0)
 		return (ret);
 	memcpy(copy, orig, DB_IV_BYTES);
-	AES_cbc_encrypt(data, data, data_len, &aes->encrypt_key, copy, AES_ENCRYPT);
+
+	if (!EVP_EncryptInit_ex(aes->encrypt_ctx, NULL, NULL, NULL, copy))
```
riverszhang89, line 163: Would you not need a key here?

frumioj: Sorry for taking so long to go back and look at this @riverszhang89 - the key is set into the EVP_CIPHER_CTX in the __aes_derivekeys method called during the init. So passing NULL along with the aes->encrypt_ctx in the __aes_encrypt is telling EVP_EncryptInit to use the already defined key in the context. This looks correct. I hope to do better testing of this once I setup another comdb2 test environment.

berkdb/crypto/aes_method.c
```diff
@@ -174,17 +198,40 @@ __aes_derivekeys(dbenv, db_cipher, passwd, plen)
...
+	// Use Openssl EVP API for SHA1 operations
+	EVP_MD_CTX *md_ctx = EVP_MD_CTX_new();
+	uint8_t key[DB_MAC_KEY];
+	unsigned int key_len;
+	
+	if (!md_ctx)
+	  return (EINVAL);
+	
+	// Initialize EVP digest context with SHA1
+	if (!EVP_DigestInit_ex(md_ctx, EVP_sha1(), NULL) ||
```
frumioj, line 217: Noting here that by using SHA-1 we are getting 20 bytes output, which means that using AES-128 and AES_set_encrypt_key and AES_set_decrypt_key will actually truncate the hash output to 16 bytes, resulting in less security than might be possible. I think what I want to do which is a combination of simplicity but still better security is to use SHA-256 instead of SHA-1, with an output of 32 bytes which can be directly input to AES (but moves to AES-256. That might be trickier than I'm hoping...

berkdb/crypto/aes_method.c
frumioj: I've now tested the code in creating an encrypted DB and it works. I ran it through valgrind as well to ensure no (new) memory leaks. @riverszhang89

berkdb/crypto/aes_method.c
frumioj: Resolving this because I decided to try and use SHA256 and AES 256 CBC, which worked.

frumioj:
Noting that using SHA256 and AES256 will produce larger encrypted files than SHA1 + AES128, but if you wish, it is easy to switch back to AES128.

PR #4850, opened 2024-11-20
[8.0]{177357479}: Fixing an sc-resume corner case

riverszhang89:
The cluster may hang after the new master fails to resume a schema change. This patch fixes it.


markhannum (approved):

riverszhang89:
```
cdb2test   Nov 20 10:25:57 2024  in-comdb2db resume_empty_tables.R20241120.4
```

@markhannum Thanks for the review!

PR #4911, opened 2024-12-16
Fixing potential infinite loops (x3)

jvanegue:
At 3 locations, a loop is indexed using variable int ii, however the upper bound of the loop is an unsigned long long int. This means ii could wrap before reaching the upper bound. This will result in an infinite loop if this happens.

### DIFF
--- a/db/fdb_fend.c
+++ b/db/fdb_fend.c
@@ -5938,7 +5938,7 @@ int fdb_push_set(int val)
     if (val) {
         gbl_fdb_push_remote = 1;
     } else {
-        gbl_fdb_push_remote_write = 0;
+        gbl_fdb_push_remote = 0;
         /* disabling push disables also push write */
         gbl_fdb_push_remote_write = 0;
     }


morgando (approved):

mponomar:
Julien, mind pushing new commit with --amend -s for DCO?  It's required for this repo.  Thanks!

jvanegue:
Let me know if this is good now. I can create a fresh PR with -s at the first place if that didnt do what you wanted.

morgando:
It looks like the merge commit here isn't signed off which is causing validation to fail. Could you make the PR just have the one, signed off commit -- thanks! Sorry the sign off stuff is kind of a hassle.

akshatsikarwar:
> It looks like the merge commit here isn't signed off which is causing validation to fail. Could you make the PR just have the one, signed off commit

What @morgando said earlier ^^
Can you make it so there is only one single commit? Right now I see 3 and only one of them is has the sign-off. Perhaps, open a new PR  if you have trouble doing this?

PR #4922, opened 2024-12-18
Handle IX_EMPTY as row stream terminator when reading sqlite_stats in remsql.

dorinhogea:
Per title; should fix remsql in robom.


mponomar (approved):

dorinhogea:
@mponomar thank you

PR #5050, opened 2025-03-04
collect blobs during check constraints validation in sc

dorinhogea:
Right now, any schema change for a table with check constraints that requires blobs for validation (long vutf8 fields for example, used with a json function, re: 178413697) will fail the check since the blob shows empty.


riverszhang89 (dismissed):

riverszhang89 (approved):

dorinhogea:
@riverszhang89 thank you!

PR #4938, opened 2024-12-27
Allow multi-line ';'-delimited cdb2sql statements

morgando:
See tests for example use

PR #5068, opened 2025-03-13
cdb2sql: Allow sql given in argv to be ';'-delimited

morgando:
Follow up on #4938, which provided this ability for file and interactive inputs.


tests/cdb2sql.test/t03.multiline.sql
```diff
@@ -1,4 +1,2 @@
--- test complete statment spanning multiple lines that contains
--- a non-terminating semicolon
```
morgando, line 2: Needed to take the leading comments out of the test cases because when their contents are provided via argv, cdb2sql reads '-' and thinks that we want to start an interactive session.

I don't think that these comments were very useful because the test cases are pretty simple.

mponomar (approved):

morgando:
`cdb2test   Mar 13 14:26:08 2025      success mulitline_for_argv_sql.R20250313.9`

PR #5081, opened 2025-03-21
{178575361} Fill autoinc value for ireq keys that are not expressions on master

chands10:
Every key is processed on replicant if there exists even one expr index, but autoinc is only applied on master. If a normal key contains autoinc value and there is an expr index, we need to process the value for the autoinc col of the key on master

Still need to fix case where expression index containing autoinc value does not work


chands10:
cdb2test   Mar 21 03:30:39 2025      success autoincreq.R20250321.0

markhannum (approved):

PR #5130, opened 2025-04-24
179012450: Fix readonly client abort

morgando:
sql engine setup sometimes needs to create views. The `sqlclntstate` struct used to create the views is the same as the one running the sql. This leads to an abort if the client is running in read-only mode. The changes in this PR avoid this by turning off read-only mode while the views are getting created.

The changes in this PR should fix the timepart_readonly test failures.

### DIFF
--- a/db/sqlinterfaces.c
+++ b/db/sqlinterfaces.c
@@ -4291,9 +4291,12 @@ check_version:
         get_copy_rootpages_nolock(thd->sqlthd);
         if (clnt->dbtran.cursor_tran) {
             if (thedb->timepart_views) {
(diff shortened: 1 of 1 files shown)


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>1/606 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>fdb_compat_rte_connect_generated</p>

PR #5181, opened 2025-05-28
Warn in case of remsql on databases with no authentication

mohitkhullar:
Warn in 8.1

### DIFF
--- a/db/db_access.c
+++ b/db/db_access.c
@@ -78,6 +78,7 @@ int check_user_password(struct sqlclntstate *clnt)
 {
     int password_rc = 0;
     int valid_user;
+    static int remsql_warned = 0;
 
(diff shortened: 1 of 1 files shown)


dorinhogea (changes requested):
In order to capture also fdb push queries, we should checkif client requests SQLITE rows forma. The is_remsql flag marks only sql that is running distributed on both coordinator and participant, as opposed to simply forwarded to participant (as in fdb push).

mohitkhullar:
> In order to capture also fdb push queries, we should checkif client requests SQLITE rows forma. The is_remsql flag marks only sql that is running distributed on both coordinator and participant, as opposed to simply forwarded to participant (as in fdb push).

Added the check for sqlite format

dorinhogea (approved):
thank you

PR #5184, opened 2025-05-30
cdb2api does not send or retry abbreviated hint queries

dorinhogea:
Sql hint queries that do not have full statement are only sent by legacy comdbapi sql.  For newsql interface, do not try to retry, since the api will not handle it.

RE: 177199209


dorinhogea:
@riverszhang89 thank you

PR #5187, opened 2025-05-30
Add error messages in the missing places where we return SQLITE_INTERNAL

dorinhogea:
per title

re: 179431713


dorinhogea:
@mponomar thank you

PR #5207, opened 2025-06-09
fix crashes in S/GET_CURSOR_RECORDING 

dorinhogea:
Fail selectv that requires too many cursors; do not try to read or write beyond the recording array.

re: 179191425


mponomar (approved):

dorinhogea:
@mponomar thank you

PR #5214, opened 2025-06-10
{179028562} Limit queue depth scan for comdb2_queues

mponomar:


tests/queue_scan_limit.test/out
```diff
@@ -0,0 +1,2 @@
+(queuename='__qno_odh', spname='no_odh', head_age=1, depth=10000, total_enqueued=40000, total_dequeued=0)
```
akshatsikarwar: added accidentally?

plugins/dbqueuedb/dbqueuedb.c
```diff
@@ -537,6 +538,8 @@ static int stat_callback(int consumern, size_t length,
         }
         stats[consumern].has_stuff = 1;
         stats[consumern].depth++;
+        if (stats[consumern].depth == gbl_nonodh_queue_scan_limit)
```
akshatsikarwar: Make it so `0` disables?

mponomar: It should already - if it's 0 it'll never match that condition.  I'll make it more explicit in the code.

akshatsikarwar (approved):

PR #5237, opened 2025-06-23
Add dbname when available for fdb query trace

mohitkhullar:
or fdb_push when query is proxied


mohitkhullar:
Remote sql being used on database with authentication disabled, please enable IAM on this database. sourcedb:mohitdb1
///
Remote sql being used on database with authentication disabled, please enable IAM on this database. sourcedb:fdb_push

PR #5268, opened 2025-07-09
Add bulk import tests

morgando:
- Test dropping src table while we're copying files to dst db
- Test running election on src db while we're copying files to dst db


tests/bulkimport.test/runit
```diff
@@ -380,6 +403,116 @@ function test_resume_is_blocked() {
 	)
 }
 
+function test_src_db_table_dropped_during_import() {
```
morgando, line 406: * table is dropped before src db sends any files -> dst db will not see src table and import will fail. This is tested in `test_source_table_does_not_exist`
* table is dropped after src db sends all files -> not worth testing: from the perspective of the dst db this is equivalent to the drop not happening
* table is dropped while src db sends files -> tested in `test_src_db_table_dropped_during_import`. `comdb2_files` sends transaction log files after sending all btree files. So if the drop happens after btree files have been sent, then the import will error if the txn log reflecting the drop has been flushed to disk by the time it is sent to the src db and succeed if it hasn't (because the dst db brings up a copy of the src db with the files that it gets and runs recovery on it). Either of these outcomes is fine.

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>10/615 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>disttxn [setup failure]<br>cldeadlock<br>unionpar_maxqueue<br>reco-ddlk-sql<br>analyze<br>sc_lotsoftables_logicalsc_generated<br>sc_lotsoftables<br>insert_lots_ssl_generated<br>insert_lots_large_tran_generated</p>

PR #5289, opened 2025-07-28
fix crash during group by sorting involving datetimes and strings

dorinhogea:
We need to pass the proper timezone  to Mem objects that are processed by the sorter engine.

It is possible for a query to pass values that require a datetime conversion to string, which in turn requires a timezone.

Example:
select case when d <= '2020-10-01T' then "2020 and change" else d end as mixes from t group by mixes

re: 180106112

### DIFF
--- a/sqlite/src/vdbeaux.c
+++ b/sqlite/src/vdbeaux.c
@@ -5145,6 +5145,9 @@ int sqlite3VdbeRecordCompareWithSkip(
 
 #if defined(SQLITE_BUILDING_FOR_COMDB2)
   memset(&mem1, 0, sizeof(Mem));
+  /* pass tzname to Mem1 as well, in case it is a datetime that needs conversion */
+  mem1.tz = pRhs->tz;
+  mem1.enc = pPKey2->pKeyInfo->enc;
 #endif /* defined(SQLITE_BUILDING_FOR_COMDB2) */
   /* If bSkip is true, then the caller has already determined that the first
   ** two elements in the keys are equal. Fix the various stack variables so
--- a/sqlite/src/vdbemem.c
+++ b/sqlite/src/vdbemem.c
@@ -482,8 +482,9 @@ int sqlite3VdbeMemStringify(Mem *pMem, u8 enc, u8 bForce){
 
     if( convMem2ClientDatetimeStr(pMem, tmp, sizeof(tmp), &outdtsz) ){ 
       char *z;
-      sqlite3ErrorWithMsg(pMem->db, SQLITE_CONV_ERROR,
-            "can't convert datetime value to string");
+      if (pMem->db)
+          sqlite3ErrorWithMsg(pMem->db, SQLITE_CONV_ERROR,
+                              "can't convert datetime value to string");
       pMem->n = strlen("conv_error");
       z = sqlite3GlobalConfig.m.xMalloc(pMem->n+2);
       if( !z ) return SQLITE_NOMEM;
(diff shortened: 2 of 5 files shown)


sqlite/src/vdbemem.c
```diff
@@ -482,8 +482,9 @@ int sqlite3VdbeMemStringify(Mem *pMem, u8 enc, u8 bForce){
 
     if( convMem2ClientDatetimeStr(pMem, tmp, sizeof(tmp), &outdtsz) ){ 
       char *z;
-      sqlite3ErrorWithMsg(pMem->db, SQLITE_CONV_ERROR,
-            "can't convert datetime value to string");
+      if (pMem->db)
+          sqlite3ErrorWithMsg(pMem->db, SQLITE_CONV_ERROR,
+                              "can't convert datetime value to string");
```
morgando, line 487: This change looks great to me!

sqlite/src/vdbeaux.c
```diff
@@ -5145,6 +5145,9 @@ int sqlite3VdbeRecordCompareWithSkip(
 
 #if defined(SQLITE_BUILDING_FOR_COMDB2)
   memset(&mem1, 0, sizeof(Mem));
+  /* pass tzname to Mem1 as well, in case it is a datetime that needs conversion */
+  mem1.tz = pRhs->tz;
+  mem1.enc = pPKey2->pKeyInfo->enc;
```
morgando, line 5150: I'm confused about the purpose of the changes to this file and to `sqlite/src/vdbesort.c`. If my understanding is correct, you added this so that non-datetime strings directly convert to datetime strings?

I don't think that we define any conversion rules going from non-datetime strings to datetime strings (as occurs in the test case). If you try to insert '2020 and change' into a datetime column it will say that it is 'not compatible'. Therefore I think it is right to expect either an error or undefined behavior when a conversion like this happens.

also, with your changes, I think that the test query without the `group by` still returns unintuitive results sometimes:

```
cdb2sql> select case when d > '2020-10-01T' then "2020 and change" else d end as mixes from t
(mixes="2007-10-01T000000.000 US/Eastern")
(mixes="2007-10-01T000000.000 US/Eastern")
```

I think that maybe instead of making these changes we should stress to clients not to mix incompatible values in this way and maybe alarm when it happens. If enough people want well-defined behavior here then maybe we can have a bigger project to define and implement that behavior.

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>5/617 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>disttxn [setup failure]<br>insert_lots_large_tran_generated<br>phys_rep_tiered_nosource_generated<br>unionpar_maxqueue<br>sc_lotsoftables_logicalsc_generated</p>

sqlite/src/vdbeaux.c
dorinhogea: This is to convert a datetime to a string.

dorinhogea: Irrespective of the confusion, the server should not crash. In addition, the sorter does not receives timezone information, and there are plenty of cases when this is valid usage.  Last but not the least, the client statement makes sense.

morgando: Okay now my understanding is that the changes from these two files are to fix an assertion failure occurring in the debug build because timezone and encoding are not set inside of the sorting engine and also to allow conversion from datetime to string, which should work because string is a more general type than datetime. That makes sense.

morgando (approved):
LGTM, although clients shouldn't rely on the results of queries like the test query unless we're 100% sure that sqlite always tries to convert the datetime to a string and not the other way around

dorinhogea:
@morgando thank you

PR #5302, opened 2025-08-04
Fix clang build warning

morgando:
```
comdb2/sqlite/src/comdb2lua.c:172:17: error: variable length array folded to constant array as an extension [-Werror,-Wgnu-folding-constant]
  172 |     char spname[MAX_SPNAME_FOR_TRIGGER];
      |                 ^~~~~~~~~~~~~~~~~~~~~~
comdb2/sqlite/src/comdb2lua.c:27:43: note: expanded from macro 'MAX_SPNAME_FOR_TRIGGER'
   27 | #define MAX_SPNAME_FOR_TRIGGER MAX_SPNAME-strlen(Q_TAG) // includes null terminator
```

The changes in this PR fix this warning by using `sizeof` instead of `strlen`. This fixes the error because `sizeof` is evaluated at compile time.

### DIFF
--- a/sqlite/src/comdb2lua.c
+++ b/sqlite/src/comdb2lua.c
@@ -24,7 +24,7 @@ int bdb_get_sp_get_default_version(const char *, int *);
 extern int gbl_create_default_consumer_atomically;
 extern int gbl_sc_protobuf;
 
-#define MAX_SPNAME_FOR_TRIGGER MAX_SPNAME-strlen(Q_TAG) // includes null terminator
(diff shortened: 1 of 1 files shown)


mponomar (dismissed):

chands10 (changes requested):
Wait should there be parentheses around this? `(sizeof(Q_TAG)-1)`

So `(MAX_SPNAME-(sizeof(Q_TAG)-1))`

chands10 (approved):

PR #5306, opened 2025-08-05
Broadcast newmaster from watcher-thread

markhannum:
Fixes a case where incoherent physical replicants request log-records on a lower generation than the current master's generation.


akshatsikarwar:
Does this need more work, or ready to merge?

PR #5314, opened 2025-08-06
Fix flakiness in catchup.test

morgando:
The catchup test restarts a node but doesn't wait until it is actually available before exiting the test. Therefore, the test sometimes fails with 'db not available at finish'.

Example from recent execution:

Test ends at 00:50:17
```
00:50:17> Success
```

Node doesn't become available until 00:50:47
```
2025/08/05 00:50:47 I AM READY.
```


tests/catchup.test/runit
```diff
@@ -51,6 +51,10 @@ while :; do
     sleep 1
 done
 
+# wait for the node to actually become available (which may happen after it becomes coherent)
+# if we don't do this, then the test will fail with 'db unavailable at finish'.
```
morgando, line 56: Double check that it isn't a bug that we're marking nodes coherent before they are available

mponomar: Thanks Morgan - it's a bug in the test - db becomes coherent before it's ready - we check for that at the beginning of all tests, but not at the exit of this one, which purposely brings a db down and then back up.  Thanks for fixing my bugs.

PR #5324, opened 2025-08-08
fix: validate table names on database creation

SChakravorti21:
When creating a database (`comdb2 --create ...`), table names are not validated in the same way as they are when tables are created via DDL (`CREATE TABLE ...`). This causes silent failures, such as the table name being truncated and the database being non-recoverable. Run the same validations on table names during database creation and reuse the validation logic across database creation and DDL for consistency.

This fixes `{180259286}`.


tests/sc_validate_table_name.test/t1_illegal_characters.expected
```diff
@@ -0,0 +1 @@
+[create table "table-with>illegal<chars" (i int)] failed with rc -3 Table name contains illegal characters. name='table-with>illegal<chars'
```
morgando, line 1: This is similar to [sql.test/t16_01.req](https://github.com/morgando/comdb2/blob/main/tests/sql.test/t16_01.req). I think you might need to update the expected output for that test since the error message changed (or you could remove the table name tests from sql.test/t16_01.req so that they're all here).

SChakravorti21: Ah! Thanks for pointing that out, I was having a hard time figuring out why `sql.test` was failing. Let me take a look at it, I think having all the table creation tests in one place makes sense.

SChakravorti21: Didn't find too many duplicate tests so I've just updated the expected error messages in `sql` for now.

db/comdb2.c
```diff
@@ -2875,6 +2876,17 @@ static int db_finalize_and_sanity_checks(struct dbenv *dbenv)
                    dbenv->envname, ii);
         }
 
+        const char *error = NULL;
+        if (is_valid_table_name(db->tablename, strlen(db->tablename),
+                                /*check_for_illegal_chars=*/1, &error) != 0) {
+          have_bad_schema = 1;
+          logmsg(LOGMSG_FATAL, "%s\n", error);
```
morgando: nit: Maybe you should check that error is not null before trying to print it (put it in the block with `free(error)` right below)?

schemachange/sc_util.c
```diff
@@ -182,6 +184,35 @@ int validate_ix_names(struct dbtable *db)
     return rc;
 }
 
+int is_valid_table_name(const char *const name, size_t len,
+                        int check_for_illegal_chars, const char **error) {
+  if (len == 0) {
+    if (error) {
+      *error = comdb2_asprintf("Table name cannot be empty", len, name);
```
morgando: Too many args passed to `comdb2_asprintf`?

schemachange/sc_util.c
```diff
@@ -182,6 +184,35 @@ int validate_ix_names(struct dbtable *db)
...
+    }
+    return 1;
+  }
+
+  if (len > MAXTABLELEN - 1) {
+    if (error) {
+      *error =
+          comdb2_asprintf("Table name is too long. max_length=%d length=%zu",
+                          MAXTABLELEN - 1, len, name);
```
morgando: Too many args passed to `comdb2_asprintf`?

schemachange/sc_util.c
```diff
@@ -182,6 +184,35 @@ int validate_ix_names(struct dbtable *db)
...
+    }
+    return 2;
+  }
+
+  if (check_for_illegal_chars &&
+      !str_is_alphanumeric(name, NON_ALPHANUM_CHARS_ALLOWED_IN_TABLENAME)) {
+    if (error) {
+      *error = comdb2_asprintf("Table name contains illegal characters", name);
```
morgando: Too many args passed to `comdb2_asprintf`?

tests/README.md
```diff
@@ -130,7 +131,7 @@ Variable | Description
 --- | ---
 `TESTID`        | Randomly generated; all output will go into `test_${TESTID}`
 `TESTDIR`       | `Path/to/source/tests/test_${TESTID}`<br>You can have the db run in a directory of your choosing by passing `TESTDIR=/tmp/somedirfortest` to `make`, ex: <br>`make testname TESTCASE=/tmp/somedirfortest`
-`TESTCASE`      | Name of the testcase currently running.  You can assume `${TESTDIR}/${TESTCASE}.test/ is where your files are.`
+`TESTCASE`      | Name of the testcase currently running.  You can assume `${TESTDIR}/${TESTCASE}.test/` is where your files are.
```
morgando, line 134: Thanks for fixing these docs :)

db/comdb2.h
```diff
@@ -3384,8 +3384,6 @@ char *get_full_filename(char *path, int pathlen, enum dirtype type, char *name,
                         ...);
 int query_limit_cmd(char *line, int llen, int toff);
 
-int is_valid_tablename(char *tbl);
-
```
morgando, line 3388: Nice cleanup 🧹

tests/sql.test/t11_01.req.exp
```diff
@@ -1,4 +1,4 @@
 [GRANT READ ON 't1' TO 'default'] failed with rc -3 Table not found
 [GRANT READ ON 't123456789012345678901234567890' TO 'default'] failed with rc -3 Table not found
-[GRANT READ ON t1234567890123456789012345678901 TO 'default'] failed with rc -3 Table name is too long
-[GRANT READ ON 't1234567890123456789012345678901' TO 'default'] failed with rc -3 Table name is too long
+[GRANT READ ON t1234567890123456789012345678901 TO 'default'] failed with rc -3 Table name is too long. max_length=31 length=32
+[GRANT READ ON 't1234567890123456789012345678901' TO 'default'] failed with rc -3 Table name is too long. max_length=31 length=32
```
morgando: This is a better error message :)

db/comdb2.c
```diff
@@ -2875,6 +2876,17 @@ static int db_finalize_and_sanity_checks(struct dbenv *dbenv)
                    dbenv->envname, ii);
         }
 
+        const char *error = NULL;
+        if (is_valid_table_name(db->tablename, strlen(db->tablename),
```
morgando: when I use a function with a name like `is_<condition>` in C then I would expect 0 to mean that the condition was not satisfied and non-0 to mean that it was satisfied (I follow this convention in my own code). I think that C standard libraries [also follow this convention](https://pubs.opengroup.org/onlinepubs/9699919799/functions/isspace.html). I couldn't find a great source discussing this convention but it looks like there is some consensus that it is a thing [in this stackoverflow thread](https://stackoverflow.com/questions/559061/should-i-return-true-false-values-from-a-c-function).

To make this more conventional I think that you could either flip the return codes (so that 1 is returned if the table name is valid and 0 is returned if it is not), or you could rename the function `validate_table_name` and then have a return code of 0 mean 'validation success' and non-0 mean 'validation failure'.

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>1/623 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>sc_lotsoftables_logicalsc_generated</p>

tests/sc_validate_table_name.test/t0_long_name.req
```diff
@@ -0,0 +1 @@
+create table t2345678901234567890123456789012 (i int)
```
morgando: I like the names of your test files! I often give them very undescriptive names like t00.req

schemachange/sc_util.h
```diff
@@ -17,6 +17,12 @@
 #ifndef INCLUDE_SC_UTIL_H
 #define INCLUDE_SC_UTIL_H
 
+#include <stddef.h>
+
+struct dbtable;
+struct tran_tag;
+typedef struct tran_tag tran_type;
```
morgando, line 24: nit: do we need these?

morgando (changes requested):

schemachange/sc_util.c
SChakravorti21: Oops, forgot to remove them after updating the error message. Thanks for catching!

Edit: sorry for seesawing, I'll update the error message instead. It'll be a little redundant in some cases but clearer in others.

schemachange/sc_util.h
SChakravorti21: I was getting warnings from clangd since this header doesn't include anything that declares these structs. I think we'd need to `#include __bdb_api_h__` for these, but I'm not sure if we wanted to do that (seems like it would pull in more than necessary). Adding forward declarations seemed least intrusive. I'm not sure if we have any preference on how to deal with these warnings though... I'm fine with removing these either way, there's probably a better solution.

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>2/623 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>pmux_sqlite_memory_generated<br>pmux_sqlite_file_generated</p>

morgando:
robomark passed - `cdb2test   Aug 12 10:38:05 2025      success long-table-name.R20250812.0`

PR #5325, opened 2025-08-08
Fix test flakiness from downgrade timing

morgando:
sc_lotsoftables and sc_truncate are frequently failing because they don't wait for all downgrades to complete before exiting, and therefore the test infrastructure can't reach the db at finish.

The changes in this PR aim to provide a downgrade function that waits until election completes and verifies that all nodes are coherent and reachable before returning.

### DIFF
--- a/tests/tools/cluster_utils.sh
+++ b/tests/tools/cluster_utils.sh
@@ -258,3 +258,102 @@ function wait_for_cluster
         waitmach ${node}
     done
 }
+
+get_node_gen() {
+    local -r node=${1}
+    local gen
+
+    set -o pipefail
+
+    gen=$(${CDB2SQL_EXE} --host ${node} ${CDB2_OPTIONS} ${DBNAME} default \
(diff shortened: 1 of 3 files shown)


tests/tools/cluster_utils.sh
```diff
@@ -258,3 +258,84 @@ function wait_for_cluster
...
+            next_gen=${gen}
+        fi
+    done
+
+    while ! get_master_and_fail_if_not_found; do
+        sleep ${downgrade_wait_seconds}
+    done
+
+    # Make sure everyone's available before we return
+    wait_for_cluster
```
morgando, line 358: also need to make sure truncates have finished?

morgando: Did this by waiting for all nodes to get marked coherent (https://github.com/bloomberg/comdb2/pull/5325/files#diff-94bda2736b1475886edc4529447d7c649804d1e4389e54be993003d9cd62a962R346-R351)

markhannum (approved):

PR #5341, opened 2025-08-15
ci: check that project compiles with GCC and Clang

SChakravorti21:
/skipbuild

8f5cdab inadvertently broke Clang builds, which was subsequently fixed in 7b1c452. Some developers on the team use Clang as it provides better diagnostics than GCC. To avoid disrupting developers who use Clang, it would be good to proactively check that project changes compile with Clang. Ensure this by compiling the project with Clang in CI. For completeness, do the same for GCC.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,117 @@
+name: Build
+
+on:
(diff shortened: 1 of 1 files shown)


morgando:
Love this!!!

PR #5343, opened 2025-08-18
Copy recent oplog entries after truncate

markhannum:
Truncating comdb2_oplog will copy at least comdb2_oplog_copy_entries into the new comdb2_oplog table.


markhannum:
Rivers had a better idea for the design - implemented here:
https://github.com/bloomberg/comdb2/pull/5347

PR #5349, opened 2025-08-20
ci: do not run workflow on pushes to `main`

SChakravorti21:
/skipbuild

In the event that a developer is using a personal access token that does not have `workflow` permissions, pushing to `main` on the developer's fork will fail. Since we run the workflow on PRs, additionally checking commits to `main` seems redundant anyway. Do not run checks on pushes to main to avoid disrupting developers.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,15 +1,12 @@
 name: Build
 
 on:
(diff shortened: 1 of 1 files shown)


SChakravorti21:
ugh, patch didn't get created on my own fork, will make a new PR

PR #5351, opened 2025-08-20
CI: Add Clang-Format Checking to Github Actions

morgando:
Adds clang-format checking to github actions using the formatting rules listed [here](https://bloomberg.github.io/comdb2/contrib.html#indentation-etc).

See example failure [here](https://github.com/bloomberg/comdb2/actions/runs/17138600486/job/48620676123?pr=5356)

These changes include a format script in the top level directory that can format-check and reformat the repo. ([`usage`](https://github.com/bloomberg/comdb2/pull/5351/files#diff-e904c9ccfa425ff0b055d2c533462314d35a529b055e8abe41d49bb46d827427R46)).

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,0 +1,47 @@
+name: clang-format
+
+on:
+  pull_request: { branches: ["main"] }
+
+# If someone pushes a new commit to a branch, ensure that only the latest commit
+# is built to avoid wasting runner minutes.
+concurrency:
+  group: ${{ github.workflow }}-pr-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+
+jobs:
+  clang-format:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    env:
+      CLANG_FORMAT: "clang-format"
+    defaults:
+      run:
+        # Make sure to fail any "run" step early on so that steps don't silently
+        # pass even when errors may have occurred.
+        shell: bash -o errexit -o nounset -o pipefail {0}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with: { fetch-depth: 0 }
+
+      - name: Install clang-format
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y clang-format
+          $CLANG_FORMAT --version
+
+      - name: Run format check
+        run: |
+          base_repo="${{ github.event.pull_request.base.repo.full_name || github.repository }}"
+          base_url="https://github.com/${base_repo}.git"
+
+          # Fetch the base branch into a unique namespace
+          git fetch --no-tags --prune --depth=1 "${base_url}" \
+            "refs/heads/${GITHUB_BASE_REF}:refs/remotes/__base__/${GITHUB_BASE_REF}"
+
+          # Compute merge-base vs the fetched ref
+          base_commit="$(git merge-base HEAD "refs/remotes/__base__/${GITHUB_BASE_REF}")"
+
+          ./ci/scripts/format --verbose --dry-run --compare-base "${base_commit}"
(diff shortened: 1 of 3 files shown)


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: <b>Error</b>. &#x26A0;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>1/625 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>maxtable</p>

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,51 @@
+name: clang-format (check)
+
+on:
+  push: { branches: ["main"] }
+  pull_request: { branches: ["main"] }
```
morgando, line 4: I think I can change this to only run the job if files in the relevant directories with the relevant extensions are modified to save runner minutes (it takes 1 min just to install clang-format). I didn't do this to avoid duplicating the list of extensions / directories. I don't feel very strongly one way or the other though.

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,49 @@
+name: clang-format (check)
+
+on:
+  pull_request: { branches: ["main"] }
```
morgando, line 4: I didn't run this check on pushes to main: We aren't very strict about this check so I feel like we maybe don't need to track it that carefully.

.github/workflows/clang-format.yml
SChakravorti21: I like it this way, otherwise we need to keep the list of directories in sync and it might be more fragile. I'm not sure we need to be super concerned with runner minutes at this time.

.github/workflows/clang-format.yml
SChakravorti21: I think that's a good call for other reasons as well, see https://github.com/bloomberg/comdb2/pull/5350.

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,46 @@
...
+      CLANG_FORMAT: "clang-format"
+    defaults:
+      run:
+        # Make sure to fail any "run" step early on so that steps don't silently
+        # pass even when errors may have occurred.
+        shell: bash -o errexit -o nounset -o pipefail {0}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with: { fetch-depth: 0 }
```
SChakravorti21, line 27: neat! maybe we could use this for the build check as well

format
```diff
@@ -0,0 +1,188 @@
...
+  local -r extract_start_and_count_regex='s/.*\+([0-9]+)(,([0-9]+))?.*/\1 \3/'
+
+  git diff -U0 ${compare_base} -- "${file}" \
+    | grep -oP "${diff_range_regex}" \
+    | sed -E "${extract_start_and_count_regex}" \
+    | while read start count; do
+        if [ -z "${count}" ]; then count=1; fi
+        end=$((start+count-1))
+        echo "-lines=${start}:${end}"
+      done | xargs
```
SChakravorti21: Only diffing modified lines is very desirable. In case it helps simplify the logic, have you considered using `clang-format-diff`? It's able to take a git diff as input and format only the changes in the diff. As an example, this is how I've been using it to format my changes locally:

```bash
find bbinc bdb cdb2api comdb2rle csc2 db net schemachange sockpool tools/pmux tools/cdb2sql tools/cdb2sockpool -name '*.c' -o -name '*.cpp' -o -name '*.h' \
| xargs git diff main...`git rev-parse --abbrev-ref HEAD` -U0 --no-color \
| grep -v "^\-" \
| clang-format-diff -p1 -i
```

SChakravorti21:
hm, wonder why build checks didn't run on the latest commit 🤔

format
morgando: Using clang-format-diff seems much better. Going to update to use that!

PR #5361, opened 2025-08-24
Fixing client invalid response type error

riverszhang89:
Morgan demonstrated that if SSL is enabled, when the server experiences a high connection rate and is slow to complete an SSL handshake, a client may end up reading an incomplete response type 369296128 (0x16030300), which is the header of an OpenSSL's hello packet.

This patch ensures that a socket is closed if its SSL handshake fails.


akshatsikarwar:
Does this need porting to bb-cdb2api?

akshatsikarwar:
Can you rebase and resubmit to robots?

PR #5364, opened 2025-08-25
Fix crashes if transaction goes in wrong state

mohitkhullar:
Abort transaction if it goes in bad state


mohitkhullar:
subsequent queries in the handle/socket --  coming from server side
[select 1] failed with rc -8 sqlinterfaces: wrong sql handle state

db/sqlinterfaces.c
```diff
@@ -1423,6 +1423,7 @@ void sql_set_sqlengine_state(struct sqlclntstate *clnt, char *file, int line,
                newstate);
 
     if (newstate == SQLENG_WRONG_STATE) {
+        handle_sql_intrans_unrecoverable_error(clnt);
```
akshatsikarwar: Trace first before calling `handle_sql_intrans_unrecoverable_error`?

mohitkhullar: done

akshatsikarwar (approved):

PR #5376, opened 2025-09-08
When mostly idle, periodically ask for missed records

mponomar:


mponomar:
cdb2test   Sep  8 11:46:34 2025      success fix-incoherent-idle.R20250908.1

PR #5378, opened 2025-09-09
cdb2_skipws

chands10:
Migration from bb


cdb2api/cdb2api.c
```diff
@@ -494,6 +487,10 @@ static void do_init_once(void)
     }
 }
 
+#define cdb2_skipws(str)                                                                                               \
+    while (*str && isspace(*str))                                                                                      \
```
dorinhogea: maybe wrap the str in ()?

chands10: done!

cdb2api/cdb2api.c
```diff
@@ -494,6 +487,10 @@ static void do_init_once(void)
     }
 }
 
+#define cdb2_skipws(str)                                                                                               \
```
riverszhang89, line 490: I think it's a common practice to wrap it in a `do {} while (0)` this macro so it can be executed in a if branch without `{}`.

riverszhang89 (approved):

chands10:
Thanks Rivers!!

PR #5417, opened 2025-09-19
`{179630312}`: Add tunable to force reloading schemas after recovery

SChakravorti21:
Add a debug tunable to help reproduce a deadlock between clusterinfo queries and recovery.

Also, fixing the deadlock uncovered a different latent bug with tag API requests accessing schema information without holding the schema lock, resulting in a segfault if recovery is simultaneously reloading schema information. Fixing this systematically will require more effort. For now, avoid the segfault when attempting to reproduce the original issue.


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p>

PR #5441, opened 2025-09-29
Warn if running serializable transactions without enable_serial_isolation tunable

morgando:
Right now, clients can run serializable transactions just by setting `enable_snapshot_isolation`, even if they never set `enable_serial_isolation`. That works because `enable_snapshot_isolation` has always turned on logical logging under the hood, which serializable mode depends on.  

But going forward, we want to decouple those two: snapshot isolation won’t require logical logging anymore. To get there safely, we need all clients that run serializable transactions to switch over to `enable_serial_isolation`. That way, we can stop tying logical logging to `enable_snapshot_isolation` without breaking serializable workloads—since by then, anyone who needs it will be explicitly using `enable_serial_isolation`.


morgando:
Changes between first and second test run just add CDB2_OPTIONS to cdb2sql calls in test: https://github.com/bloomberg/comdb2/compare/882708bf91df810ff22c15213075ab9d4a99e36c..53c3e83f6ce776f1118fb24731ab49a8e6c5d8ec

chands10 (approved):

morgando:
cdb2test   Sep 29 12:26:26 2025      success serial_warning.R20250929.2

PR #5444, opened 2025-09-30
enable_snapshot => enable_serial in serialstep test

morgando:
/adhoc-test serialstep

This PR updates a test case to comply with the new requirement that all clients running serializable transactions must set `enable_serial_isolation`. As a result, a few outputs changed because `selectv`’s range-checking implementation kicks in when `enable_serial_isolation` is used. These differences are noted in the PR comments.

### DIFF
--- a/tests/serialstep.test/s16_01.req.exp
+++ b/tests/serialstep.test/s16_01.req.exp
@@ -10,6 +10,7 @@ done
 done
 (rows updated=1)
 done
+[commit] failed with rc -103 selectv constraints
 done
 done
 (id=1, b1=x'11')
@@ -25,5 +26,5 @@ done
 done
 (rows updated=1)
 done
-[commit] failed with rc -103 constraints error, no genid
+[commit] failed with rc -103 selectv constraints
 done
--- a/tests/serialstep.test/s2_01.req.exp
+++ b/tests/serialstep.test/s2_01.req.exp
@@ -37,5 +37,6 @@ done
 done
 [commit] failed with rc 230 transaction is not serializable
 done
+[commit] failed with rc 230 transaction is not serializable
 done
 done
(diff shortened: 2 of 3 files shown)


tests/serialstep.test/s2_01.req.exp
```diff
@@ -37,5 +37,6 @@ done
 done
 [commit] failed with rc 230 transaction is not serializable
 done
+[commit] failed with rc 230 transaction is not serializable
```
morgando, line 40: [This transaction](https://github.com/bloomberg/comdb2/blob/31663cee00488f9e2a19d11bb179ac3278ce8700/tests/serialstep.test/s2_01.req#L17) is serializable in theory, but our implementation is conservative and will sometimes abort transactions that could’ve passed.

tests/serialstep.test/s16_01.req.exp
```diff
@@ -10,6 +10,7 @@ done
 done
 (rows updated=1)
 done
+[commit] failed with rc -103 selectv constraints
```
morgando, line 13: Look into why this didn't previously produce a serializable error.

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Error &#x26A0;.<br>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p>

tests/serialstep.test/s16_01.req.exp
morgando, line 13: [This commit](https://github.com/bloomberg/comdb2/blob/31663cee00488f9e2a19d11bb179ac3278ce8700/tests/serialstep.test/s16_01.req#L8) should’ve failed all along under our [`selectv` specification](https://bloomberg.github.io/comdb2/sql.html#select-syntax-overview). The fact that it didn’t before points to a bug in the selectv genid-checking implementation (which is used outside serializable). With serializable on, we hit the alternative range-checking implementation instead, which does not have the bug.

markhannum (approved):

PR #5458, opened 2025-10-03
{181005193}: Fix appsock count from revconn

akshatsikarwar:
/silent


akshatsikarwar:
Continuing here: https://github.com/bloomberg/comdb2/pull/5459

PR #5475, opened 2025-10-08
Dead code: del 'newsi' snapshot impl

morgando:
Cleaned up the 'newsi' snapshot implementation. This was a conservative pass — I only removed code that was clearly gated behind newsi tunables. There’s probably some leftover newsi logic that isn’t behind a tunable and could be cleaned up later, but this should cover most of it.

The first snapshot implementation will be deleted in a separate PR.


morgando:
^ Tests [this change](https://github.com/bloomberg/comdb2/compare/2173a4f29d5ec48ce34164a24ec00f3eeb35c91b..f51444a03dfaf84b08f945042a49cd707489c502)

PR #5480, opened 2025-10-10
CI: Fix test infra core-dump logic

morgando:
Cores were not consistently collected in our CI system due to two separate issues:
1. Core naming issue:
   When a thread (rather than the main process) crashes, the core file’s `%e` field uses the thread’s name instead of `comdb2`.  
   Our `find_cores` script assumes `%e` is always `comdb2`, so it misses those thread cores.

2. Container core limits:  
   Test containers are not configured to have a nonzero core dump limit. 
   On our infra, it appears that the m1–m3 containers operate with `ulimit -c = 0`, while the c1 container operates with `ulimit -c = unlimited`.  
   As a result, cores are collected for standalone DBs (running on c1) but not for clustered DBs (running on m1–m3).

This PR addresses both problems:
1. Updates `find_cores` to treat `%e` as a wildcard and filter cores just by PID instead of executable name + PID.  (see 79582fc461aa54a323b560e15f89fa98d2c90b0f)
2. Explicitly configures test containers to remove the core dump size limit. (see dc55ab280821372a4e9c99217cb2d7f8fa199c7e)

### DIFF
--- a/tests/runtestcase
+++ b/tests/runtestcase
@@ -184,7 +184,7 @@ find_cores() {
         if [ ! -f $cr ] ; then 
             ln ${cr} -t $LCLDBDIR || cp ${cr} $LCLDBDIR #if can't hardlink then just copy
         fi
-        which gdb > /dev/null && [ -f $cr ] && echo 'where' | gdb -q $COMDB2_EXE $cr &>> $LCLDBDIR/core_stacktrace.$HOSTNAME.txt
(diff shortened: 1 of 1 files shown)


morgando:
See results of test that generates cores: http://roborivers.dev.bloomberg.com/tests/8e9bc0a5_22495/detail.txt

We should probably make that output less verbose by not printing the backtrace for every thread.

morgando:
Also added 5b50821260174aa7ccc323e2227f2a355acd2797 to cut down backtrace noise and tweaked the test infra so it doesn’t dump backtraces for every thread on the summary page. See new sample submission: http://roborivers.dev.bloomberg.com/tests/0178e20d_25791/detail.txt

PR #5486, opened 2025-10-13
Abort ourselves if libevent reports a fatal error

riverszhang89:


riverszhang89:
@markhannum Thanks for reviewing!

PR #5488, opened 2025-10-13
Port newsql_connect changes

chands10:
May not have set hndl->is_rejected everywhere yet that is needed since logic is different between open and bb


chands10:
cdb2test   Oct 13 16:51:33 2025      success c10.R20251013.6

chands10:
^^ fixed merge conflicts

PR #5493, opened 2025-10-14
Update altdb metrics from watcher

markhannum:
Physrep sources should update both the metadb and all alt dbs.


mohitkhullar (approved):
lgtm

PR #5496, opened 2025-10-14
remote write only transaction need to also reset clnt effects before accouting

dorinhogea:
If the local master resets a replicant clnt effects structure, and that makes the final accounting for effects work.  If a transaction has only remote writes, there is no local master so the accounting double counts remote writes if verifyretry is off.  Fix is to reset effects in this case too.


dorinhogea:
@mohitkhullar thank you

PR #5520, opened 2025-10-24
Bug Fix: Fix bug with caching modsnap table versions

morgando:
The table versions we cache need to be the ones that were active at our snapshot LSN. The changes in this PR guarantee this by holding the schema lock while we grab the snapshot LSN, and keeping it until after we’ve cached the table versions.
If this setup weren’t safe, one of two bad things would have to happen:
1. A schema change that happened at or before the snapshot LSN didn’t get cached, or
2. A schema change that happened after the snapshot LSN did get cached.

Let’s prove that neither of these things can happen:
1. Because we grab the schema lock before taking the snapshot LSN and hold it while caching table versions, we know any schema changes that finished up to and including our snapshot LSN are already done and visible to us. Schema changes always grab the schema lock before they update the snapshot LSN and keep it until they finish applying. So when we take the snapshot LSN under the same lock, those earlier changes are guaranteed to be included.
2.  Schema changes that start after our snapshot LSN will be blocked, since they can’t grab the schema lock while we’re holding it. That means they can’t proceed until after we’re done caching table versions, so they definitely won’t show up in our cache.

**note** This is still broken if we're using durable LSNs or doing PIT snapshots. To support those, we'll need to figure out some way of caching table versions from a past LSN.

### DIFF
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -4993,34 +4993,6 @@ int start_new_transaction(struct sqlclntstate *clnt)
     return 0;
 }
 
-static int cache_table_versions(struct sqlclntstate *clnt)
-{
-    if (clnt->dbtran.table_version_cache) {
-        bdb_free_table_version_cache(clnt->dbtran.table_version_cache);
-        clnt->dbtran.table_version_cache = NULL;
-    }
-
-    int rc = bdb_init_table_version_cache(&clnt->dbtran.table_version_cache);
-    if (rc) {
-        logmsg(LOGMSG_ERROR,
-               "%s failed initialize table verison cache rc=%d\n",
-               __func__, rc);
-        return rc;
-    }
-
-    int bdberr;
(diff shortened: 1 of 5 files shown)


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>6/630 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>truncatesc<br>sc_timepart<br>analyze_partial_index_off_generated<br>insert_lots_ssl_generated<br>insert_lots<br>reco-ddlk-sql</p>

morgando:
`mdouglas47 Nov 11 15:43:32 2025      success fix_modsnap_bug.R20251111.6`

PR #5526, opened 2025-10-29
Do not overwrite test certificates

riverszhang89:
The cdb2jdbc testcase regenerates and overwrites certificates in the same test directory, interfering with other tests on the fly. This patch changes the cdb2jdbc testcase to generate certificates in its own directory.


mponomar (dismissed):

riverszhang89:
@mponomar thanks for reviewing!

PR #5528, opened 2025-10-30
{181155186} stored procedure's cstring performance

rishitha957:
EN: currently lua functions like gmatch, gsub{str_gsub}, g_write(io_write, fwrite), len {str_len}, str_reverse, str_lower.. most of the strlib functions call strlen on cstring type making it expensive. (Also because lua strings are immutable, we can store len of the string for cstring as we push on to the lua's stack


lua/ltypes.c
```diff
@@ -571,16 +571,16 @@ const char *luabb_tostring_noerr(Lua L, int idx)
...
     const char *c;
-    size_t s;
+    size_t s = 0;
     int type = luabb_type(lua, idx);
     if (type == DBTYPES_LSTRING) {
         c = lua_tostring(lua, idx);
-str:    s = strlen(c);
-        ret->data = strdup(c);
+str:    ret->data = strdup(c);
         ret->length = s;
```
SChakravorti21, line 579: Maybe I'm not understanding this correctly, could you clarify where `s` gets set to the string's length if we're dealing with an `LSTRING`? I can see for `CSTRING` that we grab the cached length, but for `LSTRING` will it always be set to 0?

lua/luautil.c
```diff
@@ -591,6 +591,16 @@ const char *luabb_tocstring(Lua l, int idx)
     return ptr->val;
 }
 
+const char *luabb_tolcstring(Lua l, int idx, size_t *len)
```
SChakravorti21, line 585: `luabb_tocstring` seems to only be called in one place — would it make sense to update the existing signature or remove the old implementation?

tests/sp_perf.test/runit
```diff
@@ -0,0 +1,125 @@
...
+
+}
+
+checkdiff() {
+if [ -f "$1" ]; then
+    if ! diff -w $2 $1 > /dev/null 2>&1 ; then
+        echo "FAILED: output $1 diff from expected $2"
+        echo "diff:"
+        diff -w $2 $1
+        exit 1
```
SChakravorti21: We can use the `failexit` utility to fail a test case with an error message. It has some nice integration with the test framework as well (e.g. the error message gets logged to a dedicated file and RR can tell you where it is).

tests/tools/CMakeLists.txt
```diff
@@ -75,6 +75,7 @@ add_exe(simple_ssl simple_ssl.c)
 add_exe(sirace sirace.c)
 add_exe(slowreaders slowreaders.c)
 add_exe(smartbeats smartbeats.c)
+add_exe(sp sp.c)
```
SChakravorti21, line 79: Was the idea to use this executable in the test case?

lua/ltypes.c
rishitha957: yes, thank you, changed it to lua_tolstring now (which returns string length to s), also added t1.sql to test this path..

PR #5538, opened 2025-11-04
call recover deadlock on lock desired

dorinhogea:
This will make remote cursors more efficient while still fixing the uninterruptible cdb2api call while locking problem


markhannum (approved):

dorinhogea:
@markhannum thank you

PR #5541, opened 2025-11-04
{178862843} fix(crc32c): avoid misaligned memory accesses

SChakravorti21:
This bug was initially reported by @hallfox (thank you!). The issue is that `crc32c_sse` may generate misaligned memory accesses because of the order of operations it performs:

1. Compute the checksum until an 8 byte-aligned address is reached.
2. Compute the checksum until an exact multiple of 1024 bytes is remaining.
3. Compute the checksum for the remaining bytes in chunks of 1024 bytes.

Unfortunately, Step 2 can kick us out of alignment, which breaks the requirement that we be at an alignment boundary at the beginning of Step 3.

We can ensure alignment by doing the following instead:

1. Compute the checksum until an 8 byte-aligned address is reached.
2. Compute the checksum for as many 1024 byte chunks as possible.
3. Compute the checksum for the remaining bytes.

Reproducing the original issue uncovered another one, which was that some left-shifts performed in `crc32c_software` could result in overflow since we did not explicitly upcast the shifted bytes (`uint8_t`) to the target storage type (`uint32_t`). This is fixed by upcasting before shifting.


roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>5/631 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>insert_lots<br>analyze_partial_index_off_generated<br>truncatesc_offline_generated<br>insert_lots_ssl_generated<br>reco-ddlk-sql</p>

SChakravorti21:
Failing tests are all quarantined.

crc32c/CMakeLists.txt
```diff
@@ -1,5 +1,15 @@
+set(crc32c_include_directories ${PROJECT_SOURCE_DIR}/util)
+
+# Build the crc32c implementation as a library to be linked into the database server.
 add_library(crc32c crc32c.c)
-include_directories(${PROJECT_SOURCE_DIR}/util)
+target_include_directories(crc32c PRIVATE ${crc32c_include_directories})
+
+# Build the crc32c test executable with UBSAN.
+add_executable(crc32c_test crc32c.c)
```
akshatsikarwar: This should go under `tests/tools`

SChakravorti21: Are you suggesting moving the tests into a separate `.c` file under `test/tools` and removing them from `crc32c.c` entirely?

akshatsikarwar: That would work as well, and probably it would be better that way. I was initially thinking moving just CMake portion.

SChakravorti21: I see, let me see what works. I suspect the library code also needs to be compiled with UBSAN for it to be able to instrument it, but let me try it out to be sure.

SChakravorti21: I ended up moving out the tests because it does feel cleaner. However, the `crc32c` library itself needs to be compiled with UBSAN, not just the test driver. Tried my best to keep things organized but let me know what you think!

SChakravorti21:
Strange, changes don't build in Actions. Will take a look.

roborivers (changes requested):
<p><a href='https://bloomberg.github.io/comdb2/contrib.html#code-style'>Coding style</a> check: Success &#10003;.<br>Smoke testing: Success &#10003;.<br>Cbuild submission: Success &#10003;.<br>Regression testing: <b>0/0 tests failed</b> &#x26A0;.</p>

crc32c/crc32c.h
```diff
@@ -26,16 +26,18 @@ extern "C" {
...
+
 #if defined (__x86_64) || defined(_HAS_CRC32_ARMV7) || defined(_HAS_CRC32_ARMV8) 
     void crc32c_init(int v);
-    #define crc32c(buf, sz) crc32c_comdb2(buf,sz)
     uint32_t crc32c_comdb2(const uint8_t* buf, uint32_t sz);
 #else
     #define crc32c_init(...)
-    #define crc32c(x, y) crc32c_software((x), (y), CRC32C_SEED)
-    uint32_t crc32c_software(const uint8_t* data, uint32_t size, uint32_t crc);
+    #define crc32c_comdb2(buf, sz) crc32c_software((buf), (sz), CRC32C_SEED)
```
SChakravorti21, line 36: Added the declaration of `crc32c_comdb2` here so that the test driver compiles on Mac (the tests need both `crc32c_comdb2` and `crc32c_software` to exist).

akshatsikarwar (approved):

SChakravorti21:
@akshatsikarwar thank you!

PR #5548, opened 2025-11-05
Bug Fix: Don't treat COMDB2_SCHEMACHANGE_OK as an error

morgando:
Previously, the `COMDB2_SCHEMACHANGE_OK` return code was ignored only after blocksql commits. This PR extends that behavior to recom, snapshot, and serializable commits as well.

This change fixes a bug where those transaction types would incorrectly report an error when clients added default consumers, even though the operation had actually succeeded.


morgando:
/adhoc-test consumer consumer_non_atomic_default_consumer_generated

PR #5566, opened 2025-11-11
Make sure txns are recom in selectv_recom suite

morgando:
This suite is supposed to test selectv under read committed isolation, but only actually sets the transaction level to read committed for one test in the suite. The changes in this PR make read committed the default txn level in this suite's lrl so that the it actually tests selectv under read committed.

### DIFF
--- a/tests/selectv_recom.test/t2_01.req.exp
+++ b/tests/selectv_recom.test/t2_01.req.exp
@@ -5,7 +5,6 @@ done
 done
 (rows deleted=1)
 done
-[commit] failed with rc -103 constraints error, no genid
 done
 (rows inserted=1)
 done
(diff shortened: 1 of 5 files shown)


tests/selectv_recom.test/t2_01.req.exp
```diff
@@ -5,7 +5,6 @@ done
 done
 (rows deleted=1)
 done
-[commit] failed with rc -103 constraints error, no genid
```
morgando, line 8: Read-only selectv txns don't fail because of the genid-checking optimization. We have plans to change this so that txns like this fail, but this is the behavior for now.

mponomar (approved):

PR #5591, opened 2025-11-21
Bug Fix: Clear snapshot state in sqlite_done

morgando:
Snapshot's implementation assumes that every statement going through `get_prepared_stmt` eventually calls `sqlite3BtreeCommit`, where the snapshot is marked as finished. However, that’s not always true — for example, `select sleep(0)` doesn't result in a call to `sqlite3BtreeCommit`.

This PR fixes a bug where, in those cases, the next statement would incorrectly reuse the same snapshot because the previous one was still marked as “in progress.” The fix ensures snapshots that are still running in `sqlite3_done` and are not part of explicit client transactions are marked as finished. It also adds assertions to verify that snapshots aren’t “in progress” in unexpected places.

### DIFF
--- a/db/sqlinterfaces.c
+++ b/db/sqlinterfaces.c
@@ -1816,6 +1816,7 @@ int handle_sql_begin(struct sqlthdstate *thd, struct sqlclntstate *clnt,
                 (clnt->sql) ? clnt->sql : "(???.)");
 
     /* Latch the last commit LSN */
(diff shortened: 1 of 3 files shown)


mponomar (approved):

morgando:
`cdb2test   Nov 21 15:59:56 2025      success commit_in_sqlite_init.R20251121.4`

PR #5607, opened 2025-12-04
print debugging trace only when log_calls is enabled

riverszhang89:
/skipbuild


SChakravorti21 (approved):

riverszhang89:
Thanks for reviewing!

SChakravorti21:
ATR!

PR #5615, opened 2025-12-10
Fix union-all subquery with limit flattening optimization (port from newer sqlite)

dorinhogea:
Port fix from sqlite https://sqlite.org/src/info/2025-12-09T19:15:52Z
i.e. fix limit for union-all subqueries can that flattened.

re: 181655879


mponomar (approved):

dorinhogea:
@mponomar thank you

PR #5630, opened 2025-12-16
testcase for union all subquery with limit

dorinhogea:
Testcase for fix from sqlite https://sqlite.org/src/info/2025-12-09T19:15:52Z

  i.e. fix limit for union-all subqueries can that flattened.

re: 181655879


dorinhogea:
@mponomar thank you!

PR #5636, opened 2025-12-18
Require protobuf schema change serialization

morgando:
Deletes non-protobuf schema change serialization code.

Don't merge until we're sure we won't want to do a direct 8.0 -> 8.2 upgrade (8.0 doesn't understand the protobuf format, so we would need to have it disabled during the upgrade process so that the nodes coming up as 8.2 don't send messages that the 8.0 nodes don't understand)

### DIFF
--- a/schemachange/sc_struct.c
+++ b/schemachange/sc_struct.c
@@ -139,110 +139,6 @@ void free_schema_change_type(struct schema_change_type *s)
     }
 }
 
-static size_t dests_field_packed_size(struct schema_change_type *s)
-{
-
-    size_t len = sizeof(s->dests.count);
-    int i;
-    struct dest *d;
-    for (i = 0, d = s->dests.top; d; d = d->lnk.next, i++) {
-        len += sizeof(int);
-        len += strlen(d->dest);
-    }
-    return len;
-}
-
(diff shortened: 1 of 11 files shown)


schemachange/sc_struct.c
```diff
@@ -1902,23 +1593,12 @@ struct schema_change_type *clone_schemachange_type(struct schema_change_type *sc
...
 
-    if (is_protobuf) {
-        size_t plen;
-        rc = unpack_schema_change_protobuf(newsc, buf, &plen);
-        if (rc != 0) {
-            free_schema_change_type(newsc);
-            free(buf);
-            return NULL;
-        }
-        p_buf = p_buf_end = buf + plen;
```
morgando, line 1913: I got rid of this line because if we return zero from `pack_schema_change_protobuf`, then `buf` is guaranteed to be non-null. if `buf` is non null, then `p_buf` won't be null. If `p_buf` is never null, then we don't need to compute `p_buf` and can get rid of the `if (!p_buf)` block below.

morgando: also removed `p_buf_end` because it is not used.

PR #5638, opened 2025-12-18
newsql: No timeout parameter for add_rd_event

akshatsikarwar:
/silent


akshatsikarwar:
The only 2 callers both call `add_rd_event` with `NULL`.

mponomar (dismissed):

PR #5648, opened 2025-12-22
{181677165}: Fixing memory leak in dohast order-by processing

riverszhang89:
This patch frees the `order_dir` variable on the parent dohsql struct before overwriting it.


dorinhogea (approved):
lgtm

riverszhang89:
@dorinhogea Thanks for reviewing!

PR #5660, opened 2026-01-06
cdb2api_direct test

mohitkhullar:


tests/tools/cdb2api_direct.cpp
```diff
@@ -0,0 +1,63 @@
...
+    assert(rc == CDB2_OK_DONE);
+
+    rc = cdb2_close(hndl);
+    assert(rc == 0);
+}
+
+static void load_cfg(char *db, char *tier)
+{
+    cdb2_hndl_tp *hndl = NULL;
+    int rc = cdb2_open(&hndl, db, tier, 0);
```
chands10: tier is a hostname here. I think using 0 flag here with hostname will only work if dbinfo optimization is used, that's why it passes on btda (guess it's already in sockpool). If it doesn't grab from sockpool then will fail.

But really should be using DIRECT_CPU flag here I think

chands10: tier is a hostname here. Currently fails on roborivers when using 0 flag (which honestly makes sense). There's 2 reasons why is passes on bb cdb2api.

1. It skips dbinfo (open does this too by default). To skip dbinfo also require to have a connection in sockpool to the db already (roborivers won't have that, at least doesn't in it's current setup?) (at least for open sockpool). If bb didn't skip dbinfo then it would also complain about using 0 as flag
2: on bb sockpool typestring is comdb2/dbname/hostname/newsql/random_roo (the m doesn't fit in). bbcdb2api doesn't care that the length of the sockpool typestr and will still attempt to contact sockpool. Open does care and will check the length (https://github.com/bloomberg/comdb2/blob/main/cdb2api/cdb2api.c#L3099). Therefore since the typestring doesn't fit, open won't skip dbinfo, and going through the normal path you can't use 0 flag and hostname

chands10: Also not even sure how you get a hostname/random_roo(m) combo in sockpool to start with, and not hostname/dc. Maybe because bb sockpool will open a new connection for you if needed

tests/tools/cdb2api_direct.cpp
```diff
@@ -0,0 +1,91 @@
...
+    assert(rc == CDB2_OK);
+    rc = cdb2_next_record(hndl);
+    assert(rc == CDB2_OK_DONE);
+    rc = cdb2_close(hndl);
+    assert(rc == 0);
+}
+
+int main(int argc, char **argv)
+{
+    if (argc < 3) {
```
chands10: 4?

chands10 (dismissed):
lgtm

mohitkhullar:
Thanks

PR #5666, opened 2026-01-09
Timeout will not kill the subprocesses.  Also, buffering can keep the gawk subprocess alive

dorinhogea:
^title


dorinhogea:
@riverszhang89 thank you for the review!

PR #5703, opened 2026-02-01
Fix broken analyze for aliased tables

dorinhogea:
This affects time partitions created on existing tables; the initial analyze works; the follow-up fails because code tries to delete same rows twice by calling backup for stats for both tablename and sqlaliasname.

Also, preserve aliased shard 0 analyze, i.e. if we run "analyze $0_..." do not convert that to "analyze <partition_name>"


dorinhogea:
thank you Mohit!

PR #5708, opened 2026-02-04
{169607487} feat: enable distributed trace addon by default

SChakravorti21:
Enable the (Bloomberg-internal) distributed trace plugin by default so that users get it out of the box. For open source builds of cdb2api, this is a no-op.

Additionally, provide a way to disable this behavior. We already have the ability to do so through an environment variable
(`COMDB2_CONFIG_UNINSTALL_STATIC_LIBS=dt`) or using the config file (`comdb2_config:uninstall_static_libs_v4=dt`), but both only accept _one_ addon name. If we want to enable more plugins by default in the future, users should be able to selectively disable them as well. Support supplying a list of plugin names for both the environment variable and the config file setting.


SChakravorti21:
oops, forgot to specify plugin branch

PR #5715, opened 2026-02-06
{94886844} feat: add with_effects to db_commit

emelialei88:
Extends the Lua stored procedure `db:commit()` API to optionally return transaction effects (row-change counts) after a successful commit. `db:commit()` now accepts an optional Lua table parameter with a single boolean field `with_effects`. When called as `rc, effects = db:commit({ with_effects = true })`, it returns the usual result code plus an effects Lua table matching the struct `cdb2_effects_type`. Default behavior remains unchanged when the option is not provided.


tests/sp.test/t01.req.out
```diff
@@ -129,6 +129,12 @@ SP: exec procedure bound(@i, @u, @ll, @ull, @s, @us, @f, @d, @dt, @cstr, @ba, @b
 ($0=x'00112233445566778899001122334455667788990011223344556677889900112233445566778899')
 (version='1')
 (rows inserted=1)
+Testing db:commit with with_effects
```
akshatsikarwar, line 132: Add a test which does all operations in a single transaction as well? Also test for `num_selected`, `num_affected`.

akshatsikarwar:
Also need to  update docs - https://bloomberg.github.io/comdb2/storedprocs.html#dbcommit (and then on Tutti as well)

PR #5725, opened 2026-02-11
tests: Tools get CDB2API_TEST by default

akshatsikarwar:
/silent

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,6 @@ endif()
 option(COMDB2_TEST "Turn ON to compile with testing features" OFF)
 if(COMDB2_TEST)
   add_definitions(-DCOMDB2_TEST)
-  add_definitions(-DCDB2API_TEST)
 endif()
 
 option(WITH_TINFO "Turn ON on platforms where tinfo library is required for cdb2sql" OFF)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -18,7 +18,6 @@ endif()
 option(COMDB2_TEST "Turn ON to compile with testing features" OFF)
 if(COMDB2_TEST)
   add_definitions(-DCOMDB2_TEST)
-  add_definitions(-DCDB2API_TEST)
```
chands10, line 21: I think we should still have this line actually. There are files in plugins that also use cdb2api_test

chands10: I guess that doesn't matter for testing open though

tests/tools/CMakeLists.txt
```diff
@@ -17,10 +17,13 @@ include_directories(
...
 
-add_definitions(-DSBUF2_SERVER=0)
+add_definitions(
+  -DSBUF2_SERVER=0
+  -DCDB2API_TEST
+)
 
 macro(add_exe exe)
-    add_executable(${exe} ${ARGN})
+    add_executable(${exe} ${ARGN} ${PROJECT_SOURCE_DIR}/cdb2api/cdb2api.c)
```
chands10, line 26: need cdb2sql

PR #5731, opened 2026-02-12
fix(cdb2jdbc): preserve metadata table cursor after version lookup

### LINKED ISSUE #5243: Unable to get tables and db version from Comdb2MetaData sequentially
**Describe the bug**
When querying the database for all available tables through the Comdb2MetaData object in `cdb2jdbc`, if the database version is also queried, it breaks and is unable to find the tables after. Not sure if this is intended behavior

**To Reproduce**
Here is an example test case that reproduces the issue
https://gist.github.com/jeffrytyn/50403cb3b74dcdb8457aa537d771b6e0

**Expected behavior**
I expect the test to pass but it fails

**Additional context**
Looking inside the `Comdb2MetaData` class, it has instance variables `ps` and `stmt`. 
`ps` is creating inside `getTables` method from the `conn` object
`stmt` is created in the constructor of the class using the same `conn` object
This `conn` object itself has an instance variable `hndl`

The ResultSet `tables` is the output of `ps.executeQuery`, where `ps` is a `Comdb2PreparedStatement` created with `hndl`.
`version` is the output of `stmt.executeQuery`, where `stmt` is a `Comdb2Statment` created with `hndl`

`tables.next()` invokes `hndl.next()`, but because the same `hndl` was used to get the version, it now returns false instead of true. Seems like different `hndls` should be used

jmahotiedu:
## Summary
- Fixes an interaction in `cdb2jdbc` where `DatabaseMetaData#getDatabaseProductVersion()` could invalidate an active `getTables()` result set.
- Runs the product-version query on an isolated duplicate connection handle.
- Adds a regression test that opens a tables result set, calls product-version lookup, and then verifies the tables result set is still readable.

## Root Cause
`Comdb2DatabaseMetaData` reused the same underlying handle for both `getTables()` (prepared statement) and `getDatabaseProductVersion()` (statement). Executing the version query could consume/replace cursor state used by the tables result set.

## Verification
- Added `testGetTablesResultSetSurvivesProductVersionLookup` in `DatabaseMetaDataTest`.
- Local execution note: `./gradlew test --tests com.bloomberg.comdb2.jdbc.DatabaseMetaDataTest` is currently blocked in this environment because `gradle-4.2` does not support the installed Java runtime (`20.0.2`).

Closes #5243.

### DIFF
--- a/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2DatabaseMetaData.java
+++ b/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2DatabaseMetaData.java
@@ -28,12 +28,19 @@ public class Comdb2DatabaseMetaData implements DatabaseMetaData {
     private String url;
     Comdb2Connection conn;
     private String sqlkws = null;
+    private String databaseProductVersion = "UNKNOWN";
+    private SQLException databaseProductVersionException = null;
 
     Comdb2DatabaseMetaData(Comdb2Connection conn) throws SQLException {
         url = "jdbc:comdb2:" + conn.getDatabase() + ":" + conn.getCluster();
         this.conn = conn;
         conn.lookup();
         stmt = conn.createStatement();
+        try {
+            databaseProductVersion = readDatabaseProductVersion(stmt);
+        } catch (SQLException e) {
+            databaseProductVersionException = e;
+        }
     }
 
     public <T> T unwrap(Class<T> iface) throws SQLException {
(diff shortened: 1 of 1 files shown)


riverszhang89:
Multiple resultsets on a single comdb2 connections aren't supported. This is by design. Any reasons that you can't retrieve all tables before making another metadata call?

jmahotiedu:
> Multiple resultsets on a single comdb2 connections aren't supported. This is by design. Any reasons that you can't retrieve all tables before making another metadata call?

Thanks for the clarification @riverszhang89, agreed that multiple open resultsets on a single Comdb2 handle are not supported.

I pushed a follow-up update to avoid that pattern entirely: `getDatabaseProductVersion()` no longer executes a query while another metadata ResultSet may be active. Instead, the version is fetched once during `Comdb2DatabaseMetaData` initialization and cached.

So this keeps the single-cursor design intact while still fixing #5243 (the `getTables()` ResultSet remains readable after a product-version lookup).

PR #5735, opened 2026-02-12
{182689785} fix: normalized sql for sp

emelialei88:
Change SQL normalization to always produce the same `normalized_sql` and fingerprint regardless of argument values, types, or count. All argument tokens are stripped so that e.g. both `EXEC PROCEDURE foo(1)` and `EXEC PROCEDURE foo(1, 'hello', x'DEADBEEF')` normalize to `EXEC PROCEDURE foo();`.


rishitha957 (dismissed):
changes look good to me, tested on my local, I'd just wait for rr tests to run.,

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>0/0 tests failed</b> &#x26A0;.</p>

PR #5739, opened 2026-02-16
{182782396} Look up peer's short hostname

riverszhang89:
This patch includes a slight refactoring around hostname lookup, also fixes incorrect IP address when peer drops connection.

### DIFF
--- a/util/hostname_support.c
+++ b/util/hostname_support.c
@@ -15,6 +15,7 @@
  */
 
 #include <arpa/inet.h>
+#include <errno.h>
 #include <netdb.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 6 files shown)


util/hostname_support.c
```diff
@@ -23,19 +24,23 @@
...
 
 #ifdef CDB2API_TEST
 #include <cdb2api_ssl_test.h>
 #endif
 
+int gbl_get_peer_fqdn = 1;
 static int get_hostname_by_addr(struct sockaddr_in *addr, char *host, socklen_t hostlen)
 {
+    int rc;
+    int flags = gbl_get_peer_fqdn ? 0 : NI_NOFQDN;
```
rishitha957, line 37: curious where we defined this

rishitha957 (approved):
makes sense, we use NI_NOFQDN flag to either get short or fqdn

PR #5747, opened 2026-02-18
Hide things from cdb2api if not server build

chands10:
Create new flag cdb2api_server for server builds (I guess currently open cdb2api would contain this, but we can at least hide from bb to start)

Can't use sbuf2server since that's defined as 0 in cdb2api

Open to suggestions if there is a better method


dorinhogea (approved):
looks good to me

chands10:
cdb2test   Feb 18 17:41:00 2026      success hide_server.R20260218.8

chands10:
Thanks Dorin!

PR #5760, opened 2026-02-24
{182577691} fix: make tonumber accept comdb2 number

emelialei88:
**1. Make `luaV_tonumber` handle comdb2 integer and real types**

Extended `luaV_tonumber` to convert comdb2 `int` and `real` userdata to Lua numbers, enabling arithmetic between comdb2 types and Lua numbers without going through metamethods.

**2. Remove broken userdata handling from `OP_MOVE` and `OP_LOADK`**

Both opcodes had runtime checks attempting to detect typed comdb2 variables and route through the `__type` metamethod. These checks were buggy (check [lopcodes.c](https://github.com/bloomberg/comdb2/blob/main/lua/lopcodes.c#L62) for more information):

- `OP_LOADK` is `iABx` format (B and C are combined into one 18-bit Bx field), but the code called `RKB(i)` which reads the upper 9 bits of Bx as a register index — pointing to an unrelated register.
- `OP_MOVE` read `RKC(i)` where C is unused, always resolving to R(0).

Both could accidentally trigger the typed assignment path when an unrelated register held a comdb2 userdata.

These checks are also unnecessary: the compiler emits `OP_CAST` for typed assignment (`:=`) and plain `OP_MOVE`/`OP_LOADK` for regular assignment (`=`). Per documentation, plain `=` is intended to lose type. No tests rely on this runtime behavior.

**3. Restore `lua_to_client_type` check order**

Comdb2 userdata must be identified before `lua_isnumber`/`lua_isstring`, since `luaV_tonumber` now succeeds on comdb2 int/real and would misclassify them as `SQLITE_FLOAT`.

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -229,7 +229,6 @@ int gbl_return_long_column_names = 1;
 int gbl_maxreclen;
 int gbl_penaltyincpercent = 20;
 int gbl_maxwthreadpenalty;
-int gbl_spstrictassignments = 0;
 int gbl_lock_conflict_trace = 0;
 int gbl_move_deadlk_max_attempt = 500;
 
--- a/db/db_tunables.c
+++ b/db/db_tunables.c
@@ -100,7 +100,6 @@ extern int gbl_debug_sleep_on_set_read_only;
 extern int gbl_debug_wait_on_verify_off;
 extern int gbl_debug_disttxn_trace;
 extern int gbl_sparse_lockerid_map;
-extern int gbl_spstrictassignments;
 extern int gbl_early;
 extern int gbl_exit_alarm_sec;
 extern int gbl_fdb_default_ver;
--- a/db/db_tunables.h
+++ b/db/db_tunables.h
@@ -501,9 +501,6 @@ REGISTER_TUNABLE("enable_sparse_lockerid_map",
                  "If set, allocates a sparse map of lockers for deadlock "
                  "resolution. (Default: on)",
                  TUNABLE_BOOLEAN, &gbl_sparse_lockerid_map, READONLY | NOARG, NULL, NULL, NULL, NULL);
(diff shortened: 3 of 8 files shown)


akshatsikarwar:
This was failing robomark build:
`error: implicit declaration of function 'cdb2_is_valid_int' [-Werror=implicit-function-declaration]`

I rebased the branch and submitted for robom.

tests/sp.test/t19.req
```diff
@@ -0,0 +1,88 @@
+DROP TABLE IF EXISTS t19
```
mohitkhullar, line 1: Add test for decimal type's exclusion

lua/lvm.c
```diff
@@ -24,24 +24,38 @@
...
   }
-  else
-    return NULL;
+
+  if (ttisuserdata(obj)) {
+    const lua_dbtypes_t *bb = luabb_todbpointer(obj);
+    if (bb->is_null) return NULL;
+    switch (bb->dbtype) {
+    case DBTYPES_INTEGER:
+      setnvalue(n, (lua_Number)(((lua_int_t *)bb)->val));
```
mohitkhullar, line 50: add test for number close to max longlong?

tests/sp.test/t19.req
```diff
@@ -0,0 +1,88 @@
...
+    local row = db:exec("select i, r from t19 where i is not null"):fetch()
+    db:emit(row.i + row.r)
+end
+}$$
+exec procedure tonumber_arith_mixed()
+
+create procedure tonumber_arith_null_int version 'sptest' {
+local function main()
+    local row = db:exec("select i from t19 where i is null"):fetch()
+    db:emit(row.i + 10)
```
mohitkhullar, line 77: add tests where lua_number comes before comdb2's? like 10 + row.i

PR #5726, opened 2026-02-11
Fdb locking version 2

dorinhogea:
Idea is to use short term mutexes to protect arrays and hashes, and long term read locks for fdb and table read access.

There is a live read/write for each fdb object.  As long as there is a pointer to fdb, a read lock is acquired for that fdb.
Each remote table is protected by its own long term read lock (similar to a table lock).

There are two intervals when both table locks and fdb live lock are acquired: 1) during query preparing, if remote table discovery is needed, when we retrieve and attach remote tables; these locks are released once setup is done; the sqlite_stat tables are also locked during this phase 2) during query execution; we get an fdb live lock and table locks for each remote table locked, and we release them when unlocking the remote table

Updating a table requires an exclusive lock on that table; to avoid blocking for long duration when a remote table is schema changed, we use a mvcc scheme that COW new table objects and detects readers using trywrlock. Last reader of a stable table will free that object.

PR #5764, opened 2026-02-25
Testcase for long blocking fdb locks during table discovery.

dorinhogea:
/skipbuild 

It is added to quarantine since it is current code fails this test.  
https://github.com/bloomberg/comdb2/pull/5726 will fix and remove from quarantine.


dorinhogea:
@mponomar @SChakravorti21 thank you !

PR #5766, opened 2026-02-25
Allow optional comdb2_legacy parameter to track tag origin.

mponomar:


plugins/newsql/newsql_evbuffer.c
```diff
@@ -479,10 +486,18 @@ static int dispatch_tagged(struct sqlclntstate *clnt)
...
-    if (appdata->sqlquery == NULL || appdata->sqlquery->n_bindvars < 3 || appdata->sqlquery->bindvars[0] == NULL ||
-        appdata->sqlquery->bindvars[1] == NULL || appdata->sqlquery->bindvars[1]->value.len != sizeof(int) ||
-        appdata->sqlquery->bindvars[2] == NULL || appdata->sqlquery->bindvars[2]->value.len != sizeof(int)) {
-        return 1;
+
+    if (appdata->sqlquery->n_bindvars > 3 && appdata->sqlquery->bindvars[3] &&
+        appdata->sqlquery->bindvars[3]->value.data) {
+        // note - we specifically don't override clnt->argv0 here - we want
+        // the original sender (typically the proxy) there, and we save on whose behalf we're
+        // proxying elsewhere.
```
SChakravorti21, line 494: nit: is this wording a little clearer? I initially read "original sender" as the thing `origin_argv0` should refer to.

```suggestion
        // note - we specifically don't override clnt->argv0 here - we want
        // the immediate client (typically the proxy) there, and we save on whose behalf we're
        // proxying in clnt->origin_argv0.
```

mponomar: Yes, it is.  I'll change it in a cleanup PR.

PR #5773, opened 2026-02-27
{182881698} cdb2jdbc: Fixing driver name and version

riverszhang89:
The JDBC driver did not send driver name and version to server. This patch fixes it.


riverszhang89:
@mponomar Thanks for reviewing!

PR #5782, opened 2026-03-03
Add delay malloc flag

chands10:
bbcdb2api has this undefined. Do we need to do that?

bb also has #undef CDB2BUF_UNGETC and #undef CDB2BUF_RWFUNCS


akshatsikarwar:
Do not need CDB2BUF_DELAY_MALLOC - delete all code under it

akshatsikarwar:
`cdb2buf_ungetc` is only used by the server.

chands10:
Like this?

PR #5789, opened 2026-03-06
remove duplicated unlock

dorinhogea:
On a syntax error schema change validation we unlock csc2 subsystem twice.


dorinhogea:
@mponomar thank you

PR #5795, opened 2026-03-09
Emit correctly endianized locklists by default

markhannum:
This has been fixed for a year- this PR enables endianized locklist by default.


SChakravorti21:
On by default: https://github.com/bloomberg/comdb2/blob/d1af879d43af9fa5547bc9ced820c901ab3fac2b/berkdb/txn/txn.c#L1025

PR #5798, opened 2026-03-09
Declare sockpool function

chands10:
/skipbuild


chands10:
Would not compile if CDB2API_SERVER not defined

PR #5800, opened 2026-03-09
Add new file for shared cdb2api flags

chands10:
/plugin-branch shared


cdb2api/CMakeLists.txt
```diff
@@ -26,8 +26,7 @@ include_directories(
 )
 
 # common obj files for .so/.dylib and .a
```
akshatsikarwar: We don't produce `.so/.dylib` anymore - remove that reference

chands10:
cdb2test   Mar  9 17:22:01 2026      success shared.R20260309.8

PR #5804, opened 2026-03-11
fix json_extract function with stat4

dorinhogea:
Our custom changes to sqlite breaks json_extract(), since sqlite does check for pVdbe and sets an error if set.


dorinhogea:
cdb2test   Mar 30 20:25:52 2026      success json.R20260330.10

dorinhogea:
@SChakravorti21 thank you!

PR #5805, opened 2026-03-12
spew every 1000 send/received remtran

dorinhogea:
this will let us detect the cases when a new server talk to a legacy server requiring remtran ("SENDING REMTRAN" alert). it will also let us know when a new server receives a legacy remtran from an older server ("RECEIVED REMTRAN" alert).
Preamble to remove remtran protocol


mponomar (approved):

dorinhogea:
@mponomar thank you!

PR #5807, opened 2026-03-12
Fix sql logfill

markhannum:
Fix bug in sql-logfill where the logfill thread won't apply a record which has a generation which is less than the node's current generation, as this assertion is only true if we explicitly set the generation during recovery by reading a commit or a checkpoint record which contains it.  Additionally, this provides the sqllogfill_reset_gen test which reproduces the issue and exercises the fix.

### DIFF
--- a/tests/noresetgen.test/Makefile
+++ b/tests/noresetgen.test/Makefile
@@ -1,0 +1,10 @@
+ifeq ($(TESTSROOTDIR),)
+  include ../testcase.mk
+else
+  include $(TESTSROOTDIR)/testcase.mk
+endif
+export CHECK_DB_AT_FINISH=0
+ifeq ($(TEST_TIMEOUT),)
+	export TEST_TIMEOUT=10m
+endif
+
--- a/tests/noresetgen.test/lrl.options
+++ b/tests/noresetgen.test/lrl.options
@@ -1,0 +1,4 @@
+# Use default options - features that preserve generation are enabled by default
+# This test verifies that generation number is preserved and increments across cluster restarts
+
+logmsg level debug
--- a/tests/noresetgen.test/runit
+++ b/tests/noresetgen.test/runit
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+
+bash -n "$0" | exit 1
+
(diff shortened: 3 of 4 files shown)


SChakravorti21 (commented):
Writing down my understanding so far...

We are introducing a tunable, `gbl_emit_gen_commits`, so that we can turn _off_ the logic for emitting cluster generation in commit records. By disabling this logic, a replicant can't possibly pick up its previous cluster generation from the transaction log. In other words, it helps us test how logfill behaves when a node restarts and its cluster generation resets.

If we haven't enabled `elect_highest_committed_gen` or `match_on_ckp`, then our gen could foreseeably be much lower (defaults to 0) than that of the current master. We want logfill to be able to continue making progress in this case, hence the changes to `gen_okay`.

The `reset_gen` test makes sure the logfill makes progress even when our gen resets on restart, instead of refusing to apply log records.

The `noresetgen` test ensures that with `elect_highest_committed_gen`, `match_on_ckp`, etc. enabled, our gen is always fetched from logs on restart instead of resetting.

Questions:

* In `gen_okay`, do we need to check for `retrieve_gen_from_ckp` in addition to `match_on_ckp`? Or does `match_on_ckp` imply `retrieve_gen_from_ckp`? How does `match_on_ckp` help guarantee that we have retrieved our most recent gen from the transaction logs?

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>2026-03-18T10:46:47EDT [275862] consumer_non_atomic_default_consumer_generated  **quarantined**<br>2026-03-18T10:46:47EDT [275862] remsql_locks_rte_connect_generated  **quarantined**<br>2026-03-18T10:46:47EDT [275862] remsql_locks  **quarantined**<br>2026-03-18T10:46:47EDT [275862] truncatesc_offline_generated [timeout] **quarantined**<br>2026-03-18T10:46:47EDT [275862] reco-ddlk-sql [timeout] **quarantined**</p>

markhannum:
mhannum    Mar 24 15:47:41 2026      success fix_sql_logfill.R20260324.5

PR #5784, opened 2026-03-04
Fix resuming a retroactively time partitioning, upon a master swings.

dorinhogea:
^title

It requires multitable_ddl option.
it preserves the already partitioning effort.
It decides the sc_genids values as maximum genids for each shard and stripe.

PR #5815, opened 2026-03-13
start needs to be at least a configurable number of hours in the future (default 24)

dorinhogea:
This is simply to prevent the case when clients choose a start time for retroactive time partition in the past.  Start is the next rollout, not the first rollout.

### DIFF
--- a/db/db_tunables.c
+++ b/db/db_tunables.c
@@ -595,6 +595,7 @@ extern int gbl_transaction_grace_period;
 extern int gbl_partition_sc_reorder;
 extern int gbl_retro_tpt;
 extern int gbl_retro_tpt_verbose;
+extern int gbl_retro_tpt_start;
 extern int gbl_dohsql_joins;
 extern int gbl_altersc_latency;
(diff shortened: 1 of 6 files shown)


dorinhogea:
for https://github.com/bloomberg/comdb2/pull/5784

riverszhang89:
@dorinhogea Thank you!

riverszhang89:
How about making it a day in advance (since the smallest granularity is daily). Otherwise if the schema change takes longer than a hour, the current shard will have date which would've belonged to the next rollout. Changing it to a day makes it a lot less likely to happen.

dorinhogea:
There is no good value here; we had collapses that take days.  Less then 24 hours let use actually test the collapse feature (since daily is the fastest possible rollout).  The retroactive partitioning is not meant to be surgical; too new rows will go in the shard 0, too old rows go to shard 1.

dorinhogea:
when id doubt, make it configurable (defaults to 24 hours)

dorinhogea:
@riverszhang89 thank you

PR #5821, opened 2026-03-18
new fdb locking locking

dorinhogea:
This is a follow-up for the reverted:
https://github.com/bloomberg/comdb2/pull/5726
here
https://github.com/bloomberg/comdb2/pull/5810

The additional commits target two issues:
1) fix deadlock during table remote discovery due to bdb lock and the new mutex
    master swing thread needs bdb write lock, waits for bdb read locks (sql threads to go away)
    -> sql thread doing table discovery using cdb2api runs recover deadlock to free bdb read lock while it has tables_mtx, and waits for master swing thread to finish 
    -> sql threads finishing free the table locks, which require table_mtx lock, so block waiting for sql thread doing discovery
    Fix is to not run cdb2api while tables_mtx is acquired;  the trade-off is concurrent threads will race to discover same table duplicating effort.
   
 2) fix never completing sql threads due to a new regression in remote table versioning; blocked threads spew "Remote table ... version 0, ..."

### DIFF
--- a/db/comdb2.h
+++ b/db/comdb2.h
@@ -3707,7 +3707,7 @@ int is_tablename_queue(const char *);
 
 int rename_table_options(void *tran, struct dbtable *db, const char *newname);
 
-int comdb2_get_verify_remote_schemas(void);
+int comdb2_get_verify_remote_schemas(struct sqlclntstate *clnt);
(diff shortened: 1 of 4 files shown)


dorinhogea:
@markhannum thank you!

PR #5833, opened 2026-03-23
Disable cache for physrep metadb

markhannum:
We saw an increased number of connections against the physrep metadb with the new software.  This PR disables handle/connection caching for the physrep metadb.

### DIFF
--- a/cdb2api/cdb2api.h
+++ b/cdb2api/cdb2api.h
@@ -43,7 +43,10 @@ enum cdb2_hndl_alloc_flags {
     CDB2_REQUIRE_FASTSQL = 512,
     CDB2_MASTER = 1024,
     CDB2_ALLOW_INCOHERENT = 2048,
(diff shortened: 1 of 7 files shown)


cdb2api/cdb2api.h
```diff
@@ -43,7 +43,8 @@ enum cdb2_hndl_alloc_flags {
     CDB2_REQUIRE_FASTSQL = 512,
     CDB2_MASTER = 1024,
     CDB2_ALLOW_INCOHERENT = 2048,
-    CDB2_SET_TAGGED = 4096
+    CDB2_SET_TAGGED = 4096,
+    CDB2_DISABLE_LOCAL_CACHE = 8192,
```
chands10, line 48: only include this in cdb2api server for now?

markhannum: Sounds good

PR #5841, opened 2026-03-28
Analyze timepart inversion

markhannum:
This PR demonstrates a lock-inversion which occurred as a result of timepart_analyze_partition maintaining the views_lk while analyzing each partition.  Analyze will later block on sql which requires the schema_lk.  This can deadlock with any flavor of _view_cron_phase, all of which acquire the schema_lk, then the views_lk.  The fix was change analyze_timepart_partition to copy out shard names, and then release the views-lk before running analyze.


roborivers (changes requested):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: <b>0/0 tests failed</b> &#x26A0;.</p>

PR #5842, opened 2026-03-30
{183287032}: Fix REPLACE INTO to mark AUXDELETE for partial index deletes

emelialei88:
## Problem

When `INSERT OR REPLACE INTO` encounters a uniqueness conflict on an index, the VDBE generates inline conflict resolution that deletes the conflicting row. For example, when a conflict is found on `PRIMARY_KEY`:

```
 27 [     NoConflict]: If cursor [2] does not contain R11, then jump to 38
 ...
 35 [      IdxDelete]: Delete key in R18..R20 from index "COLLECTION_IS_ACTIVE" using cursor [1]
 36 [         Delete]: Delete current record from cursor [0] on table "t1"
 37 [         Delete]: Delete current record from cursor [2] on index "PRIMARY_KEY"
```

Line 35 uses `IdxDelete` because cursor [1] is not positioned on the old row's entry — it must look up the key by value. Line 37 uses `Delete` (positional) because `NoConflict` at line 27 already positioned cursor [2] on the conflicting entry.

The problem is that line 36 (the data row delete) is **not** marked with `OPFLAG_AUXDELETE`. In `delete.c:825`, `AUXDELETE` is only set when `eMode != ONEPASS_OFF` or `iIdxNoSeek >= 0 && iIdxNoSeek != iDataCur`. For REPLACE INTO conflict resolution, `eMode` is always `ONEPASS_OFF` (we cannot change it to `ONEPASS_SINGLE`/`ONEPASS_MULTI` without breaking data cursor positioning).

Without `AUXDELETE`, line 36 is treated as the primary delete and triggers `osql_delrec`, which computes `del_keys` without accounting for the fact that the VDBE is handling index deletions explicitly. Line 37 (the positional `Delete` on the index cursor) does not update `del_keys`, so the `PRIMARY_KEY` deletion is never reflected in the bitmask sent to the block processor. The block processor's `del_record` then uses the incorrect `del_keys`, fails to remove the old `PRIMARY_KEY` index entry, and the subsequent `delayed_key_adds` fails with a duplicate key constraint (RC 2).

Note: The issue also reproduces without a partial index — line 35 simply wouldn't be executed in that case, but the missing `AUXDELETE` flag on line 36 still causes line 37's index deletion to be lost.

## Fix

Mark the data delete (line 36) with `OPFLAG_AUXDELETE` when it is part of REPLACE INTO conflict resolution and the conflicting index cursor will be deleted separately (line 37). This ensures `del_keys` is updated correctly to include all indexes whose keys are being explicitly deleted by the VDBE.

## Note

During investigation, I found that autocommit REPLACE INTO upgrades from the default SOSQL level to read committed before execution. This is necessary because the VDBE processes each row independently: for every row, it emits OP_NoConflict against each unique index to detect conflicts, deletes any conflicting rows, and then inserts the new one.

In SOSQL, cursors are opened with BDB_OPEN_REAL and cannot see their own writes. Consider:

```sql
CREATE TABLE t (a INT UNIQUE, b TEXT);
REPLACE INTO t VALUES (1, 'hello'), (1, 'world');
```

After (1, 'hello') is inserted, the subsequent OP_NoConflict for (1, 'world') seeks the index for a=1. In SOSQL, `bdb_cursor_find_merge` only searches the real btree (STEP 1) and skips the shadow btree (STEP 3, because `cur->sd` is NULL). It finds no conflict, so the VDBE skips the delete and issues a second insert with `a=1`. The master then fails with a duplicate key violation.

Upgrading to read committed opens cursors with `BDB_OPEN_BOTH_CREATE`, which enables shadow merging. The insert of `(1, 'hello')` is recorded in the shadow index tables via `osql_save_insrec` → `insert_record_indexes`, and `check_shadows` is set. When the cursor seeks `a=1` for the second row, STEP 3 of `bdb_cursor_find_merge` finds the shadow entry and returns `IX_FND`. The VDBE sees the conflict, deletes `(1, 'hello')`, and inserts `(1, 'world')` — resolving the conflict correctly.

### DIFF
--- a/sqlite/src/delete.c
+++ b/sqlite/src/delete.c
@@ -821,7 +821,11 @@ void sqlite3GenerateRowDelete(
     if( pParse->nested==0 || 0==sqlite3_stricmp(pTab->zName, "sqlite_stat1") ){
       sqlite3VdbeAppendP4(v, (char*)pTab, P4_TABLE);
     }
-    if( eMode!=ONEPASS_OFF ){
+#if defined(SQLITE_BUILDING_FOR_COMDB2)
+  if( eMode!=ONEPASS_OFF || (iIdxNoSeek>=0 && iIdxNoSeek!=iDataCur)){
+#else
+  if( eMode!=ONEPASS_OFF ){
+#endif /* defined(SQLITE_BUILDING_FOR_COMDB2) */
       sqlite3VdbeChangeP5(v, OPFLAG_AUXDELETE);
     }
     if( iIdxNoSeek>=0 && iIdxNoSeek!=iDataCur ){
--- a/tests/partial_index.test/Makefile
+++ b/tests/partial_index.test/Makefile
@@ -1,0 +1,8 @@
+ifeq ($(TESTSROOTDIR),)
+  include ../testcase.mk
+else
+  include $(TESTSROOTDIR)/testcase.mk
+endif
+ifeq ($(TEST_TIMEOUT),)
+	export TEST_TIMEOUT=3m
+endif
--- a/tests/partial_index.test/runit
+++ b/tests/partial_index.test/runit
@@ -1,0 +1,217 @@
+#!/usr/bin/env bash
+bash -n "$0" | exit 1
+
+set -e
+dbnm=$1
+
+r="cdb2sql ${CDB2_OPTIONS} $dbnm default"
+
+assert_eq() {
+    local actual="$1" expected="$2" msg="$3"
+    if [[ "$actual" != "$expected" ]]; then
+        echo "FAIL: $msg"
+        echo "  expected: $expected"
+        echo "  actual:   $actual"
+        exit 1
+    fi
+}
+
+# =====================================================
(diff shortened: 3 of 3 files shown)


roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>truncatesc_offline_generated  **quarantined**<br>consumer_non_atomic_default_consumer_generated  **quarantined**</p>

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>sc_resume_logicalsc_generated  **quarantined**<br>sc_resume <br>reco-ddlk-sql  **quarantined**<br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_transactional_rowlocks_generated  **quarantined**</p>

PR #5859, opened 2026-04-09
un-disable yast, fix tcl so path

rishitha957:
currently `make yast` wouldn't work --> cause of wrong libtclcdb2.so path - it's being added to DISABLED_TESTS


roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p>

chands10 (approved):

PR #5861, opened 2026-04-10
Reports JAVA_HOME (JDK/JRE path) as the task name

hgeraldino:
feature

Other interpreted languages (like Python) report their runtimes. This PR brings JVM applications on par with applications written in other languages (both interpreted languages and compiled ones)

Currently, the `comdb2_clientstats` table reports the calling class as the task name.

```
> bash-5.2$ cdb2sql bmskconn dev "select task, stack FROM comdb2_clientstats  where task like 'org.%' LIMIT 1"
(task='org.jdbi.v3.core.statement.SqlLoggerUtil', stack='org.jdbi.v3.core.statement.SqlLoggerUtil.wrap(SqlLoggerUtil.java:32) 
org.jdbi.v3.core.statement.SqlStatement.internalExecute(SqlStatement.java:1819) 
org.jdbi.v3.core.result.ResultProducers.lambda$createResultBearing$4(ResultProducers.java:110) 
...
...
```

With this change, the task name will report the JDK/JRE used to launch the task, while the calling class will still be the first entry on the stack.

Helps determine which task/runtime is being used by the calling process.

### DIFF
--- a/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2ClientInfo.java
+++ b/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2ClientInfo.java
@@ -51,8 +51,15 @@ public class Comdb2ClientInfo {
         driverName = name;
         driverVersion = version;
     }
-    
+
+    public static String getJavaHome() {
+        return System.getProperty("java.home");
+    }
+
+    @Deprecated
     public static String getCallerClass() {
+        logger.warn("getCallerClass() is deprecated and will be removed in a future version");
+
         String pkg = Comdb2ClientInfo.class.getPackage().getName() + ".";
         StackTraceElement[] stes = Thread.currentThread().getStackTrace();
         for (int i = 1, len = stes.length; i < len; ++i) {
--- a/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2ClientInfo.java
+++ b/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2ClientInfo.java
@@ -56,20 +56,6 @@ public class Comdb2ClientInfo {
         return System.getProperty("java.home");
     }
 
-    @Deprecated
-    public static String getCallerClass() {
-        logger.warn("getCallerClass() is deprecated and will be removed in a future version");
-
-        String pkg = Comdb2ClientInfo.class.getPackage().getName() + ".";
-        StackTraceElement[] stes = Thread.currentThread().getStackTrace();
-        for (int i = 1, len = stes.length; i < len; ++i) {
-            StackTraceElement ste = stes[i];
-            if (!ste.getClassName().startsWith(pkg))
-                return ste.getClassName();
-        }
-        return "Unknown Java class";
-    }
-
     public static String getCallStack(int layers) {
         String pkg = Comdb2ClientInfo.class.getPackage().getName() + ".";
         StringBuilder sb = new StringBuilder("");
--- a/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
+++ b/cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2Handle.java
@@ -691,7 +691,7 @@ public class Comdb2Handle extends AbstractConnection {
 
         if (!sentClientInfo) {
             sqlQuery.cinfo = new Cdb2ClientInfo();
-            sqlQuery.cinfo.argv0 = Comdb2ClientInfo.getCallerClass();
+            sqlQuery.cinfo.argv0 = Comdb2ClientInfo.getJavaHome();
             sqlQuery.cinfo.stack = Comdb2ClientInfo.getCallStack(32);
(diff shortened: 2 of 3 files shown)


cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2ClientInfo.java
```diff
@@ -51,8 +51,15 @@ public class Comdb2ClientInfo {
...
     }
-    
+
+    public static String getJavaHome() {
+        return System.getProperty("java.home");
+    }
+
+    @Deprecated
     public static String getCallerClass() {
+        logger.warn("getCallerClass() is deprecated and will be removed in a future version");
```
riverszhang89: Do we need to print it every time this function is called?

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2ClientInfo.java
```diff
@@ -51,8 +51,15 @@ public class Comdb2ClientInfo {
         driverName = name;
         driverVersion = version;
     }
-    
+
+    public static String getJavaHome() {
+        return System.getProperty("java.home");
```
riverszhang89, line 56: Looks good. I'd add a test [here](https://github.com/bloomberg/comdb2/blob/main/cdb2jdbc/src/test/java/com/bloomberg/comdb2/jdbc/ClientInfoTest.java#L20), too.

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>logfill [db unavailable at finish] **quarantined**<br>sc_truncate [db unavailable at finish]<br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2ClientInfo.java
akshatsikarwar: +1

hgeraldino: Just wanted to note that this function is no longer called anywhere in this codebase.

I don't expect clients to be calling this function explicitly (it is not part of the JDBC API), this could've been a private or package-private method. If there's any client who is calling this function for any reason (can't think of one), then it is fine for them to see  this warning IMO. 

I don't think it's worth adding extra complexity to avoid additional logging for something like this, but if you strongly feel like we should, I'm happy to refactor it.

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2ClientInfo.java
hgeraldino: I haven't figured out how to run tests, but also the value for this property will be different on each environment

akshatsikarwar: Follow the build instructions from `README.md`. To build test-tools: `make test-tools all` (or ninja if you use that.)

To run the java tests: `make cdb2jdbc`

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2ClientInfo.java
akshatsikarwar: IMO it  wouldn't be very complex to make it log just once.

riverszhang89: If it's not called at all, we can remove it.

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2ClientInfo.java
riverszhang89: I'd compare the `task` column of the `comdb2_clientstats` table with `$JAVA_HOME`.

cdb2jdbc/src/main/java/com/bloomberg/comdb2/jdbc/Comdb2ClientInfo.java
akshatsikarwar: OK - Lets just remove it then

hgeraldino: removed

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>logfill [db unavailable at finish] **quarantined**<br>queuedb_rollover  **quarantined**<br>cldeadlock <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>api_tst <br>reco-ddlk-sql [timeout] **quarantined**</p>

akshatsikarwar (approved):

PR #5943, opened 2026-05-15
{184231472}: Fixing time_t arithmetic overflow for cnonce

riverszhang89:
This patch also adds protection for clock going backwards, which I think is not that uncommon.

### DIFF
--- a/cdb2api/cdb2api.c
+++ b/cdb2api/cdb2api.c
@@ -5288,26 +5288,26 @@ static int next_cnonce(cdb2_hndl_tp *hndl)
        4. Otherwise, return an error. */
 
     int rc;
-    struct timeval tv;
+    struct timespec ts;
     uint64_t cnt, seq, tm, now;
     cnonce_t *c;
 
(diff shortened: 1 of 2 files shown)


cdb2api/cdb2api.c
```diff
@@ -5288,26 +5288,26 @@ static int next_cnonce(cdb2_hndl_tp *hndl)
...
-    rc = gettimeofday(&tv, NULL);
+    rc = clock_gettime(CLOCK_MONOTONIC, &ts);
     if (rc != 0)
         return rc;
     c = &hndl->cnonce;
     seq = c->seq;
     tm = (seq & TIME_MASK) >> CNT_BITS;
-    now = tv.tv_sec * 1000000 + tv.tv_usec;
+    now = (uint64_t)ts.tv_sec * 1000000 + ts.tv_nsec / 1000;
     if (now == tm) {
```
mohitkhullar: just to be safe.. handle if (now > tm) and use this logic for else (which includes now == tm)?

PR #5961, opened 2026-05-22
fix luatests, rm

rishitha957:
had to get pos again, as sqlinterfaces.c:3459 doesn't set it, also newsql_param_value, (param_value func) that set p.pos, returns err, before it's set .. so refetched 

+ one thing to note is that sqlite3_bind_parameter_index is 1 based index, so changed rm expected output too ..

### DIFF
--- a/db/sqlinterfaces.c
+++ b/db/sqlinterfaces.c
@@ -3564,7 +3564,7 @@ carray_err:     *err = sqlite3_mprintf("carray_bind: invalid param:%s count:%d t
         }
     }
 out:if (rc) {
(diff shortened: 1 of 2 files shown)


roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

mponomar (approved):

PR #6003, opened 2026-06-02
Fdb push for read and writes need to pass CDB2_REQUIRE_FASTSQL if request is comdb2api

dorinhogea:
^title

### DIFF
--- a/db/fdb_push.c
+++ b/db/fdb_push.c
@@ -447,8 +447,11 @@ int handle_fdb_push(sqlclntstate *clnt, struct errstat *err)
     int first_row = 1;
     int rc = 0, irc;
     int client_redir;
+    int cdb2open_flags = clnt->features.require_fastsql ? CDB2_REQUIRE_FASTSQL : 0;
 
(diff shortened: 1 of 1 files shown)


mponomar (approved):

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>recover_deadlock <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

dorinhogea:
NOTE: sql remote write path is affected because we do not forward the CDB2_REQUIRE_FASTSQL flag to remote node.
sql remote read path returns sqlite rows, skipping conversion from sqlite to comdb2; the last part is done by the source server, so we do not need the flag in this situation (and the read code path was never broken).  It might still be useful to know if the originator was legacy or cdb2api, so we keep the flag set.

dorinhogea:
@mponomar thank you

PR #6059, opened 2026-07-02
reqlog: log the client rcode

riverszhang89:
reqlog logs the same `SQLITE_ABORT` return code on many failure cases, such as FK violation, null-key and verify-error. This patch ensures we log the actual client return code of the failure.


mohitkhullar (approved):
lgtm

riverszhang89:
@mohitkhullar Thanks for reviewing!

PR #6064, opened 2026-07-07
{184305174}: Close displaced DB_AM_RECOVER handles in __ufid_add_dbp

chands10:
When __ufid_add_dbp is called for a file that already has an entry in the ufid-hash, the existing handle is displaced by the new one. The previous code cleared the added_to_ufid flag on the old handle but did not close it, causing a file descriptor leak.

Root Cause:
- During replication, dbreg_register log records cause __dbreg_do_open to be called, creating a new DB_AM_RECOVER handle for the file
- __ufid_add_dbp replaces the previous handle in the ufid-hash but only clears its added_to_ufid flag without closing it
- The displaced handle remains in gbl_db_open_list holding an open file descriptor
- Each subsequent dbreg_register for the same file displaces and leaks another handle
- When delfiles deletes old file versions, these leaked handles keep the inodes alive, causing files to appear as "(deleted)" in lsof

Fix:
- When displacing an existing DB_AM_RECOVER handle, update the ufid-hash to point to the new handle first, then close the old handle
- The close removes the old handle from gbl_db_open_list, releasing the file descriptor

### DIFF
--- a/berkdb/dbreg/dbreg_rec.c
+++ b/berkdb/dbreg/dbreg_rec.c
@@ -61,6 +61,8 @@ static const char revid[] = "$Id: dbreg_rec.c,v 11.120 2003/10/27 15:54:31 sue E
 static int __dbreg_open_file __P((DB_ENV *,
     DB_TXN *, __dbreg_register_args *, void *));
 
+int gbl_test_force_dbreg_reopen = 0;
+
 int
 __dbreg_register_print(DB_ENV *dbenv, DBT *dbtp, DB_LSN *lsnp, db_recops notused2, void *notused3);
 
@@ -424,6 +426,12 @@ __dbreg_open_file(dbenv, txn, argp, info)
 			if (F_ISSET(dbp, DB_AM_RECOVER))
 				__db_close(dbp, NULL, DB_NOSYNC);
 
+			goto reopen;
+		}
+
+		if (gbl_test_force_dbreg_reopen) {
+			MUTEX_THREAD_UNLOCK(dbenv, lp->mutexp);
+			(void)__dbreg_revoke_id(dbp, 0, DB_LOGFILEID_INVALID);
 			goto reopen;
 		}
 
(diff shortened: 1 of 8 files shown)


chands10:
The DB_AM_RECOVER handles get leaked — not the files themselves directly. The chain is:

  1. __dbreg_do_open creates a DB_AM_RECOVER handle for a file and adds it to the ufid-hash via __ufid_add_dbp
  2. Later, a regular handle opens the same file (e.g., via bdb_open_more_tran during schema change reload)
  3. __ufid_add_dbp finds the existing entry, replaces ufid->dbp with the new handle, clears added_to_ufid on the old handle — but never closes the old handle
  4. The old DB_AM_RECOVER handle is now orphaned: it's not in the ufid-hash anymore, but it's still in gbl_db_open_list and still holds an open file descriptor

  The file itself is fine — it's still on disk, still referenced by the new handle. The problem is the orphaned handle holding a duplicate file descriptor. Later, when delfiles removes old file versions and unlinks the files from disk, the orphaned handle's file descriptor keeps
  the inode alive. That's what shows up as (deleted) in lsof — the directory entry is gone but the inode can't be freed because a leaked handle still has it open.

roborivers (changes requested):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: <b>4/685 tests failed</b> &#x26A0;.</p><p>The first 10 failing tests are: <br>truncatesc_offline_generated [db unavailable at finish] **quarantined**<br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>phys_rep <br>skipscan  **quarantined**<br>sc_downgrade [timeout] **quarantined**<br>phys_rep_tiered [timeout]<br>phys_rep_tiered_nosource_generated [timeout]<br>phys_rep_tiered_firstfile_generated [timeout]<br>truncatesc [timeout] **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

PR #6085, opened 2026-07-26
Fingerprint runtime stats

markhannum:
# Attribute bufferpool page-ins to SQL fingerprints

## Summary
Adds per-fingerprint runtime I/O accounting. For each SQL fingerprint, comdb2 now tracks how many bufferpool page-ins its execution caused and how many of those required an actual disk read, and exposes them through the `comdb2_fingerprints` system table.

## Motivation
Fingerprints already capture query shape and CPU/timing cost, but there was no way to see which query shapes drive bufferpool traffic or physical disk I/O. This lets you answer "which queries are causing our page-ins / disk reads?" straight from a system table, with no per-query tracing.

## What it does
- Adds a berkdb-layer hash of per-fingerprint counters (`n_pagein_read`, `n_pagein_read_io`), keyed by the 16-byte fingerprint and kept separate from the existing `gbl_fingerprint_hash`.
- At statement prepare, points the executing thread's TLS at the current fingerprint's counter entry (`get_prepared_stmt` → `bdb_fingerprint_rtstats_set`); clears it when the statement finishes (`sqlite_done` and the Lua finalize paths).
- In the bufferpool, every successful `__memp_fget_internal()` bumps the current thread's counters, incrementing the disk-I/O counter only when the fetch actually read from disk (cache miss).
- Page-ins with no armed fingerprint (recovery, replication apply, or fingerprinting disabled) accumulate in a shared NO-FINGERPRINT bucket rather than being dropped.

## Interface
`comdb2_fingerprints` gains two columns:

| column | meaning |
| --- | --- |
| `total_pagein_read` | cumulative bufferpool page-ins (cache hit + miss) attributed to the fingerprint |
| `total_pagein_read_io` | subset of the above that required an actual disk read |

Gated by the existing `fingerprint_queries` switch (default on).

## Notes / limitations
- Attribution assumes the preparing thread also executes the statement — true for the appsock path today. For parallel (dohsql) queries, page-in I/O is attributed to each shard's sub-query fingerprint rather than rolled up to the parent UNION.
- Counter entries live for the lifetime of the process (never evicted).

## Testing
- `tests/fingerprints.test` — verifies the page-in counters via `comdb2_fingerprints`, and that counters stay separate across Lua inner statements.


roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>0/0 tests failed</b> &#x26A0;.</p>

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>noresetgen <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

roborivers (approved):

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>ssl_san <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>ssl_set_cmd <br>ssl_prefer <br>ssl_dbname <br>sc_downgrade [timeout] **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**<br>sql_logfill_autodisable [timeout]</p>

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**<br>truncatesc_offline_generated [timeout] **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>sc_resume <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>sc_resume_logicalsc_generated  **quarantined**<br>noresetgen <br>reco-ddlk-sql  **quarantined**<br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**</p>

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>queuedb_rollover_noroll1_generated  **quarantined**<br>noresetgen <br>ssl_san <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>ssl_set_cmd <br>ssl_prefer <br>ssl_dbname <br>sc_downgrade [timeout] **quarantined**</p>

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**</p>

berkdb/build/db.h
```diff
@@ -3128,6 +3128,12 @@ struct berkdb_thread_stats *bb_berkdb_get_process_stats(void);
 void bb_berkdb_thread_stats_init(void);
 void bb_berkdb_thread_stats_reset(void);
 
+void bb_fingerprint_rtstats_init(void);
+void bb_fingerprint_rtstats_set(const unsigned char *fingerprint, size_t fplen, int has_main_entry);
+void bb_fingerprint_rtstats_clear(void);
+int bb_fingerprint_rtstats_get(const unsigned char *fingerprint, size_t fplen,
+    uint64_t *n_pagein_read, uint64_t *n_pagein_read_io);
```
SChakravorti21: Can we name these `bb_berkdb_...` to be consistent with the rest of the functions in the file? I read that as "BB additions to BerkDB", correct me if I misunderstood.

berkdb/mp/mp_fget.c
```diff
@@ -312,6 +313,9 @@ __memp_fget_internal(dbmfp, pgnoaddr, flags, addrp, did_io)
 		++mfp->stat.st_map;
 		if (gbl_bb_berkdb_enable_memp_timing)
 			bb_memp_hit(start_time_us);
+		/* mmap fast-path: page is served from the mapping without a
+		 * cache fetch, so it never does fetch-time disk I/O. */
+		bb_fingerprint_rtstats_bump_pagein(0);
```
SChakravorti21: I am confused - how do we know that reading from an mmap'ed region doesn't cause disk I/O? Or is it that we can't know? Or did I misunderstand what is happening here?

berkdb/mp/mp_fget.c
```diff
@@ -945,9 +951,12 @@ __memp_read_recovery_pages(dbmfp)
...
 		}
 
-		/* Read this page into the bufferpool. */
+		/* Read this page into the bufferpool.  Pass a real did_io so the
+		 * page-in I/O is attributed (to the NO-FINGERPRINT bucket, since
+		 * no fingerprint TLS is set during recovery). */
+		did_io = 0;
 		if (0 == __memp_fget_internal(dbmfp, &inpg,
-			DB_MPOOL_RECP, &fpage, NULL))
+			DB_MPOOL_RECP, &fpage, &did_io))
```
SChakravorti21: Do we care about I/O caused by recovery? Also I might be wrong but looks like we don't do anything with `did_io` after this line..

lua/sp.c
```diff
@@ -1252,6 +1254,7 @@ static struct sp_tmptbl *create_temp_table(Lua lua, const char **name)
     }
     lua_end_step(sp->clnt, sp, stmt);
     set_tmptbl(NULL);
+    bdb_fingerprint_rtstats_clear();
     sqlite3_finalize(stmt);
```
SChakravorti21: Could we combine the `bdb_fingerprint_rtstats_clear()` and `sqlite3_finalize(stmt)` calls into a helper function like `lua_finalize_sql`? It could help avoid missing this cleanup step if a new code path requires it in the future.

db/sqlinterfaces.c
```diff
@@ -3368,11 +3368,17 @@ static int get_prepared_stmt_int(struct sqlthdstate *thd,
...
+            t = prepare_fingerprint(clnt, rec, fingerprint, flags);
         reqlog_set_fingerprint(thd->logger, (const char *)fingerprint, FINGERPRINTSZ);
 
         sqlite3_resetclock(rec->stmt);
         thr_set_current_sql(rec->sql);
+
+        /* t is this fingerprint's gbl_fingerprint_hash entry (NULL until its
+         * first execution completes); pass its presence as has_main_entry. */
+        if (gbl_fingerprint_queries)
+            bdb_fingerprint_rtstats_set(clnt->work.aFingerprint, FINGERPRINTSZ, t != NULL);
```
SChakravorti21, line 3383: Do we need to check for `gbl_fingerprint_queries`? Doesn't seem like the surrounding code is checking it, and it seems to be hardcoded in the build with no way to change it..

berkdb/mp/mp_fingerprint_rtstats.c
```diff
@@ -0,0 +1,193 @@
...
+ * NO-FINGERPRINT default). Atomic adds because many threads can share the
+ * same fingerprint's stat entry concurrently.
+ */
+void
+bb_fingerprint_rtstats_bump_pagein(int did_io)
+{
+	struct fingerprint_rtstats *t = NULL;
+
+	if (fingerprint_rtstats_inited)
+		t = pthread_getspecific(fingerprint_rtstats_key);
```
SChakravorti21, line 150: We have wrappers for all the other `Pthread...` funcs, can we make one for this too while we're here?

SChakravorti21 (commented):
I guess this assumes that whatever thread is calling `get_prepared_stmt` will also be the one executing it (the work won't be handed off to another thread). I suppose that's true today and unlikely to change, but were any options considered that don't involve thread-locals?

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>sc_resume_logicalsc_generated  **quarantined**<br>sp_snapshot_generated <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

berkdb/build/db.h
markhannum: Done — renamed all five to `bb_berkdb_fingerprint_rtstats_*` to match the `bb_berkdb_*` convention. Also moved `..._bump_pagein` into `db.h` with the others (it was the odd one out, forward-declared via a local `extern` in `mp_fget.c`).

berkdb/mp/mp_fget.c
markhannum: `did_io` *is* used, just by reference: `__memp_fget_internal` writes `*did_io=1` on a real read and consumes it to attribute the page-in before returning, so the local isn't read in the caller — it's the mechanism the callee uses. And yes, we do want recovery-page I/O (this is torn-page protection, not crash recovery) counted toward total I/O; it lands in the NO-FINGERPRINT bucket since no fingerprint TLS is armed here. Clarified the comment.

berkdb/mp/mp_fget.c
markhannum: Good question — we can't know here. In practice it's moot: our b-tree files are never mmap'd (they're read-write and register pgin/pgout, so `MP_CAN_MMAP` is always cleared in `__memp_fopen`), so this branch is dead for our data files. And even if a file were mapped, the page is served straight from the mapping with no fetch-time read; any real disk I/O would happen lazily via a page fault on later dereference, which is invisible here — so we record it with `did_io=0`. Updated the comment to say this.

lua/sp.c
markhannum: Done — added `lua_finalize_sql(stmt)` (clear + finalize) and switched the 6 plain sites to it. Left `donate_stmt` explicit since its `clear()` has to run before the cache-vs-finalize branch.

db/sqlinterfaces.c
markhannum: It's actually a runtime switch, not hardcoded — `register_int_switch("fingerprint_queries", …)` in `comdb2.c:5533` (default on), flippable at runtime. And the same function already guards on it a bit above at line 3282. I did spot an asymmetry, though: the paired `clear()` in `sqlite_done` was unguarded, so I added the guard there too.

berkdb/mp/mp_fingerprint_rtstats.c
markhannum: Left this one raw on purpose. The `Pthread_*` wrappers (`WRAP_SYSFUNC`) abort on any nonzero return and discard the value — but here a NULL return is normal and load-bearing: it's how we redirect accounting to the NO-FINGERPRINT bucket. `util/thread_util.c` sets the same precedent (raw `pthread_getspecific`, wrapped abort-checking calls).

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>sc_resume_logicalsc_generated  **quarantined**<br>ssl_san <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>ssl_set_cmd <br>ssl_prefer <br>ssl_dbname <br>sc_downgrade [timeout] **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

markhannum:
> I guess this assumes that whatever thread is calling `get_prepared_stmt` will also be the one executing it (the work won't be handed off to another thread). I suppose that's true today and unlikely to change, but were any options considered that don't involve thread-locals?

Right that it assumes prepare and execute share a thread — which holds for the normal appsock path, since a thdpool work item runs prepare → step → clear start-to-finish on one worker. The safety valve is the static NO-FINGERPRINT bucket: a thread that fetches without arming its TLS just mis-attributes there, never corrupting another fingerprint's counts. The load-bearing case is parallel/dohsql shards, where each shard prepares on its own thread and arms its own TLS. Didn't see a non-TLS option that's cleaner without threading the fingerprint through the whole cursor/page-fetch call chain, so I kept the `pthread_key_t` approach.

SChakravorti21 (approved):
Failing SSL tests are a known issue on one RoboRivers host, does not look related to this PR.

Internal build is broken (being fixed asap), but an earlier run of RoboMark tests did pass. Only change since then was renaming of `bb_fingerprint` functions -> `bb_berkdb_fingerprint`.

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>sc_resume_logicalsc_generated  **quarantined**<br>sc_resume <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**</p>

PR #6100, opened 2026-07-31
{185333845}: reject type-incompatible foreign-key constraints at sche…

chands10:
…ma-change time

comdb2 matches foreign-key columns to the referenced key positionally (not by name), converting only the first min(local, parent) columns via stag_to_stag_buf_ckey. If a positionally-corresponding pair is type-incompatible -- e.g. a swapped key order that lines an int up against a cstring -- the conversion returns -1 at runtime: every write to the child table is rejected and existing rows fail verify with 'error loading rc = -1'. Nothing caught this at schema-change time: verify_constraints_exist only checked column counts, and even that is short-circuited when prefix_foreign_keys and superset_foreign_keys are both on (the legacy default).

Add a per-column type-class compatibility check to verify_constraints_exist so such a constraint is rejected up front on add/alter, guarded by a new runtime tunable fk_constraint_type_check (default ON). Startup is unaffected: existing tables load via populate_reverse_constraints, which is not touched, so a db that already has such a constraint still boots.

Add tests/fk_type_check.test and update tests/tunables.test per AGENTS.md.

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -642,6 +642,8 @@ int gbl_fk_allow_prefix_keys = 1;
 
 int gbl_fk_allow_superset_keys = 1;
 
+int gbl_fk_constraint_type_check = 1;
+
 int gbl_update_delete_limit = 1;
 
 int verbose_deadlocks = 0;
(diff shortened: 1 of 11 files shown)


roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>noresetgen <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**</p>

PR #6132, opened 2026-08-13
Handle ix_addk rcode correctly during master swing

riverszhang89:
ERR_NOMASTER is treated as a duplicate key error, which causes upsert to misbehave during master swing: an upsert may miss its index entry, or may fail incorrectly during verify-retry.

Case 1) upsert misses its index entry

```
mydb4> create table t2 (i int primary key)$$
[create table t2 (i int primary key)] rc 0
```

Upsert receives a good rcode, but misses its index entry:
```
mydb4> insert into t2 values(3) on conflict do nothing
(rows inserted=1)
[insert into t2 values(3) on conflict do nothing] rc 0
mydb4> select * from t2
(i=3)

mydb4> select count(*) from t2 where i = 3
(count(*)=0)
```

Case 2) upsert fails incorrectly during verify-retry

Insert-on-conflict-do-nothing, still receives a dupkey error:

```
mydb4> begin
[begin] rc 0
mydb4> insert into t2 values(6) on conflict do nothing
[insert into t2 values(6) on conflict do nothing] rc 0
mydb4> insert into t2 values(6) on conflict do nothing
[insert into t2 values(6) on conflict do nothing] rc 0
mydb4> update t2 set i = sleep(1) where i = 1
[update t2 set i = sleep(1) where i = 1] rc 0
mydb4> commit
[commit] failed with rc 2 Transaction is uncommittable: Duplicate insert on key 'COMDB2_PK' in table 't2' index 0
```

Database logs the following:

```
[ERROR] Forced VERIFY-FAIL for uncommittable blocksql transaction
```

### DIFF
--- a/db/comdb2.c
+++ b/db/comdb2.c
@@ -369,6 +369,8 @@ static int gbl_db_is_exiting = 0; /* Indicates this process is exiting */
 
 int gbl_debug_omit_dta_write;
 int gbl_debug_omit_idx_write;
+int gbl_debug_ix_addk_nomaster;
+int gbl_debug_ix_addk_nomaster_skip;
 int gbl_debug_omit_blob_write;
 int gbl_debug_omit_zap_on_rebuild = 0;
 int gbl_debug_txn_sleep = 0;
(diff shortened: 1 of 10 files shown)


markhannum (approved):

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>incoherent_startup <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**</p>

PR #6136, opened 2026-08-13
Make tranlog exit aware

markhannum:
An exiting node can hang until the exit alarm fires while serving transaction-log requests from its replicants.

A tranlog request runs on an appsock-sql thread, and `clean_exit()` waits for those threads to exit in `thrman_wait_for_all()` under a 300s alarm. `sql_flush_int()` polls until `wr_buf` drains, and `pollms` is only cleared when the buffer empties or the write hard-errors. A replicant that breaks out of its fetch loop — sqllogfill and physrep both do, on gap-filled, generation change, or lock-desired — leaves the node mid-stream with nobody reading. The buffer never drains, `poll()` returns 0 every second forever, and `thrman_stop_sql_connections()` doesn't help because it shuts down the read side only.

Give up on a write that has made no progress for `exit_flush_timeout_sec` (default 10, 0 disables) while exiting. `blocked_at` is reset by every successful write, so a slow-but-still-draining client is unaffected and the clean-exit grace period for in-flight queries is preserved.

Also stop tranlog from handing the thread more work on the way out. The poll loop already checked `db_is_exiting()` but ended the scan quietly, which the client reads as an empty result and immediately re-issues. Raise an error from `tranlogFilter()` instead — sqlite treats the `xEof` return as a boolean, so an error raised from `tranlogNext()` by way of `tranlogEof()` is swallowed as an end-of-scan — and check for exit in the `xNext` paths, where errors do propagate.

Tests: `sql_logfill_exit` exits the master while replicants are actively logfilling; `phys_rep_exit` exits the source node a physrep is streaming from. Both require the node to reach `goodbye` rather than hang.

### DIFF
--- a/db/db_tunables.c
+++ b/db/db_tunables.c
@@ -105,6 +105,7 @@ extern int gbl_debug_disttxn_trace;
 extern int gbl_sparse_lockerid_map;
 extern int gbl_early;
 extern int gbl_exit_alarm_sec;
+extern int gbl_exit_flush_timeout_sec;
 extern int gbl_fdb_default_ver;
 extern int gbl_fdb_track;
 extern int gbl_fdb_track_hints;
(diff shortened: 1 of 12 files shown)


roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>cdb2jdbc <br>ssl_san <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>ssl_set_cmd <br>ssl_dbname <br>ssl_prefer <br>sc_downgrade [timeout] **quarantined**</p>

PR #1963, opened 2019-12-04
asynchronous distributed commit 

aakash10292:
To help us review your pull request, please consider providing an overview of the following:
* What is the type of the change (bug fix, feature, documentation and etc.) ?
* What are the current behavior and expected behavior, if this is a bugfix ?
* What are the steps required to reproduce the bug, if this is a bugfix ?
* What is the current behavior and new behavior, if this is a feature change or enhancement ?
* [Optional] Why is the new behavior better than the current behavior, if this is a feature change ?

PR #6144, opened 2026-08-17
Hand the replication-ack wait to a background thread (port of #1963, with measurements)

emelialei88:
Port of bloomberg/comdb2#1963 (3126 commits behind main, hand-ported rather than rebased) onto current main, plus an A/B measurement of what it actually buys.

**Why.** Block processors commit locally and then block in `trans_wait_for_seqnum_int` until every replicant acks. With the default 8 writer threads, that caps in-flight commits at 8 no matter how many clients are connected.

**What.** A dedicated seqnum-wait thread takes the wait; the block processor returns to the queue immediately. The ack still happens, so replicants stay bounded — unlike `rep-sync none`, which drops it and drove two nodes permanently incoherent inside one 5-minute run at 64 clients. Off by default (`async_dist_commit`), togglable at runtime.

Rather than carry #1963's copy of `bdb_wait_for_seqnum_from_all_int` — which predated durable-majority accounting, `gbl_2pc`, `non_durable_retry` and `set_durable_lsn` — the shared logic is factored out of main into `bdb_wait_for_seqnum_finish` and `bdb_wait_for_seqnum_mark_incoherent`.

**Measured.** 8-node btda cluster, 300s runs, single-row autocommit INSERTs, one table per client, inserts/sec:

| clients | control | waiter on | rep-sync none | ctrl spread | waiter gain |
| ------- | ------- | --------- | ------------- | ----------- | ----------- |
| 4       | 7,375   | 7,553     | 10,309        | 0.23%       | +2.4%       |
| 8       | 11,656  | 12,077    | 14,111        | 1.02%       | +3.6%       |
| 16      | 12,146  | 12,455    | 13,722        | 0.81%       | +2.5%       |
| 32      | 12,040  | 12,453    | 13,653        | 1.31%       | +3.4%       |

Scope confirmed reachable: 726,674 commits enqueued vs 726,658 rows, zero inline fallbacks.

**Risk / why this is a draft.** The gain is +2.4–3.6% and flat across client counts, so writer-thread starvation was not the binding constraint — the working hypothesis going in. Raising `async_dist_commit_max_outstanding_trans` 8→32→128 does not help either. `rep-sync none` shows 12–35% of real headroom, but this is not the lever that reaches it. Opening for the port and the numbers; I would not merge it on a 3% return for ~500 lines of concurrent code that has to be kept in step with `bdb_wait_for_seqnum_from_all_int` forever.

Four latent bugs found and fixed during the port, which would have hit the original too: `thrman_register(THRTYPE_GENERIC)` made `begin_clean_exit()` wait forever on a thread that never exits; the waiter took `BDB_READLOCK` with no bdb lock slot and killed the master with `[FATAL] bdb lock not inited in this thread`; `gbl_set_coherent_state_trace` defaults to 1, whose per-node-per-poll trace is catastrophic for a polling waiter; and the lock-desired early exit skipped the durable-majority accounting the inline path does, so a role change or recovery with commits in flight silently under-counted `distributed_commits`/`not_durable_commits` (reproduced on a 4-node cluster: 23 of 24 caught commits vanished from the counter).


roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>reco-ddlk-sql  **quarantined**<br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>tunables <br>sc_downgrade [timeout] **quarantined**</p>

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>osql_cleanup [failed with core dumped]<br>analyze_partial_index_off_generated [failed with core dumped] **quarantined**<br>analyze [failed with core dumped] **quarantined**<br>tsa <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>tunables <br>sc_downgrade [timeout] **quarantined**<br>skipscan [timeout] **quarantined**</p>

roborivers (changes requested):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: <b>harness did not run</b> - no test results were produced &#x26A0;.</p>

PR #6035, opened 2026-06-19
Recover deadlock on a cursor move

dorinhogea:
Extend lock release to non-snapshot transactional modes, by caching the data cursors associated with index cursors before lock release.

re: 185495987

PR #6150, opened 2026-08-20
tests: Fix rollover race in manual_partition

emelialei88:
### Why

`manual_partition` fails intermittently with a two-line diff — shard `$1_7DA96DB` reports `LOW/HIGH = 1/2` where `run.log.exp` wants `2147483647/2147483647`. Seen on two separate roborivers runs at different commits.

`timepart_stats()` read the partition layout before its `sleep 6` and the cron tables after it, so the expected output encoded a single snapshot straddling the epoch-1 rollover: not-yet-rolled in the layout, already-rolled in the cron tables. Only two inserts separate the counter increment from that first read, so when the cron thread wins the race the layout shows the rolled-over ranges and the test fails.

### What

Sleep once on entry to `timepart_stats()` instead of mid-function, so every read in a snapshot sees the same side of the rollover, and update the two `run.log.exp` lines accordingly. The pre-rollover state is still asserted by the earlier `timepart_stats` call, which runs before the counter is advanced.

### DIFF
--- a/tests/manual_partition.test/run.log.exp
+++ b/tests/manual_partition.test/run.log.exp
@@ -59,7 +59,7 @@ XXXXX we should see to empty partitions with no counters
 (name='Truncate', type='testpart2', epoch=10, arg1='testpart2', arg2=NULL, arg3=NULL)
 (rows inserted=1)
 (rows inserted=1)
-XXXXX we should see the data in part1, and the counter, no rollout yet
+XXXXX we should see the data in part1, the counter, and the rollout
 (a=1, alltypes_vutf8='hi')
 (a=2, alltypes_vutf8='ho')
 [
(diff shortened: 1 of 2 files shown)


dorinhogea:
addressed here:
https://github.com/bloomberg/comdb2/pull/6035

emelialei88:
Fixed in another PR

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>osql_cleanup [failed with core dumped]<br>sp_snapshot_generated <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**</p>

emelialei88:
Fixed by PR 6035

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>sc_resume_logicalsc_generated  **quarantined**<br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**<br>queuedb_rollover_noroll1_generated [timeout] **quarantined**</p>

PR #6155, opened 2026-08-25
Fix llmeta schema change object leaks

dorinhogea:
osql_finalize_scs deleted the sc_list keyed by the schema change uuid, set only for TRANLEVEL_SOSQL; key it by the osql session uuid instead.

Only kinds running through do_ddl() reach mark_schemachange_over_tran(), so stop persisting an sc seed for the others (sp, trigger, lua func) -- nothing ever removed it.  Resumed sc lists no longer mark those kinds as resuming, as they have no persisted state to pick up.

OSQL_FLAGS_SCDONE is set when a schema change is dispatched, before a downgrade can be seen by its async conversion, and was never cleared, so the abort path deleted the sc_list the new master needed to resume from.

A failed multi-table resume dropped its ireq, orphaning the schema changes that had started and stranding the sc_list, and gave up on the remaining sc lists.

re: 185631797

### DIFF
--- a/db/osqlblockproc.c
+++ b/db/osqlblockproc.c
@@ -1445,9 +1445,16 @@ int bplog_schemachange_wait(struct ireq *iq, int rc)
             }
             sc = next;
         }
-        if (rc == ERR_NOMASTER)
+        if (rc == ERR_NOMASTER) {
             iq->sc_pending = NULL;
+            /* this is resumable: don't let the abort callback delete the
+             * sc_list, the next master needs it
+             */
+            iq->osql_flags &= ~OSQL_FLAGS_SCDONE;
+        }
     }
+    /* osql_scdone_abort_callback keys its sc_list cleanup off this */
+    assert(rc != ERR_NOMASTER || iq->sc_pending == NULL);
     logmsg(LOGMSG_INFO, ">>> DDL SCHEMA CHANGE RC %d <<<\n", rc);
 
     assert(!iq->sc_locked);
@@ -1517,10 +1524,14 @@ void *resume_sc_multiddl_txn_finalize(void *p)
     struct ireq *iq = (struct ireq*)p;
     struct schema_change_type *sc;
     tran_type *parent_trans = NULL;
-    int error = 0;
-    uuid_t uuid;
-
-    comdb2uuidcpy(uuid, iq->sc_pending->uuid);
+    /* set by the caller if some of the schema changes of this transaction
+     * failed to even start; whatever did start has to be backed out
+     */
+    int error = iq->sc_should_abort;
+    /* if we are the ones going away, the schema changes that failed to start
+     * did so because of the downgrade, not because of anything wrong with them
+     */
+    int downgrade = error && get_stopsc(__func__, __LINE__);
 
     iq->sc = sc = iq->sc_pending;
     iq->sc_pending = NULL;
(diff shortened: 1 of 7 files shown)


roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**<br>truncatesc_offline_generated [timeout] **quarantined**</p>

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>sc_truncate [db unavailable at finish]<br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>unifiedcancel  **quarantined**<br>sc_downgrade [timeout] **quarantined**</p>

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

db/osqlblockproc.c
```diff
@@ -1545,7 +1558,11 @@ void *resume_sc_multiddl_txn_finalize(void *p)
         sc = iq->sc;
     }
     iq->sc_locked = 0;
-    iq->osql_flags |= OSQL_FLAGS_SCDONE;
+    /* marking scdone makes the abort path remove the sc_list from llmeta; on a
+     * downgrade we want the opposite -- leave it for the next master to resume
+     */
+    if (!downgrade)
+        iq->osql_flags |= OSQL_FLAGS_SCDONE;
```
WalidNejmi: Here, would we want to also add an else statement when it is not going to be a downgrade removing the flag like this `iq->osql_flags &= ~OSQL_FLAGS_SCDONE` ? Let's say there is a downgrade from A -> B during sc. SC_DONE = OSQL_FLAGS_SCDONE because we want to resume the schema change even during a master swing. Now let's say there is a second downgrade from B -> C, the `iq->osql_flags` could already have `OSQL_FLAGS_SCDONE` even when not intended because B did not inherit the flags of A's iq (it created its own fake one and marked SC_DONE). So here even if there is a downgrade, C's iq->osql_flags will have `OSQL_FLAGS_SCDONE`.

WalidNejmi, line 1565: Here, would we want to also add an else statement when it is not going to be a downgrade removing the flag like this iq->osql_flags &= ~OSQL_FLAGS_SCDONE ? Let's say there is a downgrade from A -> B during sc. SC_DONE != OSQL_FLAGS_SCDONE because we want to resume the schema change even during a master swing. Now let's say there is a second downgrade from B -> C, the iq->osql_flags could already have OSQL_FLAGS_SCDONE even when not intended because B did not inherit the flags of A's iq (it created its own fake one and marked SC_DONE). So here even if there is a downgrade, C's iq->osql_flags will have OSQL_FLAGS_SCDONE.

dorinhogea: osql_flags are not replicated across nodes, so iq->osql_flags;  what happens here is A will not set change osql_flags; it is possible that it is set and phase 1 failed after starting the schema change, so we want finalize to run and clear schema change and free the llmeta also so it does not resume, even if this is a downgrade.

roborivers (changes requested):
<p>Failed to merge main into this PR - resolve conflicts and re-run &#x26A0;.</p>

PR #6168, opened 2026-08-27
physrep: truncate ignored queues to their own pagesize

markhannum:
physrep: truncate ignored queues to their own pagesize

**Type:** bug fix.

**Current behavior.** A physrep truncates the queues it ignores to twice the pagesize comdb2 would pick for them today, but Berkeley uses the pagesize in the file's meta-page. When those disagree the truncate leaves a size that is not a page multiple, `__memp_fopen` fails with "file size not a multiple of the pagesize", and the database will not start until it is re-copied. On by default for legacy `.queue` files; latent longer on the queuedb path.

**New behavior.** Read the pagesize from the meta-page and never grow the file, so an ignored queue is either shrunk to a page-aligned size or left alone.

**Test.** `physrep_queue_pagesize` in comdb2-legacy-tests, with 4096 on disk against 1024 in the physrep lrl.

A copy the old code already damaged still needs re-copying: recovery opens the queue file before `open_dbs` and panics on a partial page.

### DIFF
--- a/bdb/file.c
+++ b/bdb/file.c
@@ -94,6 +94,7 @@
 #include <logmsg.h>
 
 #include <build/db_int.h>
+#include <dbinc/db_page.h>
 #include "dbinc/db_swap.h"
 
 #include <dbinc/db_am.h>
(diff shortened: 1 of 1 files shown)


roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>comdb2sys_queueodh_generated [db unavailable at finish]<br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**</p>

mponomar (approved):

PR #6169, opened 2026-08-27
cdb2_dump: don't skip records larger than the bulk buffer

emelialei88:
### 🎯 The "Why" (Intent)
`cdb2_dump` silently omitted any key/data pair bigger than its 1MB bulk-get buffer — no error, exit code 0 — so dumps of tables with large blobs were quietly incomplete.

### 🛠️ The "What" (Critical Changes)
- On `ENOMEM`, berkeley db deliberately leaves the cursor parked on the oversized pair so the caller can grow its buffer and re-read with `DB_CURRENT`. `dump()` instead resumed the loop with `DB_NEXT`, stepping straight over it. It now re-reads the parked pair before resuming.
- New `cdb2dump_bigblob` testcase: stores a 4MB blob alongside small ones and asserts `cdb2_dump` emits it. Fails on the old binary, passes on the new one.

### DIFF
--- a/tests/cdb2dump_bigblob.test/Makefile
+++ b/tests/cdb2dump_bigblob.test/Makefile
@@ -1,0 +1,8 @@
+ifeq ($(TESTSROOTDIR),)
+  include ../testcase.mk
+else
+  include $(TESTSROOTDIR)/testcase.mk
+endif
+ifeq ($(TEST_TIMEOUT),)
+	export TEST_TIMEOUT=2m
+endif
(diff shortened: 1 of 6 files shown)


mponomar (approved):
Looks good.

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>timepart_retro <br>sc_resume <br>ssl_san <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>ssl_set_cmd <br>ssl_prefer <br>ssl_dbname <br>sc_downgrade [timeout] **quarantined**<br>sql_logfill_autodisable [timeout]</p>

PR #6171, opened 2026-08-28
Logfill bounded tranlog stream

markhannum:
## Problem

A replicant running `sql_logfill` destroys one connection to the master per gap-fill, and each of those is a sockpool miss. Observed in production at roughly 890 log-requests/sec across seven replicants, sustained for hours.

## Cause

`request_logs_from_master()` asks for an unbounded stream (`NULL` stop LSN, `TRANLOG_FLAGS_BLOCK`), so it never ends on its own. logfill consumes only as far as the gap and abandons the rest — in production the master averaged 636 rows per query against a gap of a couple of records. `newsql_disconnect()` will not donate a handle whose last response was not `LAST_ROW`, and `cdb2_discard_unread_socket_data` defaults to off, so the socket is closed rather than pooled and the next request misses.

This happens on both sides of `0ed0d3f3b`: with it, logfill hangs up and reconnects explicitly; without it, `consume_previous_query()` drains at most 10 rows and then disconnects from inside `cdb2_run_statement`.

## Fix

Pass `gap_lsn` as the stop LSN. `tranlogEof()` already terminates on it, so the result set ends with `LAST_ROW`, the existing `consumed` path skips the disconnect, and the connection persists across gap-fills.

A second hunk stops `tranlogNext()` from entering the blocking poll once it has returned `maxLsn` — that loop checks only `blockLsn` and `timeout`, so otherwise the bound would trade connection churn for a stall.

## Test

`tests/sql_logfill_socket_reuse.test` drops every Nth rep message on one replicant to force out-of-order arrivals, then counts log-requests against master disconnects. A gap is an out-of-order arrival rather than simply being behind — `gap_lsn` is repdb's `waiting_lsn` — so dropping messages is what reproduces the production condition. Unbounded, disconnects track requests one-for-one; the test allows a tenth. Verified 8 consecutive passes with the fix, and confirmed failing without it.

### DIFF
--- a/db/sqllogfill.c
+++ b/db/sqllogfill.c
@@ -555,13 +555,23 @@ static int request_logs_from_master(bdb_state_type *bdb_state)
         /* Release bdb-lock before querying */
         BDB_RELLOCK();
 
+        /* Already at the gap: the bound would be at or below the start, and the
+         * empty result set that comes back reads as a stream failure below. */
+        DB_LSN start_lsn = {last_lsn.file, last_lsn.offset};
+        if (log_compare(&start_lsn, &gap_lsn) >= 0) {
+            return 1;
+        }
+
         int timeout = gbl_sql_logfill_next_timeout;
         timeout = timeout > 0 ? timeout : 10;
 
-        /* Query transaction log with BLOCK and SENTINEL flags */
+        /* Stop the stream at the gap.  An unbounded stream never ends, so we
+         * abandon it mid-result-set and cdb2api has to drop the socket rather
+         * than donate it back to sockpool. */
         rc = snprintf(sql_cmd, SQL_CMD_LEN,
-                      "select lsn, generation, payload from comdb2_transaction_logs('{%u:%u}', NULL, %d, %d)",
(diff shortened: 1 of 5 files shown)


roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>logfill [db unavailable at finish] **quarantined**<br>sc_resume_logicalsc_generated  **quarantined**<br>sp_snapshot_generated <br>comdb2sys_queueodh_generated <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**</p>

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>logfill [db unavailable at finish] **quarantined**<br>ssl_san <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>ssl_set_cmd <br>ssl_prefer <br>ssl_dbname <br>sc_downgrade [timeout] **quarantined**</p>

PR #6177, opened 2026-08-28
Keep a list of nodes allowed to cluster with us.

mponomar:
Two phases:
In first, just warn if a node is able to join today but won't be once `enforce_cluster_allow_list` is enabled, eg:

```
host cdb2machine2 clusters with us now, but the cluster allow list would reject it.  Run "allow cluster with cdb2machine2" to keep it once the list is enforced.
```

Hosts can be added via, eg `allow cluster with cdb2machine` or `allow cluster with group c2machinegroup`

`bdb add`'ed nodes are allowed.  Nodes that are allowed to join bring their `lrl`-configured list with them, and nodes on that list are also allowed to join.

### DIFF
--- a/bdb/info.c
+++ b/bdb/info.c
@@ -23,6 +23,7 @@
 
 #include <segstr.h>
 
+#include "cluster_allow_list.h"
 #include "net.h"
 #include "bdb_int.h"
 #include "locks.h"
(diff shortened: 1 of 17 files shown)


roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>sc_truncate_multiddl_generated [db unavailable at finish] **quarantined**<br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**</p>

PR #6183, opened 2026-09-01
  serial: swap rkeylen along with rkey when coalescing read ranges

akshatsikarwar:
currangearr_merge_neighbor() swaps the two ranges' rkey pointers so the
  merged range keeps the larger upper bound, but left rkeylen behind. The
  surviving range then described one buffer with another buffer's length,
  which osqlcomm.c ships to the master and glue.c compares against.

  Ranges on the same index really do differ in length: a bound from a
  partial-key probe carries the probe length, one from a cursor move the
  full found-key length.

### DIFF
--- a/db/sqlglue.c
+++ b/db/sqlglue.c
@@ -281,7 +281,6 @@ static void currangearr_merge_neighbor(CurRangeArr *arr)
     i = 1;
     int n = arr->size;
     CurRange *p, *q;
-    void *tmp;
     if (!n)
         return;
     while (i < n) {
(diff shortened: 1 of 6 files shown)


roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>timepart_retro <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**<br>truncatesc_offline_generated [timeout] **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

WalidNejmi (approved):

mponomar (approved):

PR #6199, opened 2026-09-08
{185343794}: analyze new indexes during schema change

chands10:
A newly built index starts with default statistics, so the query planner can pick poor plans -- skipping the index entirely, choosing a bad join order, or misjudging skew and range selectivity -- until some later ANALYZE happens to run. Today a schema change only *invalidates* statistics (scdone_sc_analyze bumps gbl_analyze_gen); it never collects any.

Collect them during the schema change instead, against the pre-commit newdb, and commit before finalize. Replication is ordered, so the stats transaction lands on every replicant ahead of the schema change itself: no node can observe the new index without its statistics.

This covers CREATE INDEX, ALTER ADD INDEX, REBUILD INDEX and altered index definitions -- all of them reach do_alter_table, and ix_plan[i] == -1 identifies the indexes that were actually built. Only those are analyzed. Reused indexes hold no data in newdb (their files are adopted by pointer swap at finalize), so analyzing the whole table would sample them empty and strand their existing, still-valid statistics in cdb2.<tbl>.sav.

Mechanics:

  - clnt->custom_rootpages gives sqlite newdb's schema. That is not enough on its own: get_sqlite_db() resolves storage by name through thedb->db_hash, and newdb is not in that hash until finalize, so cursors would resolve to the live table. clnt->custom_dbtable closes that gap.

  - prepare_engine reconciles rootpages per query, so a pool thread never leaves a custom view behind for the next client. The analyze also runs on a dedicated "scanalyze" pool so client-facing threads are not touched.

  - Index names are generated without the ".NEW." tag prefix, so the rows are keyed the way the planner will later look them up. Renaming afterwards is not possible: the rows belong to the same osql transaction that inserted them.

  - The schema-change-in-progress guards in bdb_summarize_table() and cursor_move_preprop() are bypassed for this path only. They exist to stop an unrelated analyze from competing with a schema change; here the schema change is scanning the table it just built.

  - flush_db() before sampling: the sampler reads index files directly off disk, so pages the schema change just wrote must be flushed or it samples an empty index and writes empty statistics.

Statistics never fail a schema change. If another analyze holds the global flag, or anything errors, it is logged and the schema change proceeds. Gated by the analyze_new_indexes tunable, default on.

Tested standalone and on a four-node cluster, including that the statistics are present on every node as soon as the schema change returns.


roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>timepart_retro <br>sc_resume <br>yast_stat4scan_generated <br>yast <br>sc_analyze <br>tsa <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>comdb2expert_snapshot_generated <br>tunables <br>sc_downgrade [timeout] **quarantined**</p>

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>comdb2sys_queueodh_generated [db unavailable at finish]<br>sc_resume_logicalsc_generated  **quarantined**<br>tsa <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>comdb2expert_snapshot_generated <br>sc_downgrade [timeout] **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

PR #6203, opened 2026-09-10
Fix bound parameters for fdb push writes inside transactions

dorinhogea:
A pushed write in a client transaction now sends only its own bound parameters. Previously every statement after the first also carried the parameters of the preceding statements, so the remote rejected it with "parameters in stmt:N parameters provided:M" and the transaction failed.
re:  181155679

### DIFF
--- a/db/fdb_push.c
+++ b/db/fdb_push.c
@@ -458,6 +458,11 @@ static int _run_statement(sqlclntstate *clnt, cdb2_hndl_tp *hndl, struct errstat
     if (!dts)
         return -1;
 
+    /* hndl is reused across the statements of a client transaction, and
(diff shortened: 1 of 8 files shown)


roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>tsa <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**</p>

mponomar (approved):

dorinhogea:
@mponomar thank you

PR #6205, opened 2026-09-10
{178575361} Fill autoinc value for expression fields in keys

chands10:
The sequence value for an autoinc column is only assigned on the master, by set_master_columns().  A replicant forming index keys therefore evaluates every index expression against a NULL placeholder, so any expression referencing such a column ships a key holding NULL.

Re-evaluate those expression fields in add_record(), right after the sequence is assigned, rewriting iq->idxInsert[] in place.  Every later create_key_from_ireq() caller - including delayed_key_adds(), which re-reads the record by genid and has no blobs in scope - then sees a correct key without needing the blobs the expression may reference.

### DIFF
--- a/db/indices.c
+++ b/db/indices.c
@@ -326,6 +326,56 @@ static inline void append_genid_to_key(dtikey_t *ditk, int ixkeylen)
     memcpy(&ditk->ixkey[ixkeylen], &ditk->genid, sizeof(ditk->genid));
 }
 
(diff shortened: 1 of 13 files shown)


roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>tsa <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>unifiedcancel  **quarantined**<br>sc_downgrade [timeout] **quarantined**</p>

markhannum (approved):

PR #6214, opened 2026-09-15
sockbplog: keep the appsock buffer alive until the reply is written

akshatsikarwar:
The appsock framework owned the connection buffer, so it was closed as soon as handle_sockbplog_request() returned or the peer dropped -- while a block processor could still be writing the dispatched session's reply over it (sess->target->sb).

Wrap the buffer in a refcounted bplog_sock instead: the appsock holds a reference for the connection, init_bplog_socket_master() takes one per session it dispatches, and whichever finishes last calls close_appsock(). The handler sets *arg->keepsocket so the framework stops closing the buffer underneath them.

Dispatch also leaves the session unsafe to touch: handle_buf_sorese() drops our client ref, and closes the session outright when it fails. osql_sess_rcvop_socket() and osqlcomm_bplog_socket() now report that through sess_gone so the caller returns without inspecting sess.

osql_sess_close() waits on a condvar for the last client reference rather than polling the counter -- the count reads zero a moment before the last osql_sess_remclient() has released the mutex.  For the same reason cleanup_entry() takes and drops entry->mtx before destroying it, since osql_chkboard_sqlsession_rc() can still be inside that critical section and destroying a held mutex returns EBUSY.

Also break the sockbplog loop on a bad request instead of spinning on it, and fail a bplog that ends with neither a done message nor an error. Covered by tests/sockbplog_reply_race.test.


akshatsikarwar:
Perhaps it is time to just get rid of OSQL_OVER_SOCKET altogether.

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>sc_truncate_multiddl_generated [db unavailable at finish] **quarantined**<br>sc_truncate [db unavailable at finish]<br>timepart_retro <br>sc_resume_logicalsc_generated  **quarantined**<br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>sc_downgrade [timeout] **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

akshatsikarwar:
removing sock-bplog here: https://github.com/bloomberg/comdb2/pull/6216

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>ssl_san <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>ssl_set_cmd <br>ssl_dbname <br>ssl_prefer <br>sc_downgrade [timeout] **quarantined**<br>reco-ddlk-sql [timeout] **quarantined**</p>

dorinhogea:
maybe fix format? 
there are a few issues with the feature I will mention here for the record, but nothing new related to these changes, and if we are gonna remove the feature, maybe not worth pursuing:
1. Raw rc returned to the appsock framework is now a latent UAF (plugins/sockbplog/sockbplog.c:206). handle_sockbplog_request returns whatever error code bubbled up, but the framework interprets the return as APPSOCK_RETURN_*, and CONT == 2. If any path ever yields 2, thd_appsock_int loops and calls cdb2buf_gets(…, sb) after bplog_sock_deref() may have closed the buffer. Today no path returns 2 (codes are 1, −1, ENOMEM, ERR_BADREQ=199, ERR_REJECTED=188, …), and before this commit the same value would only have caused a harmless extra read since the buffer was still alive. Worth making explicit: return rc ? APPSOCK_RETURN_ERR : APPSOCK_RETURN_OK;.

2. Dispatch-failure leak now also leaks the fd (db/osqlsession.c:679-684). If osql_bplog_build_sorese_req() fails, handle_buf_sorese logs "bug in code" and returns rc, which is still 0 — no osql_sess_close. The session leak is pre-existing, but the leaked session now pins a bplog_sock reference, so the appsock fd and its active_appsock_conns slot leak too. And with sess_gone = 1 the caller returns 0 and keeps looping. It's an OOM-only path, but closing the session there is a one-liner and it's exactly the failure mode the test README calls out as uncovered.

3. The "bplog ended without a done message" case is unreachable (plugins/sockbplog/sockbplog.c:101-102). osqlcomm_bplog_socket's while (!is_msg_done) can only exit with rc != 0, or with is_msg_done set — and is_msg_done implies either the cancel path (is_cancelled, checked first) or the dispatch path (sess_gone, returned earlier). Harmless as defence, but see below on the description.

4. Nit — init_bplog_net() (db/osqlsqlnet.c:40) clears target->sb but not the new target->sbref. Safe today (master-side targets are calloc'd and replicant-side ones never set it), but it's the one place that's supposed to reset a target.

dorinhogea (approved):
lgtm

PR #6221, opened 2026-09-16
Remove the sockbplog feature

markhannum:
Retires the "osql block-log over a plain socket" path; the offload net is the only transport left. Two commits: the appsock plugin (233 lines), then the machinery -- `db/osqlsqlsocket.{c,h}`, the `gbl_sockbplog*` tunables, the helpers they left dead, the `set sockbplog` handler in newsql, and the tests and docs. 926 deletions; no `sockbplog` references remain.

The `osql_target` type/sb transport indirection is deliberately left in place.

Tested: cmake configure and the `comdb2` target build clean with no new warnings; `tests/tunables.test` expected output updated per AGENTS.md.


roborivers (changes requested):
<p>Failed to merge main into this PR - resolve conflicts and re-run &#x26A0;.</p>

PR #6222, opened 2026-09-16
{186098196}: Remove min_retries, return on the first connect failure

chands10:
min_retries capped how many times a non-hasql statement would retry before giving up with "Cannot connect to db". That give-up was gated on hndl->sb being NULL, which is true both when we could not reach any node and when we connected fine and then tore the connection down after a failed query. newsql_disconnect() nulls sb, so a server that is up but shedding load looked identical to an unreachable cluster.

That misfires once a cluster is large. The give-up fires at max(num_hosts, min_retries) + 1, but ALLOW_QUEUING is not sent until num_hosts + 3, so from roughly fourteen nodes upward the handle quits two iterations before it would ever ask the server to queue the request. On a sixteen node cluster a saturated database therefore surfaced as a connection error while the nodes were up and would have queued the work.

Drop min_retries entirely and restore the behaviour cdb2api had before it: return as soon as cdb2_connect_sqlhost() fails. That call already sweeps every replicant, falls back to the master and re-runs discovery, so one failure means a full pass reached nothing and sweeping again buys nothing. A query rejected or dropped after connecting does not reach that return - it re-enters the loop and reconnects - so those retries now run to max_retries and ALLOW_QUEUING is sent as intended.

A snapshot-hasql commit moves hndl->queries into commit_query_list and not every path back to retry_queries restores it, so the new return frees it. in_trans needs no reset: a begin is rejected up front when in_trans is already set, and nothing between there and the return can set it.

sqllogfill set min_retries to 1 so a saturated master gives up quickly rather than grinding; it now uses cdb2_hndl_set_max_retries() for the same effect. logfill_request_failed() already matches CDB2ERR_TRAN_IO_ERROR, so the resulting error code is still caught.


chands10:
If we want to keep min retries, this can be used as an alternative fix: https://github.com/bloomberg/comdb2/pull/6223

chands10:
Use same test as https://github.com/bloomberg/comdb2/commit/dea6bc0566a837db854a3f3cbe33bb25a512331c

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>sc_truncate_multiddl_generated [db unavailable at finish] **quarantined**<br>comdb2sys_queueodh_generated [db unavailable at finish]<br>incoherent_sql [db unavailable at finish]<br>ruleset <br>pmux_secure_route <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>api_events <br>appsock <br>sc_downgrade [timeout] **quarantined**</p>

markhannum (dismissed):

chands10:
/adhoc-test appsock api_events pmux_secure_route ruleset incoherent_sql

markhannum (approved):

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: <b>harness did not run</b> - no test results were produced &#x26A0;.</p>

roborivers (approved):
<p>Cbuild submission: Success &#10003;.<br>Regression testing: Success &#10003;.</p>

PR #6223, opened 2026-09-16
{186098196}: Count connect failures instead of retries

chands10:
The non-hasql give-up was gated on hndl->sb being NULL, which is true both when we could not reach any node and when we connected fine and then tore the connection down after a failed query. newsql_disconnect() nulls sb, so a server that is up but shedding load looked identical to an unreachable cluster.

That misfires once a cluster is large. The give-up fired at max(num_hosts, min_retries) + 1, but ALLOW_QUEUING is not sent until num_hosts + 3, so from roughly fourteen nodes upward the handle quit two iterations before it would ever ask the server to queue the request. On a sixteen node cluster a saturated database surfaced as a connection error while the nodes were up and would have queued the work.

Count the thing the give-up is actually about. hndl->connect_failures is incremented at the single failure exit of cdb2_connect_sqlhost() and cleared in hndl_set_comdb2buf(), the only place sb becomes non-NULL, so it tracks consecutive failures to reach the cluster and nothing else. A rejected or dropped query reconnects successfully, leaves the counter at zero and now runs to max_retries, so ALLOW_QUEUING is sent as intended.

min_retries is renamed to max_connect_failures to match what it now bounds. The value and the !is_hasql guard are unchanged: hasql statements can be replayed safely and keep retrying, non-hasql statements still stop once the cluster looks unreachable, which is what the limit was added for.

Note that max_connect_failures counts only failures to connect, so a master that is reachable but saturated no longer short-circuits sqllogfill; it retries until max_retries. Bound total retries with cdb2_hndl_set_max_retries() if the old behaviour is wanted there.


chands10:
This is an alternative fix to the above ^^

roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>incoherent_sql [db unavailable at finish]<br>recover_deadlock <br>ruleset <br>pmux_secure_route <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>api_events <br>appsock <br>sc_downgrade [timeout] **quarantined**<br>truncatesc_offline_generated [timeout] **quarantined**</p>

PR #6224, opened 2026-09-17
comdb2uuid: generate v7 uuid from a thread-local prng

riverszhang89:
libuuid's `uuid_generate` goes through glibc random() which takes a process-global lock, serializing every id under concurrency. Emit RFC 9562 v7 (ms-timestamp prefix, time-ordered) from a per-thread prng. Guarded by tunable uuid_v7.


riverszhang89:
Benchmark (72-core host, standalone db, 128 parallel writers, autocommit inserts):

  | uuid_v7 | rows/s (mean ± stdev) |
  |---|--:|
  | on (this change) | 41,880 ± 3,000 |
  | off (libuuid)    | 31,444 ± 927   |
  
  **+33% insert throughput.** Guarded by the `uuid_v7` tunable (default on; off falls back to libuuid).

riverszhang89:
I've also changed our guid sqlite function to use the new per-thread seeded prng implementation. A guid-filled column is most likely used as an index, so it will be very beneficial if guid is monotonically increasing, just as this implementation does (it's time based).

Another benchmark on inserting into a dbstore=guid key'd table (see tests/guid_v7_bench):

Single writer inserts into a table whose byte[16] key is filled by the guid() dbstore default, across three uuid configurations:

    case 1  both off (libuuid v4)          4303 rows/s   baseline
    case 2  uuid_v7 (internal gen)         4977 rows/s   +15.7%
    case 3  uuid_v7 + sqlite_uses_uuid_v7  5884 rows/s   +36.7% (+18.2% over case 2)

case 2 gains from the thread-local prng in internal comdb2uuid callers  (e.g. genids); case 3 adds the SQL guid() path, whose v7 keys are monotonic and append to the right edge of the index btree instead of scattering and splitting pages like v4.

riverszhang89:
The contention on uuid seems to come from rqid:

```
#0  __lll_lock_wait_private ()
#1  random ()
#2  rand () 
#3  random_get_fd ()
#4  random_get_bytes ()
#6  uuid_generate_random ()
#7  uuid_generate ()
#8  comdb2uuid ()
#9  osql_sock_start_int ()
#10 osql_sock_start ()
#11 osql_insrec ()
...
More sqlite thread signatures skipped ...
```

PR #6229, opened 2026-09-19
Release page locks from inside a sort, and test it

dorinhogea:
A non-SI reader releases page locks from cursor_move_postop, once per cursor move.  A sort has no cursor moves: OP_SorterSort is one VDBE instruction that does the whole in-memory quicksort.  A writer wanting a page the reader holds waits out the entire sort.  The lock survives into the sort on plans that defeat early cursor close, such as MULTI-INDEX OR.

The fix checks for a page-lock waiter inside the sorter's merge loop and does a page-locks-only release there, keeping the curtran so table locks survive.  It belongs inside the loop because the waiter usually arrives during the sort, and because a nested sort must not pay a release per iteration with nobody waiting. This is safe: the merge touches no btree and every row is already in the sorter.

Tunables added:
  sort_release_check_interval      check for a waiter once every N record
                                   comparisons in the merge loop; 0 disables
                                   the check (default 1024)
  sort_release_probe_interval_us   minimum microseconds between those waiter
                                   probes (default 1000)
  debug_trace_sorter_rewind        log the time spent in the blocking window
                                   (testing only)

New comdb2_metrics counters sorter_rewinds_started and sorter_rewinds_finished publish how many sorts are in the blocking window, letting the test start its writer exactly when the window opens.  The sort is made slow with tying ORDER BY columns, which force a full record decode per comparison.  Disabled, the writer waits out the sort (5.6s, 7.9s); enabled, it gets through in 0.03s and 0.04s.


roborivers (changes requested):
<p>Cbuild submission: Error &#x26A0;.<br>Regression testing: Success &#10003;.</p><p>The first 10 failing tests are: <br>analyze_partial_index_off_generated [failed with core dumped] **quarantined**<br>analyze [failed with core dumped] **quarantined**<br>timepart_retro <br>sort_lockhold <br>consumer_non_atomic_default_consumer_generated  **quarantined**<br>datetimeusec <br>datetime <br>cdb2api <br>sc_downgrade [timeout] **quarantined**<br>skipscan [timeout] **quarantined**</p>

db/sqlglue.c
```diff
@@ -3691,6 +3710,80 @@ void sync_index_data_cursors(struct sql_thread *thd)
...
+void comdb2_sort_release_check(void)
+{
+    struct sql_thread *thd = pthread_getspecific(query_info_key);
+    if (!thd) /* e.g. a sorter worker thread, which has no sql_thread */
+        return;
+    struct sqlclntstate *clnt = thd->clnt;
+    if (!clnt || !clnt->dbtran.cursor_tran)
+        return;
+    if (clnt->sort_pagelocks_released)
+        return;
```
WalidNejmi: if i understood correctly, this line here prevents releasing locks when not necessary if they were already released and it works when there is only one batch of sorting perfectly. But vdbeSoryerMerge() also runs during pma flushes even before rewind. So let's say pma#1 finds a waiter, releases the page locks and then sets `sort_pagelocks_released = 1`. Then it will continue scanning for rows and get new page locks. Let's say now pma#10 see `sort_pagelocks_released = 1`and skip checking whether there is a waiter, even though there could be another waiter waiting on these new locks and we could in theory optimize letting this waiter go through by releasing these new page locks. maybe we should only have the sort_page_locks_released in the vdbe rewind function?

